PR #56, opened 2026-02-22
Mike's Grpc metadata service

jtnkminimal:


jtnkminimal:
# gRPC Metadata Service Guide

## Overview

The in-sandbox metadata service lets users dynamically add packages to a running sandbox via `min add`. It was migrated from HTTP/JSON (axum) to gRPC (tonic) for type-safe proto codegen and future streaming support.

## Architecture

- **Proto definition**: `crates/min-client/proto/min.proto` defines `MinService` with 4 RPCs
- **Server**: `crates/minimal/src/shim_listener.rs` implements `MinService` via tonic, runs on a background thread during sandbox tasks
- **Client**: `crates/min-client/src/main.rs` is a tonic gRPC client binary injected into the sandbox as `/usr/bin/min`
- **Shared types**: `min-client` is a lib+bin crate — the lib exports proto-generated types used by both server and client

## RPCs

**Add** — add packages to the sandbox rootfs. Resolves transitive deps, checks cache, builds if needed, hardlinks into rootfs, computes env vars.

**ListPackages** — list all packages currently installed in the rootfs with their spec hashes.

**GetEnv** — get the computed environment variables (PATH, LD_LIBRARY_PATH, etc.) for all installed packages.

**GetMeta** — get task metadata (task name, rootfs path, minimal.toml path).

## Using it in a sandbox

From inside a sandbox shell:

```bash
# Add a package (triggers env export via bashrc wrapper)
min add zlib

# Add multiple packages
min add zlib expat libffi

# Add without persisting to minimal.toml
min add --ephemeral gcc
```

The `min` bash function in `/etc/bashrc` wraps the binary and evals its output, so `export` statements take effect in the current shell.

## Building

```bash
# Build everything
cargo build

# Build just the client binary
cargo build --package min-client

# Build just the server (part of minimal)
cargo build --package minimal
```

## Testing

```bash
# Run all shim_listener tests (21 tests: 13 gRPC integration + 8 unit)
cargo test --package minimal -- shim_listener

# Run a specific test
cargo test --package minimal -- shim_listener::tests::test_grpc_add_single_package

# Run the full workspace test suite
cargo test
```

## Verifying the client binary

```bash
# Should print "error: not running in a managed minimal environment"
# because there's no /state/.min/port file outside a sandbox
./target/debug/min-client add zlib
```

## How the server starts

1. `cmd_run` creates a `ShimListenerConfig` with the dep graph, cache, rootfs path, etc.
2. `shim_listener::start(config)` binds to `127.0.0.1:0`, writes the port to `/state/.min/port`, and spawns a background thread running a tonic gRPC server
3. The server stays alive for the duration of the sandbox task
4. `shim_handle.shutdown()` stops it when the task exits

## Modifying the proto

1. Edit `crates/min-client/proto/min.proto`
2. Run `cargo build --package min-client` — `build.rs` regenerates the Rust types
3. Update the server implementation in `shim_listener.rs` and client in `main.rs` to match
4. Run `cargo test --package minimal -- shim_listener` to verify

PR #73, opened 2026-03-24
Cargo.lock: bump rustls-webpki & tar to fix advisories

twitchyliquid64:


Cargo.lock
```diff
@@ -68,7 +68,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
```
msample, line 71: multiple deps on this pulling it back?  Is this a nag a transitive dep package maintainer kind of task?    Why are we using windows-sys in the first place given that we compile for Linux?

msample (approved):
LGTM but would be nice to remove the windows-sys dependency if possible

PR #75, opened 2026-03-24
assorted: plumb cancellation to sandbox executions, handle ctrl-c for `minimal package`

twitchyliquid64:
This will gracefully shutdown a package build, cleaning up resources as well as committing successful builds when interrupted.

Will also allow us to abort sandbox tasks in res-server.


msample (approved):
LGTM

PR #76, opened 2026-03-25
.github: update minimal-check job to not fetch all packages

twitchyliquid64:
It seems a refactor removed the logic where we would only fetch packages local to the repo, rather than everything in the graph. Whoops.

Sets `--no-fetch` for now to unblock CI.


msample (approved):
Is this the right long term fix?

twitchyliquid64:
> Is this the right long term fix?

I think we should fix it so it doesnt download the whole graph, but even if we do, that wont go out for another week on account of the freeze.

PR #80, opened 2026-03-27
crates/minimal: add spec_hash to --packages dump output

### LINKED ISSUE #78: Add spec hash to minimal dump --packages output
## Context

When using repro-check to scan ~/.cache/minimal/built/, findings are reported by spec hash (e.g., 4d/ff9724...). There is currently no way to map these back to package names without file fingerprinting.

## Request

Add the computed spec hash to the `minimal dump --packages` JSON output. Something like:

```json
{
  "name": "python",
  "spec_hash": "4dff9724e3d7210fd69d603ac9f05f2dde491ef07cad42661a4cc7adb80ad320",
  ...
}
```

This would let tooling (like repro-check) resolve cache directories to package names directly.

## Use case

`repro-check inspect ~/.cache/minimal/built/ --summary` auto-detects the cache structure and reports per spec hash. With the spec hash in the dump output, it could report by package name instead, and also filter to only scan the current (non-stale) builds.

twitchyliquid64:
Fixes #78

### DIFF
--- a/crates/minimal/src/cmd_dump.rs
+++ b/crates/minimal/src/cmd_dump.rs
@@ -1,7 +1,7 @@
 use std::{fmt::Display, io::stdout};
 
 use decode::AttrValue;
-use graph::{BuildDep, BuildOutput, BuildSpec, Graph, RuntimeDep, SourceInput};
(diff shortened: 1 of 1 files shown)


bryane42 (approved):
LGTM

PR #81, opened 2026-03-29
crates/decode: fix panic caused by missing field value in build_args

twitchyliquid64:
Not the best fix, but will stop the panic which breaks everything else + at least the pretty-print will highlight the broken section


msample (approved):
LGTM

PR #84, opened 2026-03-30
assorted: refactor keyword search to common implementation, add version in search output

twitchyliquid64:
* Move `fuzzy_search` to `common` crate, create real result type
 * Tests
 * Prioritize exact matches over prefix matches over fuzzy matches
 * Use for both dump filter and `min search`
 * Show version in `min search`:

```shell
bash-5.3$ min search ll
 * llvm
 * llama.cpp (version b8041)
 * llvm-bootstrap
 * zellij (version 0.43.1)
 * nushell (version 0.110.0)
 * shellcheck (version 0.10.0)
 ```

### DIFF
--- a/crates/common/src/fuzzy_search.rs
+++ b/crates/common/src/fuzzy_search.rs
@@ -1,0 +1,178 @@
+//! Fuzzy string matching for package name search.
+
+use std::cmp::Ordering;
+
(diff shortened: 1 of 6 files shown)


msample (approved):
LGTM

PR #86, opened 2026-04-01
crates/common: fix error checking in Gcs fetcher impl, check request error in CachingDownloader

twitchyliquid64:
Fixes a panic when trying to populate an entry in `CachingDownloader` (i.e. source fetch) where the `FetchBackend` returns a non-200 code, e.g.

```
called `Result::unwrap()` on an `Err` value: Error { kind: Transport(TransportDetails { status_code: Some(404), headers: Some({"content-type": "text/html; charset=UTF-8", "x-guploader-uploadid": "<redacted>", "date": "Wed, 01 Apr 2026 <redacted>", "vary": "Origin", "vary": "X-Origin", "cache-control": "private, max-age=0", "content-length": "76", "server": "UploadServer", payload: Some(b"No such object: minimal-staging-archives/clang-tools-extra-21.1.8.src.tar.xz") }), source: None }
```

### DIFF
--- a/crates/common/src/fetchers.rs
+++ b/crates/common/src/fetchers.rs
@@ -182,7 +182,10 @@ impl FetchResponse for Result<google_cloud_storage::read_object::ReadObjectRespo
         self.is_ok()
     }
     fn error_for_status(self) -> Result<Self, Self::Error> {
-        Ok(self)
+        match self {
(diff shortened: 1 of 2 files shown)


msample (approved):
LGTM

PR #87, opened 2026-04-01
crates/{check,sandbox2}: set SOURCE_DATE_EPOCH/PYTHONHASHSEED in all build sandboxes

twitchyliquid64:
This also adds a lint for build scripts that set it themselves, which would be annoying if we couldnt just tell a clanker to fix it.

image


msample (approved):
LGTM - won't a lot of checks fail now?

twitchyliquid64:
Yepp! But we need to do a ralph loop fix the world pass already

PR #90, opened 2026-04-01
assorted: synchronize only hakoniwa spawn, not file writes

twitchyliquid64:
Hakoniwa has closed unneeded FD's before spawn returns for some time now, which afaict negates the need for us to sequences writing out package files with sandbox spawns.

### DIFF
--- a/crates/common/src/archive.rs
+++ b/crates/common/src/archive.rs
@@ -198,7 +198,6 @@ fn extract_tar_impl<R: Read>(
     dest_dir: &Path,
     strip_prefix: Option<&String>,
 ) -> Result<(), ArchiveError> {
-    let _guard = super::FdSynchronizer::lock_writing_files();
     let mut archive = tar::Archive::new(reader);
 
     if let Some(prefix) = strip_prefix {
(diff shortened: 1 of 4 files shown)


twitchyliquid64:
I'm going to do a full rebuild on my framey to make sure before merging.

twitchyliquid64:
Full build on amd64 without any regression.

msample:
Do you recall why we locked filed during spawn before? Was it locking *writing* or was it locking opening so we didn't leak new FDs to the child process?

msample:
I have seen some idle times in big `minimal pkg` runs - I wonder if this is related.

twitchyliquid64:
It was because hakoniwa didnt close file descriptors before returning from spawn, so there was nothing for us to synchronize to to prevent an open writeable handle to an executable file getting caught in a hakoniwa fork and not closed in time. So we prevented that from ever happening by locking all the writes of executable files.

I'm not even sure we need to lock spawn's of sandboxes at all, but being conservative for now.

> I have seen some idle times in big minimal pkg runs - I wonder if this is related.

Yepp! particularly for sources it hasnt fetched yet.

msample (approved):
LGTM. Good idea!

PR #91, opened 2026-04-01
crates/multi: Create new subsystem for running detached, interactive tasks

twitchyliquid64:


jtnkminimal:
LGTM!

msample (commented):
Nice decomposition of the problem! 

Seems like *a lot* of significant changes that may need a lot of internal testing before promoting.  Can this live on a branch until we get stable/next in place so we don't block promotion of other potentially highly need changes?

twitchyliquid64:
It can, but its only going to get larger lol.

While a crate in this repo, its mostly separate to the other code here that powers the CLI, so I don't expect work here to cause regressions elsewhere.

msample (approved):
This code is not wired in to the active execution path so OK to merge - it won't impact our users until we wire up, and we'll likely have stable/next by then.

PR #98, opened 2026-04-06
crates/sandbox2: fix overflowing small stack for min commands

twitchyliquid64:
Fixes:

```sh
xxx@lptp:~/minimal$ cargo run -- -C ~/pkgs shell

bash-5.3$ min check bash

thread '<unknown>' (1568181) has overflowed its stack
fatal runtime error: stack overflow, aborting
bash-5.3$ Aborted
```


msample (approved):
LGTM

PR #99, opened 2026-04-06
crates/check: add checks for missing interpreter for executable scripts

twitchyliquid64:
This will cause a dozen or so packages to go red but thats okay, I will fix before this gets promoted


msample (approved):
LGTM

PR #104, opened 2026-04-07
crates/orchestrator: make cpu shares more aggressive for critical-path builds

twitchyliquid64:
Shares are linear and relative to every other cgroup in the parent, which is still too even across the concurrent builds. Bump up the multiple to make critical-path builds get an even greater share.


msample (approved):
LGTM

PR #106, opened 2026-04-07
crates/sandbox2: do not configure cgroups without systemd

twitchyliquid64:
Systemd is not a given, but necessary for unpriv cgroups. Its best effort, so disable it if not available. Fixes:

```shell
2026-04-07T17:48:46.841195Z ERROR orchestrator: build { full: false, hash: "9a467b5ac7af0a1a0a156498c1954b962bb2b49e494d7c6cbbd8aa9576037060", spec: "sed" }
    construction failed: sandbox error: Invocation 0 failed with exit code 125:
    systemd error: dbus error: dbus bus address error: session bus address /run/user/1000/bus does not exist
```

### DIFF
--- a/crates/sandbox2/src/lib.rs
+++ b/crates/sandbox2/src/lib.rs
@@ -399,15 +399,19 @@ impl<C: Channel> Sandbox<C> {
             container.hostname(hn);
         }
         if let Some(s) = &self.config.cpu_weight {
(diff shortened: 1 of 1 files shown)


msample (approved):
LGTM - systemd - the single CGroups V2 writer to enforce consistency solution.

PR #107, opened 2026-04-08
crates/remote{client,proto}: impl download RPC from res, pass remote cache bucket to build

twitchyliquid64:
The `minimal` repo half of:

1. Letting buildbot decide what cache to use (passes through as RPC arg)
2. Having a way for buildbot to download the artifacts from a successful build (download RPC).


bryan-minimal (approved):
LGTM

PR #108, opened 2026-04-08
crates/common: increase default num jobs based on CPU count

twitchyliquid64:
Now that cgroups help prioritize builds, we can increase the load multiple even more.


msample (approved):
LGTM

PR #111, opened 2026-04-10
Support configurable architecture for OCI image materialize

bryan-minimal:
## Summary

OCI image architecture was hardcoded to `amd64`. This adds configurable arch support so arm64 images can be built for Cloud Run (which is ~20% cheaper on arm64).

## Architecture priority (highest first)

1. **CLI flag**: `minimal materialize image -o out.tar --arch arm64`
2. **minimal.toml**: `[outputs.image]` with `arch = "arm64"`
3. **Host default**: `arm64` on aarch64 hosts, `amd64` on x86_64

## Changes

- `mfile/lib.rs`: add optional `arch` field to `Output` struct
- `mctx/lib.rs`: add `graph_from_all_packages_for_target()` and `graph_from_package_names_for_target()` — builds the package graph for the requested architecture
- `cmd_materialize.rs`: resolve target arch, build graph for that target, pass arch to OCI image builder. Accepts `aarch64`/`x86_64` as aliases.
- `op/oci_image.rs`: use `self.arch` instead of hardcoded `"amd64"` (2 locations)

The graph is built for the correct target so packages/binaries match the architecture stamped in the OCI image metadata.

## Usage

```toml
# minimal.toml
[outputs.image]
ty = "oci-image"
arch = "arm64"
```

```bash
# Or override at CLI
minimal materialize image -o dist/app.tar --arch arm64
```

## Tests (4 new, 17 total in mfile)

- `output_arch_defaults_to_none` — no arch field = None (host default)
- `output_arch_parses_arm64` — `arch = "arm64"` deserializes
- `output_arch_parses_amd64` — `arch = "amd64"` deserializes
- `output_with_arch_and_other_fields` — arch works alongside packages, entrypoint, vars
- Existing `deserialize_smoketest` updated to include `arch: None`

## Note

`cargo check` on macOS fails on `procfs` (Linux-only transitive dep from the sandbox crate). Compiles in the minimal sandbox / CI.

### DIFF
--- a/crates/mctx/src/lib.rs
+++ b/crates/mctx/src/lib.rs
@@ -396,6 +396,11 @@ impl Context {
 
     /// Builds & returns a graph of all packages.
     pub fn graph_from_all_packages(&mut self) -> Result<Graph, Error> {
+        self.graph_from_all_packages_with_target(Target::host())
+    }
+
+    /// Builds & returns a graph of all packages for a specific target.
+    pub fn graph_from_all_packages_with_target(&mut self, target: Target) -> Result<Graph, Error> {
         let leaf_layer = self.repo_origin();
 
         let start = SystemTime::now();
@@ -404,12 +409,23 @@ impl Context {
             &mut graph::LayerCacheDir(self.config.layer_cache_dir()),
             leaf_layer,
             self.stdlib_dir.clone(),
-            Target::host(),
+            target,
         )
         .map_err(|e| e.into());
         tracing::trace!("graph parse/load took {:?}", start.elapsed());
 
         res
+    }
+
+    /// Builds & returns the graph with the given packages for a specific target.
+    pub fn graph_from_package_names_with_target<S: PackageSelection>(
+        &mut self,
+        pkgs: S,
+        target: Target,
+    ) -> Result<Graph, Error> {
+        let mut graph = self.graph_from_all_packages_with_target(target)?;
+        graph.top_levels = pkgs.as_bsrs(&graph)?;
+        Ok(graph)
     }
 }
 
--- a/crates/minimal/src/cmd_materialize.rs
+++ b/crates/minimal/src/cmd_materialize.rs
@@ -1,6 +1,8 @@
 //! Command to materialize outputs defined in the minimal file.
 
 use anyhow::anyhow;
+use common::Target;
+use common::target::{Arch, OS};
 use std::path::PathBuf;
 
 use mctx::{Context, Error};
@@ -10,6 +12,11 @@ pub struct MaterializeArgs {
     /// The output file to write
     #[arg(short, long)]
     output: PathBuf,
+
+    /// Target architecture for OCI images (e.g., "amd64", "arm64").
+    /// Overrides the `arch` field in minimal.toml and the host default.
+    #[arg(long)]
+    arch: Option<String>,
 
     /// The name of the output in `minimal.toml` to materialize
     output_name: String,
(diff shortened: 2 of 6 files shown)


crates/minimal/src/cmd_materialize.rs
```diff
@@ -36,6 +41,15 @@ pub async fn cmd_materialize(args: MaterializeArgs, ctx: &mut Context) -> Result
     // Make sure the packages are built
     crate::cmd_pkg::pkg_build_impl(&graph, ctx, cache.clone(), false, None).await?;
 
+    // Architecture priority: CLI flag > minimal.toml > host default
+    let arch = args.arch.or(output.arch).unwrap_or_else(|| {
+        if cfg!(target_arch = "aarch64") {
```
twitchyliquid64: Use `Target::host()` as the default

crates/op/src/oci_image.rs
```diff
@@ -119,9 +121,12 @@ impl Runnable for OciImageCreate {
         let mut tb = tar::Builder::new(w);
 
         // ImageConfig - written as blob object by hash
+        let arch = &self.arch;
+        info!("OCI image architecture: {arch}");
+
         let image_config_bytes = serde_json::to_vec(
             &ImageConfigurationBuilder::default()
-                .architecture("amd64")
+                .architecture(arch.as_str())
```
twitchyliquid64: We should probably read this value from the arch that the `Graph` was constructed for, otherwise its possible to stamp an image config that declares arch XY while all the cache objects / files / binares are arch YZ.

twitchyliquid64: It doesn't seem like `Graph` tracks what arch it was built for today, so we might need a new field / getter that we populate on initialization from the `for_target` field of the `Layer`.

crates/minimal/src/cmd_materialize.rs
twitchyliquid64: Its at this point (where we call for a graph) that we need to plumb arch-awareness - the graph needs to be built for a certain target.

twitchyliquid64 (commented):
TL;DR: we need to make the graph that is constructed match the arch that is requested. So theres some code changes there to plumb through the requested arch to the code that constructs the graph.

Given its possible for the graph and the requested value to be out of sync (due to some future bug), its probably also more reliable for us to get the arch value we stamp in `ImageConfigurationBuilder.architecture` from the `Graph` - whatever it was constructed for.

crates/minimal/src/cmd_materialize.rs
bryan-minimal: Done — defaults to `Target::host()` now. Priority is CLI flag > minimal.toml > host default. Also accepts `aarch64`/`x86_64` as aliases.

crates/op/src/oci_image.rs
bryan-minimal: Addressed — the graph is now built for the requested target via `graph_from_package_names_for_target(pkgs, target)`. The arch stamped in the OCI image comes from the same resolved arch string that was used to construct the target, so they can't diverge.

bryan-minimal: Rather than adding a field to Graph, I plumbed the arch through from the resolution in `cmd_materialize` — the same `arch_str` that determines `Target::new(arch, OS::Linux)` for graph construction is passed to `OciImageCreate.arch`. If Graph grows a `target()` getter later, we could read it from there instead.

crates/minimal/src/cmd_materialize.rs
bryan-minimal: Done — added `graph_from_package_names_for_target()` and `graph_from_all_packages_for_target()` to Context. The materialize command resolves the arch, constructs `Target::new(arch, OS::Linux)`, and passes it through to graph construction. Packages are built for the correct target before the OCI image is assembled.

crates/mctx/src/lib.rs
```diff
@@ -396,6 +396,14 @@ impl Context {
 
     /// Builds & returns a graph of all packages.
     pub fn graph_from_all_packages(&mut self) -> Result<Graph, Error> {
+        self.graph_from_all_packages_for_target(Target::host())
+    }
+
+    /// Builds & returns a graph of all packages for a specific target.
+    pub fn graph_from_all_packages_for_target(
```
twitchyliquid64: Nit: in rust it is idiomatic to do specialization methods with the suffix verb `with`, so this would be `graph_from_all_packages_with_target`. Here and below

twitchyliquid64 (commented):
Last thing: lets have `Graph` track what `Target` is was built for, that way in `cmd_materialize.rs` you can query it as the source of truth rather than pass it through as a struct parameter.

crates/mctx/src/lib.rs
bryan-minimal: Good call — renamed to `graph_from_all_packages_with_target` and `graph_from_package_names_with_target`. Also ran cargo fmt + clippy + mfile tests (17 passing) before pushing.

bryan-minimal: Done — `Graph` now tracks the `Target` it was built for:

- Added `target: Target` field to `Graph`, set from `for_target` in `new_from_chain()` (defaults to `Target::host()` in `new()`)
- Added `Graph::target() -> &Target` accessor
- `OciImageCreate` no longer has an `arch: String` field — reads it from `opts.graph.target()` instead, so the graph is the single source of truth
- Wire format carries the target in the `HeaderRecord` (as `Option<Target>` with `#[serde(default)]` so old streams still deserialize)
- 44 graph tests pass, wire roundtrip tests confirm the target survives serialization

twitchyliquid64 (approved):

PR #113, opened 2026-04-11
Fix /etc/passwd root home to match sandbox HOME

### LINKED ISSUE #94: macOS: /etc/passwd sets root home to /root, breaks Go programs that use user.Current()
## Summary

On macOS (Apple Containerization VM), Go programs that resolve the home directory via `user.Current().HomeDir` (which reads `/etc/passwd`) get `/root` instead of the actual `$HOME`. Since `/root` doesn't exist on the sandbox rootfs (which is read-only), these programs fail with:

```
failed to create '/root/.pulumi': mkdir /root: read-only file system
```

## Root Cause

This is in the **minimal CLI's sandbox** (sandbox2), not the VM itself. The VM sets `HOME` correctly, but sandbox2 creates a new rootfs from packages and `synth_user_group_config` in `common/src/lib.rs:332` hardcodes root's home as `/root` in the synthesized `/etc/passwd`:

```rust
format!(
    "root:x:0:0:root:/root:/usr/bin/bash\n{}:x:1000:1000:User,,,:/home/{}:/bin/bash",
    username, username
),
```

The sandbox correctly sets `HOME=/host/Users/<user>` via env var, but `/etc/passwd` still says `/root`. The sandbox rootfs is built from package directories (read-only overlayfs layers), so `/root` doesn't exist and can't be created.

On Linux, `/root` typically exists in the package-derived rootfs, so this never surfaced. On macOS (where the sandbox runs inside a VM with a different filesystem layout), it does.

Go's `os.UserHomeDir()` checks `$HOME` first (works fine), but `os/user.Current().HomeDir` reads `/etc/passwd` (returns `/root`). Programs like Pulumi use the latter path.

## Suggested Fix

`synth_user_group_config` should set root's home directory in `/etc/passwd` to match the actual `$HOME` env var (or at least to a path that exists and is writable in the sandbox, like `/state/home`).

## Workaround

Setting `PULUMI_HOME="$HOME/.pulumi"` in the task's bash script bypasses the issue for Pulumi specifically, but any Go/C program that reads `/etc/passwd` for the home directory will hit this.

## Reproduction

```toml
# .minimal/minimal.toml
[tasks.demo]
bash = "pulumi login gs://some-bucket"
patch.dir."~/.pulumi" = "rw"
```

```
$ minimal run demo
error: problem logging in: failed to create '/root/.pulumi': mkdir /root: read-only file system
```

## Environment

- macOS Tahoe (Apple Silicon)
- minimal running via Apple Containerization VM
- Issue is in sandbox2's rootfs, not the VM rootfs

bryan-minimal:
## Summary

- `synth_user_group_config` hardcoded root's home as `/root` in synthesized `/etc/passwd`
- Go's `user.Current().HomeDir` reads `/etc/passwd` (ignores `$HOME`), gets `/root`, which doesn't exist on read-only sandbox rootfs
- Now passes the actual home dir matching the `HOME` env var the sandbox sets

No behavior change on Linux — `/root` happened to exist there, but the value was still wrong. Now it's correct on both platforms.

Fixes #94

## Changes

- `common/src/lib.rs`: `synth_user_group_config` takes a `root_home` parameter
- `sandbox2/src/config.rs`: computes root home from `WdSetup` (isolated = `/state/home`, bound = host `$HOME`)

## Tests

65 tests pass (common + graph), fmt + clippy clean in minimal container.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/sandbox2/src/config.rs
+++ b/crates/sandbox2/src/config.rs
@@ -358,9 +358,15 @@ impl Config {
             common::synth_dns_config(&sd)
                 .map_err(|e| Error::IO("synthesizing DNS configuration", sd.clone(), e))?;
         }
+        let root_home = match &self.wd {
+            WdSetup::Isolated { .. } => "/state/home".to_string(),
+            WdSetup::BoundDir { .. } => {
+                std::env::var("HOME").unwrap_or_else(|_| "/state/home".to_string())
+            }
+        };
         match &self.username {
-            Some(n) => common::synth_user_group_config(&sd, n),
-            None => common::synth_user_group_config(&sd, "build"),
+            Some(n) => common::synth_user_group_config(&sd, n, &root_home),
+            None => common::synth_user_group_config(&sd, "build", &root_home),
         }
         .map_err(|e| Error::IO("synthesizing user/group configuration", sd, e))?;
 
--- a/crates/sandbox2/src/config.rs
+++ b/crates/sandbox2/src/config.rs
@@ -360,9 +360,9 @@ impl Config {
         }
         let home = match &self.wd {
             WdSetup::Isolated { .. } => "/state/home".to_string(),
-            WdSetup::BoundDir { .. } => {
-                std::env::var("HOME").unwrap_or_else(|_| "/state/home".to_string())
-            }
+            WdSetup::BoundDir { .. } => std::env::home_dir()
+                .and_then(|p| p.to_str().map(String::from))
+                .unwrap_or_else(|| "/state/home".to_string()),
         };
         match &self.username {
             Some(n) => common::synth_user_group_config(&sd, n, &home),
--- a/crates/sandbox2/src/config.rs
+++ b/crates/sandbox2/src/config.rs
@@ -358,15 +358,15 @@ impl Config {
             common::synth_dns_config(&sd)
                 .map_err(|e| Error::IO("synthesizing DNS configuration", sd.clone(), e))?;
         }
-        let root_home = match &self.wd {
+        let home = match &self.wd {
             WdSetup::Isolated { .. } => "/state/home".to_string(),
             WdSetup::BoundDir { .. } => {
                 std::env::var("HOME").unwrap_or_else(|_| "/state/home".to_string())
             }
         };
         match &self.username {
-            Some(n) => common::synth_user_group_config(&sd, n, &root_home),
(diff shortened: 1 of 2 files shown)


twitchyliquid64 (commented):
iiuc this wires the homedir for the root user to the same home as the main user (uid 1000) ?

bryan-minimal:
You're asking the right question, and I was wrong about which user matters.

I re-verified inside a fresh sandbox shell and processes actually run as `uid=1000 (build)`, not root. When I first debugged this back in April, `whoami` returned `root`, so I built the fix around that. Something changed in sandbox2 between then and now, but the key thing is that `user.Current()` looks up uid 1000 (the runtime uid), which returns `/home/build` from the unchanged passwd line. So this PR as originally written did nothing to fix the actual bug.

I reproduced the current failure in gominimal/infra without the `PULUMI_HOME` workaround:

```
error: could not get cloud url: failed to create '/home/build/.pulumi': mkdir /home: read-only file system
```

Same class of bug as the original `/root/.pulumi` report, just at a different path now that the runtime uid is different.

I pushed a new commit that applies the computed home to both passwd entries so it works regardless of which uid ends up running. The sandbox already sets `HOME` to the same value via env var, so now passwd and the env var agree. I also renamed `root_home` to just `home` since it's no longer root-specific.

bryan-minimal:
Just pushed a smaller version: only the uid 1000 entry gets `$HOME` (or `/state/home` for isolated). Root stays as `/root` since processes don't run as root in the sandbox anyway.

crates/sandbox2/src/config.rs
```diff
@@ -358,9 +358,15 @@ impl Config {
             common::synth_dns_config(&sd)
                 .map_err(|e| Error::IO("synthesizing DNS configuration", sd.clone(), e))?;
         }
+        let home = match &self.wd {
+            WdSetup::Isolated { .. } => "/state/home".to_string(),
+            WdSetup::BoundDir { .. } => {
+                std::env::var("HOME").unwrap_or_else(|_| "/state/home".to_string())
```
twitchyliquid64: use `std::env::home_dir()`

bryan-minimal: Done in de8e9cda.

twitchyliquid64 (approved):

PR #118, opened 2026-04-14
crates/decode: impl support for partitioned packages/ directories

twitchyliquid64:
All package managers that represent config in a file tree eventually grow to a size where its inefficient for both the git index and (some) filesystem directory impls to have thousands of directories, and end up using prefixes as partitions.

We aren't there yet, but adding support now isnt a bad idea so we don't have a breaking change down the line.

### DIFF
--- a/crates/decode/src/load.rs
+++ b/crates/decode/src/load.rs
@@ -96,17 +96,12 @@ macro_rules! annotate_record {
     };
 }
 
-fn build_decls_in_dir<P: AsRef<Path>>(
-    dir: P,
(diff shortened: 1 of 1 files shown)


msample (commented):
are there any tests for this?

twitchyliquid64:
Yepp! `build_decls_in_dir_partitioned`. The diff looks weird cuz its interleaved itself with an old test that isnt used anymore (build_decls_in_dir_filtered).

msample (approved):
LGTM

PR #119, opened 2026-04-14
all: nickel-lang-core 0.15.3 → 0.17.0

twitchyliquid64:
## Me

This was mostly AI, but i have pretty high confidence in its correctness:

 - Extensive test suite passes
 - Running it on `pkgs` yields exactly the same specs and spec-hashes as before
 - Ran `minimal dump` with both this commit and main, manually verified the diff

## Clanker

All 108 tests pass across all affected crates, clippy is clean, code is formatted.

  Here's a summary of the migration:

  nickel-lang-core 0.15.3 → 0.17.0 Migration Complete

  Core changes applied:

  1. `RichTerm` → `NickelValue` across all files. The fundamental value type changed from a term wrapper to a tagged-pointer value type (`nickel_lang_core::eval::value::NickelValue`).
  2. Term matching replaced with accessor methods: `Term::Str(s)` → `val.as_string()`, `Term::Bool(b)` → `val.as_bool()`, `Term::Record(r)` → `val.as_record()`, `Term::Array(a, _)` → `val.as_array()`, `Term::ForeignId(id)` →
  `val.as_foreign_id()`, etc.
  3. `eval_if_closure()` simplified: No longer checks for Term::Closure (which was removed). Uses val.is_whnf() + `Closure::from(val)` instead of `Closure::atomic_closure()`.
  4. `Container::Empty` handling: New API represents empty records/arrays as inline values. Added `is_record()` helper and empty-record fallbacks throughout.
  5. Position access: `rt.pos` → `val.pos(&program.pos_table())` or `val.pos_idx()`.
  6. custom_transform signature: Now takes `(&mut CacheHub, &mut PosTable, NickelValue)` instead of `(&mut CacheHub, RichTerm)`.
  7. `annotate_record!` macro: Updated for `Term::RecRecord(Box<RecRecordData>)` (single boxed field) and `NickelValue::foreign_id_posless()` instead of `RichTerm::from(Term::ForeignId(...))`.

  Files changed:

  - `Cargo.toml` - version bump
  - `crates/common/src/ncl_eval.rs` - NickelValue return type
  - `crates/decode/src/{lib,load,builds,attrs,profiles,harnesses,decl_tests,error}.rs` - full migration
  - `crates/check/src/{harness,profile}.rs` - record access pattern
  - `crates/minimal/src/cmd_dep.rs` - removed bytecode::ast::Node impor

### DIFF
--- a/.minimal/minimal.toml
+++ b/.minimal/minimal.toml
@@ -1,7 +1,7 @@
 [upstream] # Source of software & tooling
 repo = "https://github.com/gominimal/pkgs"
 branch = "main"
-locked_commit = "f50ffcb0dd274f8cc063d66ea9963819406f10d4"
+locked_commit = "20fb6d2868ed90693be7561f6c9e70a7a76d70ca"
 
 [harness]
 use = "rust"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


twitchyliquid64:
Okay just did:

```shell
$> cargo run -- -C ~/pkgs/ dump --packages > /tmp/dump_output_new
$> git checkout HEAD~1
$> cargo run -- -C ~/pkgs/ dump --packages > /tmp/dump_output_old
$> diff /tmp/dump_output_old /tmp/dump_output_new
```

And the diff is just position detection that the update fixed: i.e. a bunch of this
```
36176c40742,40745
<         "pos": null
---
>         "pos": [
>           871,
>           876
>         ]
36187c40756,40759
<             "pos": null
---
>             "pos": [
>               954,
>               962
>             ]
36191c40763,40766
<             "pos": null
---
>             "pos": [
>               979,
>               984
>             ]
36197c40772,40775
<         "pos": null
---
>         "pos": [
>           232,
>           239
```
I just scrolled the whole diff and 100% of it is just position information that can now be determined.

msample (approved):
LGTM - full pkgs rebuild successful.

PR #120, opened 2026-04-14
crates/check: spawn futures concurrently to speed up check

twitchyliquid64:


msample (approved):
LGTM.  Nice use of a macro.

PR #128, opened 2026-04-15
.github: conserve cache in build task

twitchyliquid64:
* ~Use `checkout@v5` which actually exists, I guess v6 was just falling back~
 * Change the cache key to invalidate the ones that are already too big
 * Use `cargo-cache` to trim unneeded objects at the end

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,8 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
Nvm there is a v6, switching back

msample (approved):
LGTM - don't they recommend pinning to hashes (vs @v6) to avoid re-tagging games?

twitchyliquid64:
Seems they just recco `@v6`, see https://github.com/actions/checkout?tab=readme-ov-file#usage

PR #129, opened 2026-04-15
Let `--no-{cache,fetch}` be specified after subcommand, support minimal package --rebuild

twitchyliquid64:
* You can now do `minimal <command> --no-cache --no-fetch` instead of always having those options before the subcommand, feels more natural sometimes
 * * NOTE: I havent done this for `-C` as I'm not sure thats natural yet
 * You can now do `minimal package --rebuild <packages>` to force a rebuild of specifically the named packages.


msample (approved):
LGTM

PR #133, opened 2026-04-16
crates/{args,mfile}: move argument schema code to own crate

twitchyliquid64:


crates/args/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+//! Typed arguments support for tasks & sideloads.
```
msample, line 1: a bit more context here would helpful - e.g. how is is this related to toml and how does the Rust CLI parser clap come into this?  What are the typical input sources and what consumes the outputs?

msample (approved):
LGTM - would be nice to address the comments comment in subsequent updates.

crates/args/src/lib.rs
twitchyliquid64: `clap` is used to parse the invocation when args are used for a task and its run from the CLI: i.e. `minimal run <task> [<args>]`. `toml::Value` is used as a container for a hydrated schema - a more mature implementation would probably implement our own Value struct/enum.

PR #136, opened 2026-04-17
Implement sideloading

twitchyliquid64:
- Moved the logic for walking the supply chain out of a `Graph` method and into its own type
 - Refactored the storage types for arguments to use `DiskArg`, plumbed through loader methods
 - Implemented Newtype `DiskArg` for loading arguments from config when schema is not yet known, and method to hydrate/validate based on schema
 - Plumb writing parameters into nickel `__injected_config__.ncl`. We will do this without writing a file to disk when that becomes possible
 - Test for sideloading, including parameters, parameter defaults, load/ingest order

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1598,6 +1598,7 @@ dependencies = [
 name = "graph"
 version = "0.0.1"
 dependencies = [
+ "args",
  "blake3",
  "checkouts",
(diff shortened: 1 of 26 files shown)


msample (approved):
LGTM!  Great comments for args.

PR #140, opened 2026-04-18
crates/{mctx,orchestrator}: limit max concurrent downloads to 8

twitchyliquid64:
Because 200+ concurrent downloads is probably not a good thing for TCP congestion control.


msample (approved):
LGTM

PR #142, opened 2026-04-20
crates/stdlib: add attrs: license_spdx, website variant for source_provenance

twitchyliquid64:


bryan-minimal (approved):
LGTM, this will also benefit pkgmgr

PR #144, opened 2026-04-20
crates/graph: separate API for manipulating graph from code building it

twitchyliquid64:
Not net-new code, just moving the graph builder code to a separate file `loader.rs`.


bryan-minimal (approved):
LGTM

PR #146, opened 2026-04-20
crates/minimal: cmd_dump accepts --arch to override target

### LINKED ISSUE #125: Feature: minimal dump should accept a target arch override
# Feature: `minimal dump --packages` should accept a target arch override

## Summary

`minimal dump --packages` always evaluates Nickel against the host machine's target arch. For packages that use `match { arch = 'Amd64, .. } => ..., { arch = 'Arm64, .. } => ... } target` to select between per-arch source URLs, only the host arm is visible in the dump output — the other arm is erased at evaluation time with no breadcrumb in the JSON.

This creates blind spots in tooling downstream of `minimal dump`:

1. **pkgmgr** (the package updater) silently corrupts arch-matched packages. Its rewriter replaces the first sha256 it finds in `build.ncl`, which may be either the amd64 or arm64 binding depending on file ordering — while only actually fetching the host arm's new tarball. The result is one correct hash and one stale hash, and no warning.

2. **supply-chain's scanner** only sees one arch's resolved source URL per package. For packages whose non-host arm points at a different lockfile-bearing archive, the scanner can miss advisories that affect the non-host arch.

3. **Any future SBOM/attestation tooling** would inherit the same blind spot.

Today there are ~10 packages in `gominimal/pkgs` whose primary `build_deps` Source is arch-matched — claude-code, go, deno, bun, gcloud, actions-runner, varlock, libsodium, mermaid-cli (plus potentially others). A further ~29 packages have arch-matched sources inside `replace_on_cycle.build_deps` (the prebuilt cycle-breaker path), which is a separate problem (prebuilt regen).

## Proposal

Add a target-arch override to `minimal dump` so consumers can evaluate each arch independently and aggregate:

```
minimal dump --packages --target amd64
minimal dump --packages --target arm64
```

Or as an environment variable if that fits the CLI better:

```
MINIMAL_TARGET_ARCH=amd64 minimal dump --packages
MINIMAL_TARGET_ARCH=arm64 minimal dump --packages
```

Either shape works; env-var is slightly friendlier for supply-chain because it doesn't require argv plumbing in the shell-out.

The flag only needs to override the `target` value passed to Nickel evaluation before the match expression resolves. It does **not** require a new evaluator mode, a new JSON output shape, or any schema migration — the existing dump format already has a `url`/`sha256` field per dependency; it just becomes "the url/sha256 for this target."

## Why this is the right shape

- **Minimal change**: the evaluator already takes a `target` from `config.ncl`. This flag just substitutes a different value before eval. Probably 20-30 LOC.
- **Supply-chain implementation collapses to ~50 LOC**: run `dump` N times with different targets, merge results into a new `PackageMeta.variants: Vec<SourceVariant>` field. No raw-text parsing of Nickel, no string-interpolation edge cases (which are a real problem — `%{version}` braces defeat naive regex approaches).
- **Right layering**: Nickel is typed and the source URLs are concrete after evaluation. The only variable is which target the evaluator picks. Telling the evaluator "use this target instead" is structurally honest; re-parsing the source text to recover the other arms fights against the type system.

## Alternative considered

A supply-chain-side textual parser that re-reads `build.ncl` and extracts per-arch URLs/sha256 directly. Working prototype in progress but the approach is brittle because:

- Nickel string interpolation `%{var}` embeds `{` and `}` inside string literals, breaking naive brace-depth tracking unless the parser is string-aware.
- Three distinct textual shapes are in use today (inline string literal, top-level `let amd64_sha256 = "..."`, record field `variants.amd64_sha256`), each needing separate resolution logic.
- Gets out of sync if the build.ncl convention evolves.

The textual parser is acceptable as a stopgap but should be replaced as soon as `minimal dump --target` exists.

## Package count affected

- **Primary blind spot (~10 packages):** claude-code, go, deno, bun, gcloud, actions-runner, varlock, libsodium, mermaid-cli + one or two more. These are where pkgmgr would benefit most.
- **Secondary (~29 `replace_on_cycle` packages):** not affected by this feature — those need prebuilt regeneration tooling, a separate track.

## Acceptance

- `minimal dump --packages --target amd64` (or env-var equivalent) produces valid JSON with `url`/`sha256` reflecting the amd64 arm of any `match { arch = ... } target` expressions.
- Same for `arm64`.
- Running without the flag preserves existing host-target behavior.
- `minimal check` and other consumers of Nickel evaluation are unaffected.

bryan-minimal:
## Summary

Closes #125. Adds a `--arch` flag to `minimal dump` so downstream tools can evaluate arch-matched packages for both architectures.

```
minimal dump --packages --arch amd64
minimal dump --packages --arch arm64
```

When `--arch` is absent, defaults to `Target::host()` (preserves current behavior).

## Why

`mctx::graph_from_all_packages_with_target(target)` has existed for a while but `cmd_dump` was hard-coded to `Target::host()`. For packages that use `match { arch = 'Amd64 } => ..., { arch = 'Arm64 } => ... } target` to select between per-arch source URLs (claude-code, go, deno, gcloud, etc.), this meant only the host arm's data surfaced in the dump — the other arm was erased at evaluation time.

That blind spot cascades:
- pkgmgr silently corrupts arch-matched packages on update (skipping them via a guardrail in its current form, once it can see both it can update both)
- supply-chain's scanner only sees one arch's resolved source URL per package

Exposing the target at the CLI lets consumers run the dump twice and aggregate.

## Design notes

- Follows `cmd_materialize`'s existing `--arch` pattern verbatim: accepts `amd64`/`x86_64` → `Arch::Amd64`, `arm64`/`aarch64` → `Arch::Arm64`. Rejects anything else with an error message naming the bad input and listing valid options.
- OS is pinned to `OS::Linux`. Dump consumers want per-arch evaluation of the same pkgs repo, not cross-OS targets. Easy to extend later if a cross-OS case emerges.
- `resolve_target` is factored out so it's unit-testable without spinning up a `Context`. Extracted into a private helper matching the codebase's existing style of small helpers colocated with their command.

## Test plan
- [x] New unit tests in `cmd_dump::tests` cover the four paths: default (host), amd64 aliases, arm64 aliases, unknown-arch error
- [x] `minimal test` — full suite passes, including the new cmd_dump tests
- [x] `minimal run fmt-check` — clean
- [x] `minimal run clippy` — clean (no new warnings)

## Follow-ups (not this PR)

- pkgmgr's guardrail on multi-arch packages can now be lifted: supply-chain reads the dump for each arch, aggregates the per-arch `sha256`/`url` data, and pkgmgr rewrites all arms in one pass.
- supply-chain's scanner can use the per-arch output to catch arch-specific advisories.

### DIFF
--- a/crates/minimal/src/cmd_dump.rs
+++ b/crates/minimal/src/cmd_dump.rs
@@ -1,6 +1,9 @@
 use std::{fmt::Display, io::stdout};
 
+use anyhow::anyhow;
+use common::Target;
 use common::fuzzy_search::fuzzy_match;
+use common::target::{Arch, OS};
 use decode::AttrValue;
 use graph::{BuildDep, BuildOutput, BuildSpec, BuildSpecRef, Graph, RuntimeDep, SourceInput};
 use mctx::{Cache, Context, Error};
@@ -17,6 +20,15 @@ pub struct DumpArgs {
 
     #[arg(long, default_value_t = false)]
     pub exact: bool,
+
+    /// Target architecture to evaluate packages against (e.g., "amd64",
+    /// "arm64"). When a package's `build_deps` uses
+    /// `match { arch = 'Amd64 } => ..., { arch = 'Arm64 } => ... } target`
+    /// to select between per-arch source URLs, only the matching arm's
+    /// output appears in the dump. Overrides the host default so tools
+    /// downstream can aggregate both arches by running the dump twice.
+    #[arg(long)]
+    pub arch: Option<String>,
 
     pub name: Option<String>,
 }
@@ -120,7 +132,8 @@ impl From<&AttrValue> for PkgAttr {
 pub async fn cmd_dump(args: DumpArgs, ctx: &mut Context) -> Result<(), Error> {
     let mut out_packages = Vec::new();
 
-    let graph = ctx.graph_from_all_packages()?;
+    let target = resolve_target(args.arch.as_deref())?;
+    let graph = ctx.graph_from_all_packages_with_target(target)?;
     let cache = ctx.local_cache();
 
     if args.kind.packages {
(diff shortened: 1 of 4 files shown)


crates/minimal/src/cmd_dump.rs
```diff
@@ -219,3 +232,65 @@ fn pkg_ref_from_input(g: &Graph, i: &BuildDep, _c: &Cache) -> Result<PkgRef, Err
         _ => todo!("bsi variant {:?}", i),
     }
 }
+
+/// Resolve the `--arch` flag into a [`Target`], matching `cmd_materialize`'s
+/// parsing (arm64/aarch64 → Arm64, amd64/x86_64 → Amd64). When the flag is
+/// absent, defaults to [`Target::host`]. OS is always Linux — dump consumers
+/// want per-arch evaluation of the same pkgs repo, not cross-OS.
+fn resolve_target(arch: Option<&str>) -> Result<Target, Error> {
```
twitchyliquid64, line 241: Make or use an existing method on `Target` or `Arch`: the generic problem of going from an `&str` into the specific type is something we are bound to use half a dozen times in various places.

bryan-minimal:
Good call on the shared helper — pushed 5aca2a3b: added FromStr for Arch in common with full alias set and tests, migrated cmd_dump and cmd_materialize to str.parse. The arch parsing duplication in cmd_materialize is cleaned up as a drive-by since the same point applied there.

twitchyliquid64 (commented):
Lastly, update the other place where we go from a str to a `Target`, in the nickel-parsing code.

Idk why i thought looping over all representations was a good idea at the time lol

https://github.com/gominimal/minimal/blob/3109695440eeb00df3bcfff4d3fe8679d1931c94/crates/decode/src/builds.rs#L910-L916

bryan-minimal:
Pushed 0926f001: added FromStr for Target (and OS as a building block) and migrated the ad-hoc Target::all() loop in decode/src/builds.rs. Also fixed the rustfmt diff CI flagged on cmd_materialize — chain layout preference. The new Target::FromStr has a roundtrip test over every Target::all() value so that impl and as_ref stay locked together.

twitchyliquid64 (approved):

PR #147, opened 2026-04-20
crates/{check,decode}: support running check across packages/ which are partitioned

twitchyliquid64:
I.e. can check a layer using the scheme `packages/a/abseil/build.ncl` instead of only `packages/abseil/build.ncl`.


bryan-minimal (approved):
LGTM

PR #148, opened 2026-04-22
crates/check: refactor, add check for overzealous `../` imports

twitchyliquid64:


crates/check/src/lib.rs
```diff
@@ -1023,24 +906,25 @@ static ADJACENT_IMPORT_REGEX: LazyLock<Regex> = LazyLock::new(|| {
         .expect("Invalid regex pattern")
 });
 
-struct AdjacentImportCheck;
+static ALL_IMPORT_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r#"\s+import\s+"(.+)""#).expect("Invalid regex pattern"));
```
bryan-minimal: The `(.+)` is greedy and will merge multiple imports on one line into a single 'path'. For example, on

```nickel
let a = import "../base/build.ncl" in import "../make/build.ncl"
```

the regex engine matches from the first `"` through to the LAST `"`, capturing `../base/build.ncl" in import "../make/build.ncl` as a single `import_path`. Joining that with `pkg_dir` + absolutizing produces a nonsense path that often does escape the base dir → false positive on any file with two imports on one line.

Suggested fix:

```rust
Regex::new(r#"\s+import\s+"([^"]+)""#)
```

`[^"]+` is both stricter (paths can't contain `"` anyway) and immune to greedy merging across adjacent matches. Alternatively `(.+?)` for non-greedy, but `[^"]+` is semantically better.

(Minor related nit: the leading `\s+` means `import` at the absolute start of a file — no leading whitespace at all — won't match. Probably a non-issue in practice since Nickel files never start with a bare `import`, but `(^|\s)` would cover it if you want to be thorough.)

crates/check/src/lib.rs
```diff
@@ -1023,24 +906,25 @@ static ADJACENT_IMPORT_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
+        ctx: &CheckCtx,
         _pkg: &str,
         pkg_dir: &Path,
-        _stdlib_dir: &Path,
         _ot: Option<OpTracker>,
     ) -> Result<CheckResult, Error> {
         let mut result = CheckResult {
             verdict: CheckVerdict::Pass,
-            check: "adjacent import",
+            check: "imports",
```
bryan-minimal: Silent rename: the check id moved from `"adjacent import"` to `"imports"`. Anyone whose CI or local invocation currently passes `--skip-checkers "adjacent import"` will need to update. Two options depending on how strict you want to be:

1. Mention in the commit body (e.g. `NOTE: --skip-checkers "adjacent import" → "imports"`) and let users discover it on upgrade.
2. Accept either string here for one release cycle, then drop `"adjacent import"`:

   ```rust
   if ctx.skip_checkers.iter().any(|s| s == "imports" || s == "adjacent import") {
   ```

Not a blocker either way; flagging so the rename is at least intentional.

twitchyliquid64: Should be okay, no one is using this lol

crates/check/src/lib.rs
twitchyliquid64: Done.

bryan-minimal (approved):
LGTM

PR #152, opened 2026-04-22
crates/sandbox2: catch panics in env channel handler to avoid socket permanently dying

twitchyliquid64:
Defense-in-depth against the socket channel that powers `min` being permanently borked when something unexpected happens, which seems to be happening a bunch lately.


msample (approved):
LGTM

PR #155, opened 2026-04-24
crates/rcache: split reader/writer, use compare-and-swap on index

bryan-minimal:
## Summary

Splits `rcache::RemoteCache` into a read-only type and a new GCS-only `RemoteCacheWriter` that uses `if_generation_match` compare-and-swap on `index.shisha`. Fixes the lost-update race that caused the openssl 3.6.2 incident — concurrent writers were silently dropping each other's index entries.

## What was happening

`finish_uploads` did unsynchronized read-modify-write: each writer fetched the index at run start, accumulated `(spec_hash, sha256)` entries during the build, then overwrote the file with `their snapshot + their entries` on commit. With `max_concurrent_builds=2`, two writers committing in overlapping windows would each clobber the other:

- t=0: index_v0 = {A}
- t=1: writer 1 fetches v0; pending = {B}
- t=2: writer 2 fetches v0; pending = {C}
- t=3: writer 2 commits → index_v1 = {A, C}
- t=4: writer 1 commits **based on v0 snapshot** → index_v2 = {A, B}, dropping C

The blob objects stayed in GCS — only the index entries pointing at them got lost. Yesterday: PR #99 (openssl 3.6.2) committed openssl + ~25 rebuilt transitives; a smaller concurrent run committed on top of the pre-99 snapshot, dropping all the openssl-keyed entries. Users hit `NotFound` on every consumer and paid full local-rebuild cost.

## The fix

**Bifurcation, not just a generation field.** Reason: keeping local-cache + writer in the same type is a footgun — a writer holding a cached-from-disk index has no real generation to use, and writing with `if_generation_match=0` would silently overwrite if the index didn't exist OR if we got lucky on timing. Two types make the constraint type-safe.

- **`RemoteCache<B>`** — reader. Local-cache fast path (5min TTL) preserved. Used by `mctx`, `res-server`.
- **`RemoteCacheWriter`** — GCS-only. Always fresh-fetches the index at construction, records the generation it observed. On `finish_uploads`:
  1. Merge pending into fetched index.
  2. Write with `set_if_generation_match(generation)`.
  3. On 412: refetch (new index + generation), re-merge our pending, retry.
  4. Bounded to 8 retries, exponential backoff 200ms→6.4s (~25s total cap), with jitter.

Idempotency falls out for free: `BTreeMap::extend` with the same `(spec_hash, sha256)` is a no-op, so concurrent writers committing the same entry both succeed without producing inconsistent state.

## Test plan

14 new unit tests in `remote_writer::tests`, all passing:

**Retry policy** (6) — exhaustive table:
- 412 first attempt → 200ms backoff
- 412 second → 400ms (doubling)
- 412 attempts 5/7 → capped at 6.4s
- 412 at MAX_RETRIES → GiveUp
- non-412 status (404/500/403) → GiveUp immediately
- **no status code (transport errors) → GiveUp** (caught a real bug during refactor where `None` was being mapped to `Done`)

**Merge invariant** (4):
- Empty pending → clones fetched
- Empty fetched → uses pending
- Repeated entries → idempotent
- Pending overrides fetched for same `spec_hash`

**Race simulation** (2) — directly model the lost-update bug:
- Writers W1+W2 race over base v0; all three entries (A, B, C) survive
- 10-writer chained contention; all 10 entries survive

Both race tests would FAIL under the old `index.extend(uploaded)` design.

**Jitter** (2): in-range + zero-edge-case.

```
running 14 tests
test remote_writer::tests::race_simulation_loses_no_entries ... ok
test remote_writer::tests::race_simulation_many_writers_all_survive ... ok
... (12 more)
test result: ok. 14 passed; 0 failed
```

`cargo clippy --workspace --all-targets` clean; `cargo fmt` clean.

## Callers updated

- `crates/minimal/src/cmd_upload_cache.rs` — switched to `RemoteCacheWriter` via new `Context::remote_cache_writer()`.
- (downstream) `gominimal/build-servers` `crates/buildbot/src/orchestrate.rs` — separate PR after this merges + the rev pin gets bumped.

Reader call sites unchanged. `RemoteCache::new_with_gcs_bucket` preserved for read-only GCS use.

## Out of scope (follow-ups)

- Live-GCS integration test (needs a bucket + creds; current test suite proves the merge invariant + retry policy at unit level).
- Integrity-canary endpoint that compares blob set vs index entries to catch the next class of weird drift in <5min instead of a Slack report.

### DIFF
--- a/crates/rcache/src/remote_writer.rs
+++ b/crates/rcache/src/remote_writer.rs
@@ -1,0 +1,520 @@
+//! GCS-specific writer for the shared cache that uses optimistic concurrency
+//! to safely merge index updates from multiple concurrent writers.
+//!
+//! See the historical note at the top of `remote.rs` for why this is split
+//! off from [crate::RemoteCache]: the previous design did an unsynchronized
+//! read-modify-write on `index.shisha` which lost entries when two writers
+//! raced.
+
+use crate::INDEX_FILENAME;
+use crate::remote::Error;
+use crate::remote_index::RemoteIndex;
+use common::SpecHash;
+use common::fetchers::{FetchUrl, GcsUrl};
+use google_cloud_storage::{Error as GcsError, client::Storage};
+use ot::OpTracker;
+use std::time::Duration;
+use tokio::time::sleep;
+
+/// Maximum retries on `if_generation_match` contention before giving up.
+/// With the backoff schedule below, 8 retries cap total contention-wait at
+/// ~25s (200+400+800+1600+3200+6400+6400+6400 ms). After that, the
+/// underlying GCS error is bubbled up and the run fails. Total writes
+/// attempted is `MAX_INDEX_WRITE_RETRIES + 1` (the initial attempt).
+const MAX_INDEX_WRITE_RETRIES: u32 = 8;
+
+/// Initial backoff between retries; doubled each attempt up to a cap.
+const INITIAL_RETRY_BACKOFF_MS: u64 = 100;
+
+/// HTTP status returned by GCS when an `if_generation_match` precondition
+/// fails (i.e. the object's generation has changed since we read it).
+const PRECONDITION_FAILED: u16 = 412;
+
+/// HTTP status returned by GCS when reading a non-existent object.
+const NOT_FOUND: u16 = 404;
+
+/// A writer for the shared cache that uses optimistic concurrency on the
+/// index file's GCS generation.
+///
+/// Construct with [Self::new], call [Self::upload] for each artifact, then
+/// [Self::finish_uploads] to commit the merged index. On contention with
+/// another writer (412 Precondition Failed), refetches the current index
+/// and retries with bounded exponential backoff, so the new write always
+/// includes both our pending entries AND any entries committed by the
+/// other writer in the meantime.
+///
+/// Not `Clone` on purpose — cloning the pending list and committing twice
+/// would either redo work or upload divergent state. Construct one writer
+/// per upload session and consume it via [Self::finish_uploads].
+#[derive(Debug)]
+pub struct RemoteCacheWriter {
+    backend: Storage,
+    base: GcsUrl,
+
+    /// The index as it was when we last fetched it from GCS. Used both as the
+    /// base to merge `pending` into, and as a dedup check in [Self::upload]
+    /// to skip uploads of artifacts already present.
+    fetched_index: RemoteIndex,
+
+    /// The GCS object generation that `fetched_index` was loaded from.
+    /// `0` means the index file did not exist at fetch time, which
+    /// also doubles as the "create-if-not-exists" precondition value
+    /// for `set_if_generation_match`.
+    fetched_generation: i64,
+
+    /// `(spec_hash, sha256)` entries to merge into the index on finish.
+    pending: Vec<(SpecHash, [u8; 32])>,
+
+    #[allow(dead_code)]
+    ot: Option<OpTracker>,
+}
+
+impl RemoteCacheWriter {
+    /// Initialize a writer against the given GCS bucket. Always fetches the
+    /// current index fresh from GCS — there's no local-cache fast path,
+    /// because the recorded generation must match the actual cloud state
+    /// for the compare-and-swap on commit to be meaningful.
+    pub async fn new(
+        storage: Storage,
+        bucket_id: &str,
+        ot: Option<OpTracker>,
+    ) -> Result<Self, Error<GcsError>> {
+        let base = GcsUrl {
+            bucket: format!("projects/_/buckets/{bucket_id}"),
+            object: "".to_string(),
+        };
+
+        let (fetched_index, fetched_generation) = fetch_index(&storage, &base).await?;
+
+        Ok(Self {
+            backend: storage,
+            base,
+            fetched_index,
+            fetched_generation,
+            pending: Vec::new(),
+            ot,
+        })
+    }
+
+    /// Upserts the given artifact to the GCS bucket, staging it for inclusion
+    /// in the index. Returns `false` if the given artifact's `(spec_hash, sha256)`
+    /// pair is already in the fetched index (and the upload was skipped).
+    ///
+    /// The artifact blob upload is content-addressed, so concurrent uploads
+    /// of the same blob bytes are safe (last-writer-wins on identical content).
+    /// The index update is deferred to [Self::finish_uploads].
+    pub async fn upload(
+        &mut self,
+        spec_hash: &SpecHash,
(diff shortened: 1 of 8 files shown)


crates/rcache/src/remote_writer.rs
```diff
@@ -0,0 +1,520 @@
...
+
+/// Maximum retries on `if_generation_match` contention before giving up.
+/// With the backoff schedule below, 8 retries cap total contention-wait at
+/// ~25s (200+400+800+1600+3200+6400+6400+6400 ms). After that, the
+/// underlying GCS error is bubbled up and the run fails. Total writes
+/// attempted is `MAX_INDEX_WRITE_RETRIES + 1` (the initial attempt).
+const MAX_INDEX_WRITE_RETRIES: u32 = 8;
+
+/// Initial backoff between retries; doubled each attempt up to a cap.
+const INITIAL_RETRY_BACKOFF_MS: u64 = 100;
```
twitchyliquid64: With a default max-age of 300, our retries need to at least span that. Start at 2s?

crates/rcache/src/remote_writer.rs
```diff
@@ -0,0 +1,520 @@
...
+    let mut merged = fetched.clone();
+    merged.extend(pending.iter().cloned());
+    merged
+}
+
+/// Fetch the index file from GCS, returning both its parsed content and the
+/// GCS generation it was loaded from. Returns `(default, 0)` if the index
+/// doesn't exist yet — `0` doubles as the "create-if-not-exists" precondition
+/// value for subsequent `if_generation_match` writes.
+async fn fetch_index(
```
twitchyliquid64: Can we use the same code for fetching the index for RemoteCache as here for RemoteCacheWriter? also, theres no need to fetch it again if you need a RemoteIndex as part of the construction of a RemoteCacheWriter: We should keep track of the generation in RemoteCache too.

crates/rcache/src/remote_writer.rs
bryan-minimal: Done in febadd2c. Bumped initial backoff from 100ms to 2s, so the schedule across 8 retries is now 4s, 8s, 16s, 32s, 64s, then capped at 128s for the last three (about 508s total). That spans the 300s cache window with margin. Tests updated to match.

crates/rcache/src/remote_writer.rs
bryan-minimal: Done in febadd2c, both parts.

For shared fetch, extracted `fetch_gcs_index` as `pub(crate)` in `remote_writer.rs` and added a `generation()` method on the `FetchResponse` trait. Default impl returns `None`, the GCS impl returns `Some(response.object().generation)`, and Reqwest stays default. The existing `RemoteCache::new` picks up the value through the trait so non-GCS readers are unchanged.

For tracking generation in `RemoteCache`, added a `gcs_generation: Option<i64>` field. Populated from `index_resp.generation()` on fresh fetches, set to `None` on the local-cache fast path (since we never asked GCS what generation we have). Then added `RemoteCache<Storage>::into_writer` that reuses the index plus generation when `Some`, falls back to `fetch_gcs_index` when `None`. The local-cache fallback still hits GCS once because the cached file does not carry a generation, but that only triggers for a reader that already accepted the 5min stale window for reads.

crates/rcache/src/remote_writer.rs
```diff
@@ -0,0 +1,553 @@
...
+            assert!(j < 100, "jitter {j} should be in [0, 100)");
+        }
+    }
+
+    #[test]
+    fn jitter_zero_does_not_panic() {
+        // The .max(1) guard prevents modulo-by-zero. Result is 0.
+        assert_eq!(jitter_ms(0), 0);
+    }
+}
```
twitchyliquid64: The `jitter_ms` function is 5 lines, these arent really useful

https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

crates/rcache/src/remote_writer.rs
```diff
@@ -0,0 +1,553 @@
...
+            decide_retry_after_failure(None, 0),
+            RetryAfterFailure::GiveUp
+        );
+        assert_eq!(
+            decide_retry_after_failure(None, 5),
+            RetryAfterFailure::GiveUp
+        );
+    }
+
+    // --- merge_for_commit + RemoteIndex behavior ---
```
twitchyliquid64: The code under test is effectively `BTreeMap::extend` - which has these semantics, so these tests arent really testing anything.

A useful test would be something that faked the GCS API and imitated these conditions, because then the code under test would be much more substantial, and using the public API would be more stable to change than the simple internal methods being used here.

twitchyliquid64 (approved):
Up to you if you want to do another round or just get it in as-is, i think ideally we could flesh out the tests a bit more but we can always do that later.

crates/rcache/src/remote_writer.rs
bryan-minimal: Deleted in d7fa2485. The article makes the case clearly, both `jitter_*` tests removed.

crates/rcache/src/remote_writer.rs
bryan-minimal: Removed all four `merge_*` tests and both `race_simulation_*` tests in d7fa2485. They were exercising `BTreeMap::extend` semantics rather than the writer logic.

Your point about needing mocks for the real coverage is right, and the new code in this PR is small enough (store an int, add the `if_generation_match` arg) that the mock-GCS test infrastructure would be disproportionate. The plan is to lean on the buildbot integration once #155 plus the build-servers PR land, since concurrent PRs naturally hit the contention path against staging.

Also consolidated the `decide_retry_after_failure` tests from 6 down to 3 covering the distinct branches (exponential backoff with cap, MAX_RETRIES bound, non-412 errors skip retries). Kept those because the policy table is the contract for retry behavior, but happy to drop them too if you think they fall on the change-detector side.

bryan-minimal:
Pushed b925927d. Went the `mockall` plus `gcs::stub::Storage` route since the GCS crate exposes the trait specifically for this. The writer is now generic over `S: stub::Storage = stub::DefaultStorage` (default keeps the production API unchanged) and the four new tests inject a `MockStorage` via `Storage::from_stub` to exercise the CAS retry loop end-to-end through the public `finish_uploads` API.

Tests assert on `if_generation_match` in the recorded write requests, so the retry behavior is verified by the actual contract under contention rather than by inspecting internal helpers. `tokio::test(start_paused = true)` makes the multi-second backoff sleeps auto-advance, suite still finishes in 0.00s.

Replaces the 3 retry-policy tests from the previous fixup since the new behavior tests cover the same ground through the public API.

PR #158, opened 2026-04-24
crates/rcache: verify sha256 against bincache index, add tests

twitchyliquid64:
Whelp. Looks like a refactor a while ago stopped using `FileCache` to fetch the artifact tarballs, and `FileCache` was responsible for checking the sha's.

This change restores sha checking, as well as adds adversarial tests to all the paths where we fetch artifacts that need hash checking.

Lets get this one in sooner rather than later.

## Summary by CodeRabbit

* **Improvements**
  * Remote cache now validates downloaded data integrity using hash verification
  * Corrupted downloads are automatically removed from the filesystem
  * Validation failures report both expected and computed hash values for diagnostics

* **Tests**
  * Added unit tests for hash mismatch detection scenarios

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3903,6 +3903,7 @@ dependencies = [
  "mockall",
  "ot",
  "reqwest",
+ "sha2 0.11.0",
  "tempfile",
  "tokio",
  "tracing",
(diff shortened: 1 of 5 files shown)


bryan-minimal (approved):
LGTM

PR #162, opened 2026-04-27
crates/decode: fix type error when subsetting an upstream

twitchyliquid64:
Fixes MIN-208

 * Nickel types now permit a subset defined from an upstream declaration
 * Parsing code was already correct, added test

## Summary by CodeRabbit

* **New Features**
  * Extended subset dependency support to allow upstream packages as subset targets, in addition to build specifications.

* **Tests**
  * Added validation test for subset dependency decoding with upstream packages.


bryan-minimal (approved):
LGTM

PR #163, opened 2026-04-27
crates/mctx: error min install with helpful error when file patching is required

twitchyliquid64:
Fixes: Min-178

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer now detects when host file patching is required and exits with a clear multi-line error plus restart instructions instead of attempting an in-place install.
* **Improvements**
  * Installations are performed against an isolated installation snapshot to avoid mutating the current session.
  * Cached dependency handling now reads from the installation snapshot for more reliable hardlinking and commits the snapshot on success.

### DIFF
--- a/crates/mctx/src/env.rs
+++ b/crates/mctx/src/env.rs
@@ -46,32 +46,46 @@ impl EnvChannel<'_> {
             return; // Already installed.
         }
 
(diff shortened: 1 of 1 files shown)


bryan-minimal (approved):
LGTM

PR #167, opened 2026-04-27
crates/mctx: fix CLAUDE.md not being editable

twitchyliquid64:
Fixes MIN-166

## Summary by CodeRabbit

* **Refactor**
  * Improved handling of local configuration files during sandboxed runs: temporary modifications are now applied only when necessary, cleanup is deterministic, injected helper instructions are removed during teardown, and any user edits are preserved. If a caller already provides the real config via filesystem mappings, no temporary injection or teardown occurs.
* **Style**
  * Suppressed a noisy lint warning to reduce build diagnostics without changing runtime behavior.

### DIFF
--- a/crates/mctx/src/env.rs
+++ b/crates/mctx/src/env.rs
@@ -499,29 +499,90 @@ pub struct EnvArgs<'a> {
     pub ot: Option<OpTracker>,
 }
 
+/// Tracks a CLAUDE.md that was patched with sandbox instructions, so we can
+/// write back user changes and strip the injected content on teardown.
+struct ClaudeMdPatch {
+    /// The original file content (empty if the file didn't exist or was empty).
+    original_content: Vec<u8>,
+    /// Whether CLAUDE.md existed on disk before we patched it.
+    already_existed: bool,
+    /// Path to the temp file that is bind-mounted into the sandbox.
+    temp_path: PathBuf,
+    /// Path to the real CLAUDE.md in the project directory.
+    real_path: PathBuf,
+}
+
+impl ClaudeMdPatch {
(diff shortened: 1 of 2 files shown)


crates/mctx/src/env.rs
```diff
@@ -499,29 +499,88 @@ pub struct EnvArgs<'a> {
...
+            Ok(s) => s,
+            Err(_) => return,
+        };
+
+        let cleaned = content.replace(LLM_APPEND_INSTRUCTIONS, "");
+        let cleaned_bytes = cleaned.as_bytes();
+
+        if cleaned_bytes == self.original_content {
+            // No user changes. Clean up if the file didn't exist before.
+            if self.original_content.is_empty() {
```
msample: what about if the original file existed but was empty? we'd show a git rm ?

crates/mctx/src/env.rs
```diff
@@ -499,29 +499,88 @@ pub struct EnvArgs<'a> {
...
+            Ok(c) => c,
+            Err(_) => return,
+        };
+        let content = match std::str::from_utf8(&content) {
+            Ok(s) => s,
+            Err(_) => return,
+        };
+
+        let cleaned = content.replace(LLM_APPEND_INSTRUCTIONS, "");
+        let cleaned_bytes = cleaned.as_bytes();
```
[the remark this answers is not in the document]
twitchyliquid64: Thats really not an issue

crates/mctx/src/env.rs
twitchyliquid64: Yes - an empty `CLAUDE.md` file is useless, so we'll delete in this case.

msample (approved):
LGTM

PR #170, opened 2026-04-28
crates/check: single-flight expensive I/O, add semaphore

twitchyliquid64:
## Summary by CodeRabbit

* **Performance Improvements**
  * Glob pattern matching results are now cached to reduce computation time during file discovery operations
  * Concurrent processing for package, profile, and harness validation with managed resource throttling has been added
  * Parse validation is optimized to skip redundant processing steps when dependency graph information is already available

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -457,6 +457,7 @@ dependencies = [
  "futures",
  "graph",
  "lcache",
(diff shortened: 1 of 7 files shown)


msample (approved):
LGTM

PR #176, opened 2026-05-01
crates/mctx (mostly): Refactored mfile search in mctx into an enum with multiple search strategies for different scenarios and fixed a related bug in `minimal init`

evanspearman:
## Bugfixes
- The `init` command will no longer look in ancestor directories for an mfile and will just search the current working directory or the override if `-C` is specified

## Refactored finding MFiles
The logic for determining the strategy for finding MFiles was previously hard-coded in `Context::new`. Because we needed different behaviour for `init` vs other commands, this needed to change. I refactored the logic out into an enum called `MFileSearchStrategy` which is used to specify how to find the mfile, allowing individual commands to specify this behaviour when they build their `Context` struct.

The options are:
- `OverrideWithFallback` - previous behaviour and now the default behaviour. Checks a given directory first and falls back to recursively searching ancestors of the current directory
- `OverrideOnly` - Only check the specified override directory (not actually used but included because it took a few seconds)
- `CurrentDirOnly` - Only check the current directory. This is the desired behaviour for the `init` command
- `CurrentDirRecursive` - The fallback behaviour for `OverrideWithFallback`

## Misc. Changes
- Added serial_test as a dev dependency. This was required to run tests that need to change the current working directory serially to avoid race conditions. The crate has over 100 million downloads, is MIT licensed, and is actively maintained with the last release being only 2 months ago
- Changed the `mctx` `Config` build pattern methods that take  `impl Into<PathBuf>` instead of just straight `PathBuf`. This improves the ergonomics as it allows you to pass a `&Path` while preserving the old behavior if you pass a `PathBuf`.
- Implemented `From<mfile::Error> for Error` for `mctx` `Error` as it was missing

## Testing

I added tests for the different MFileSearchStrategy behaviours. I also manually tested that the init command follows the new desired behaviour.

## Summary by CodeRabbit

* **New Features**
  * Selectable minimal-file discovery strategies (override, current-dir only, recursive).
  * New context constructor that accepts an explicit search strategy.

* **Improvements**
  * More flexible path inputs and safer override accessors for repo/stdlib paths.
  * Cleaner, more consistent error propagation when locating or loading files and resolving directories.
  * Safer init flow that surfaces failures instead of panicking.

* **Tests**
  * Added isolated tests validating search behaviors and working-directory safety; included test fixtures.

### DIFF
--- a/crates/mctx/src/lib.rs
+++ b/crates/mctx/src/lib.rs
@@ -25,6 +25,9 @@ use graph::{BinProvider, BuildSpecRef, Graph, MaskingBinProvider, Transitives};
 use mfile::{EnvPatches, EnvVarValue, LinkConfig, Task};
 pub use sandbox2::config::Invocation;
 
+mod mfile_search_strategy;
+pub use mfile_search_strategy::MFileSearchStrategy;
+
 pub use env::Env;
 use tokio::sync::Semaphore;
 use toml_edit::{Array, DocumentMut, Item, TableLike, Value};
@@ -224,7 +227,7 @@ impl Context {
         //  - The version embedded in the binary, stamped to disk
         let stdlib_dir = {
             if let Some(dir) = config.stdlib_dir_override() {
-                dir.clone()
+                dir.to_path_buf()
             } else {
                 stdlib::upsert_stdlib_to_disk(config.stdlib_dir()).map_err(|e| {
                     Error::Other(anyhow!("loading embedded standard library: {}", e))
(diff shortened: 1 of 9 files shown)


crates/minimal/src/cmd_init.rs
```diff
@@ -19,63 +19,68 @@ const DEFAULT_PKGS_BRANCH: &str = "main";
...
+                        .repo_dir_override()
+                        .clone()
+                        .unwrap_or_else(|| std::env::current_dir().unwrap()),
+                    None,
+                    Graph::new_from_chain(
+                        vcs,
+                        &mut (),
+                        LinkConfig::Git {
+                            repo: "https://github.com/gominimal/pkgs".to_string(),
+                            branch: Some("main".to_string()),
```
twitchyliquid64: While we are here lets dedupe this to the constant also


```suggestion
                            repo: DEFAULT_PKGS.to_string(),
                            branch: Some(DEFAULT_PKGS_BRANCH.to_string()),
```

crates/mctx/src/lib.rs
```diff
@@ -237,21 +240,18 @@ impl Context {
...
+        let strategy = {
+            match config.repo_dir_override().clone() {
+                Some(path) => MFileSearchStrategy::OverrideWithFallback(path),
+                _ => MFileSearchStrategy::CurrentDirRecursive,
             }
         };
+        Self::new_with_strategy(config, strategy)
+    }
 
+    pub fn new_with_strategy(config: Config, strategy: MFileSearchStrategy) -> Result<Self, Error> {
```
twitchyliquid64, line 254: Doc comment for this guy? just cuz `mctx::Context` methods are the closest thing we have to an API layer for the internals

twitchyliquid64 (approved):

twitchyliquid64 (approved):

PR #192, opened 2026-05-07
all: rename harness to stack

twitchyliquid64:
* For now `[harness]` remains supported in the `minimal.toml`, but will be updated to `[stack]` on next `minimal update`
 * For now `harnesses/<name>/harness.ncl` remains supported in layers, with no change in syntax
 * Layer loading code + check code consults either `harnesses/` or `stacks/`.
 * Minimal standard library now has a `stack` function to construct a stack, identical to the `harness` function we will eventually remove
 * A zillion references to harnesses in the code, error types, and print strings have been updated
 * Added a version int we can increment every time we meaningfully break the layer format, so the hash that keys the layer cache can be effectively invalidated

## Summary by CodeRabbit

* **New Features**
  * Introduced stack-based project configuration and checking system replacing harnesses.

* **Updates**
  * Added --stacks CLI option and updated commands to reference "stack" in status and help text.
  * Initialization, checking, dumping, and update flows now prefer stacks/ layout and migrate harness→stack keys when needed.
  * Task resolution and generated configs now use stack-scoped package lists.

* **Chores**
  * Bumped stdlib and workspace component versions.


twitchyliquid64:
`build-in-minimal` failure is expected as I updated the repo `minimal.toml` to a new version.

norrietaylor (approved):
🚢

PR #198, opened 2026-05-07
crates/minimal: add materialize type raw-file

twitchyliquid64:
This implements a new materialize type `raw-file`, which lets us pull a file out of the minimal universe. This is intended to be used with for instance the [virtio-linux](https://github.com/gominimal/pkgs/pull/154) package, which gives you a linux image for your microVM.

Now that we have more than one output type, I've also refactored the serde deserialize code so that only valid output configurations are representiable, namely by putting output-specific fields in their corresponding enum variant.

## Summary by CodeRabbit

* **New Features**
  * Added a `raw-file` output type to materialize individual files from built packages.
  * OCI-image outputs now materialize with configured entrypoint, command, and environment variables.
  * CLI `--arch` flag now takes precedence when selecting the target architecture.

* **Improvements**
  * Output configs are validated per output type, rejecting invalid field combinations and returning clearer errors.

### DIFF
--- a/crates/mfile/src/lib.rs
+++ b/crates/mfile/src/lib.rs
@@ -320,28 +320,77 @@ pub struct Stack {
 
 impl Eq for Stack {}
 
-/// Describes the specific type of output being generated.
-#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
+/// Describes the specific type of output being generated, along with any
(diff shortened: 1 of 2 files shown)


jtnkminimal:
Claude agreed that this should be fine to approve  :+1:

jtnkminimal (approved):

PR #202, opened 2026-05-08
Add --offline flag covering source-tarball + git fetches

bryan-minimal:
## Summary

Adds a new `--offline` global flag for running minimal in network-isolated environments where every input is pre-staged in the local cache. On cache miss, surfaces a clear error rather than attempting any network call.

`--no-fetch` semantics unchanged — keeps its existing meaning of "skip the remote artifact cache, build from source instead", which composes naturally with `--rebuild` for the laptop-rebuild-the-world use case.

## Motivation

Running `minimal package <pkg>` inside a network-isolated build VM (Confidential Space, no internet egress) where source bytes are pre-staged in a sha-pinned mirror. There was no flag combination that meant "use cache only — if anything's missing, fail loud" — `--no-fetch` only affected the remote artifact cache, while source URL fetches and git clones quietly hit the network on cache miss.

After this patch, the same invocation surfaces:

```
Error: failed to fetch upstream layer: offline cache miss for git remote
https://github.com/gominimal/pkgs — --no-fetch is set; pre-populate the
vcs/ cache or remove the flag
```

## Design

Per review feedback (thanks @msample, @twitchyliquid64) — extending `--no-fetch` would have conflated two distinct semantics. Cleaner shape:

- `--no-fetch` (existing): skip the remote artifact cache. Source/VCS fetches still happen on cache miss. Useful for laptop rebuild-the-world flows alongside `--rebuild`.
- `--offline` (new): treat every cache as authoritative. Source / VCS / artifact-cache misses all error rather than fetch. Implies the artifact-cache-skip half of `--no-fetch` (you can't reach the artifact cache when offline anyway), so `--offline` alone is sufficient.
- Orthogonal to `--no-cache` and `--rebuild`: `--offline` doesn't force a rebuild, it just gates the network.

## Changes

**`crates/common/src/file_cache.rs`**
- `FileCache` gains `offline: bool` + `new_with_offline(dir, offline)` ctor + `is_offline()` accessor. Existing `new(dir)` preserved (defaults to `offline=false`).
- `CachingDownloader::download` checks `cache.offline` after `read_cached` returns `None`; returns `FileCacheError::OfflineCacheMiss { sha256, filename }` instead of falling through to the network fetcher.

**`crates/common/src/remote_storage.rs`**
- `RemoteStorage::new_with_offline(dir, with_gcloud_auth, offline)` forwards `offline` into `FileCache`. Default `new` unchanged.

**`crates/checkouts/src/lib.rs`**
- `Manager` gains `offline: bool` + `new_in_dir_with_offline`. Existing `new_in_dir` preserved.
- `update()`: no-op when offline (caller intent is "use what's there"; refresh-of-known-remote is non-destructive to skip).
- `checkout_of()` known-remote path: skip `repo.fetch()` when offline.
- `checkout_of()` unknown-remote path: return `Error::OfflineCacheMiss { remote }`.

**`crates/checkouts/src/error.rs`**: new `OfflineCacheMiss` variant + `Display`.

**`crates/mctx/src/config.rs`**
- `Config` gains `offline: bool` field. New `with_offline()` builder + `is_offline()` accessor.
- `use_remote_cache()` becomes `!no_fetch && !is_offline()` so `--offline` strictly implies the artifact-cache-skip behavior.

**`crates/mctx/src/lib.rs`**: VcsManager + RemoteStorage initializers pass `config.is_offline()` through to the new `_with_offline` constructors.

**`crates/mctx/src/error.rs`**: maps `checkouts::Error::OfflineCacheMiss` to a clear top-level message.

**`crates/minimal/src/main.rs`**: new `--offline` global CLI flag wired to `ConfigBuilder::with_offline`.

## What this is *not*

- Not a cache-population mechanism. Operators are expected to pre-stage `~/.cache/minimal/downloads/<sha256>/<filename>` and the `vcs/` directory themselves.
- Not a `--trusted-source-domains`-style allowlist (per @msample's prompt). For the offline use case the URL doesn't matter — minimal looks up bytes by sha — but a trusted-domains flag for semi-online flows would be a reasonable separate addition.

## Test plan
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test` — full test suite passes
- [x] Six new unit tests cover the offline path:
  - `file_cache::tests::offline_cache_miss_errors_without_fetching` — `PanickingBackend` asserts the network path is never invoked
  - `file_cache::tests::offline_cache_hit_returns_path_without_fetching` — same assertion under cache-hit
  - `file_cache::tests::online_cache_miss_invokes_backend` — sanity check that default behavior is unchanged
  - `checkouts::tests::offline_unknown_remote_errors`
  - `checkouts::tests::offline_update_is_noop`
  - `checkouts::tests::default_manager_is_online`
- [x] Manual smoke: with the sed source cache wiped, `minimal --offline package -v --rebuild sed` errors with `OfflineCacheMiss` instead of silently re-fetching.

## Future work (out of scope here)

- proptest-style invariant tests across the cache surfaces (minimal doesn't currently use proptest; would be a structural addition worth its own PR)
- cargo-fuzz harness for parser surfaces (Nickel decode, URL parsing, statefile JSON)
- the trusted-source-domains framing @msample raised, for semi-online flows where some hosts are reachable and others aren't

## Summary by CodeRabbit

* **New Features**
  * Global --offline flag and config option; offline-capable managers and cache constructors to run without network access.
* **Behavior**
  * When offline, network/VCS/cache fetches are suppressed and operations return clear offline-cache-miss errors advising cache pre-population or removing --offline.
* **Tests**
  * Added tests for offline vs online cache behavior, checkout/update offline handling, and default online behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/202?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/checkouts/src/lib.rs
+++ b/crates/checkouts/src/lib.rs
@@ -266,15 +266,22 @@ impl Manager {
     }
 
     /// Updates all repos to latest - does nothing for refs which arent symbolic (i.e. commits).
-    /// In offline mode, returns [Error::OfflineCacheMiss] for the first remote we'd
-    /// have to fetch.
+    /// In offline mode, returns [Error::OfflineCacheMiss] — `update` is fundamentally
+    /// a network operation, and silently lameducking it would mask hard-to-debug bugs
+    /// for callers like `minimal update` that explicitly want fresh state.
     pub fn update(&mut self) -> Result<(), Error> {
         if self.offline {
-            // In offline mode, `update` is a no-op for known remotes: we have
-            // no way to refresh, but the caller's intent is "use what's there",
-            // so returning Ok is more useful than erroring on every refresh.
-            // (`checkout_of` still errors when an UNKNOWN remote is requested.)
-            return Ok(());
+            // Pick the first known remote for the error message; if there are
+            // no known remotes there's nothing to update, so a synthetic
+            // placeholder is clearer than a misleading Ok.
+            let remote = self
+                .state
+                .git_remotes
+                .keys()
+                .next()
+                .cloned()
+                .unwrap_or_else(|| "<no remotes>".to_string());
+            return Err(Error::OfflineCacheMiss { remote });
         }
         let checkouts_dir = self.git_checkouts_dir();
         for (_remote, id) in self.state.git_remotes.iter_mut() {
@@ -511,16 +518,18 @@ mod tests {
         }
     }
 
-    /// In offline mode, `update()` is a no-op (returns Ok without touching
-    /// the network). The caller's intent is "use what's there"; refresh-of-
-    /// known-remote is non-destructive to skip.
+    /// In offline mode, `update()` returns OfflineCacheMiss rather than
+    /// silently no-op'ing. Callers like `minimal update` ask explicitly for
+    /// fresh state, so noop-ing would mask hard-to-debug bugs (per review
+    /// from @twitchyliquid64).
     #[test]
-    fn offline_update_is_noop() {
+    fn offline_update_errors() {
         let base_dir = tempdir().unwrap();
         let mut manager = Manager::new_in_dir_with_offline(base_dir.path(), true).unwrap();
-        manager
-            .update()
-            .expect("offline update should succeed as no-op");
+        match manager.update() {
+            Err(Error::OfflineCacheMiss { .. }) => {}
+            other => panic!("expected OfflineCacheMiss, got: {:?}", other),
+        }
     }
 
     /// Default (online) Manager keeps its existing behavior — `offline` is
--- a/crates/checkouts/src/lib.rs
+++ b/crates/checkouts/src/lib.rs
@@ -207,12 +207,28 @@ pub struct Manager {
     base_dir: PathBuf,
     state: ManagerState,
     repos: HashMap<String, Repo>,
+    /// When true, [Manager::checkout_of] / [Manager::update] return
+    /// [Error::OfflineCacheMiss] for any remote that would require a clone or fetch
+    /// rather than performing the network operation. Mirrors the value of
+    /// [`mctx::Config::use_remote_cache`] (i.e. `--no-fetch`).
+    offline: bool,
 }
 
 impl Manager {
     /// Initializes the checkouts manager in the given directory, returning a
     /// thread-safe, cloneable handle to the manager instance.
     pub fn new_in_dir<P: Into<PathBuf>>(base_dir: P) -> Result<ManagerHandle, Error> {
+        Self::new_in_dir_with_offline(base_dir, false)
+    }
+
+    /// Same as [Self::new_in_dir], but with an `offline` flag. When `offline=true`,
+    /// any operation that would require a clone or fetch (new remote, or refresh of
+    /// a known remote) surfaces as [Error::OfflineCacheMiss] rather than performing
+    /// the network operation.
+    pub fn new_in_dir_with_offline<P: Into<PathBuf>>(
+        base_dir: P,
+        offline: bool,
+    ) -> Result<ManagerHandle, Error> {
         let base_dir = base_dir.into();
         let db_path = base_dir.join("git").join("db");
         fs::create_dir_all(&db_path)?;
@@ -237,6 +253,7 @@ impl Manager {
             base_dir,
             state,
             repos,
+            offline,
         };
         Ok(ManagerHandle(Arc::new(Mutex::new(manager))))
     }
@@ -249,7 +266,16 @@ impl Manager {
     }
 
     /// Updates all repos to latest - does nothing for refs which arent symbolic (i.e. commits).
+    /// In offline mode, returns [Error::OfflineCacheMiss] for the first remote we'd
+    /// have to fetch.
     pub fn update(&mut self) -> Result<(), Error> {
+        if self.offline {
+            // In offline mode, `update` is a no-op for known remotes: we have
+            // no way to refresh, but the caller's intent is "use what's there",
+            // so returning Ok is more useful than erroring on every refresh.
+            // (`checkout_of` still errors when an UNKNOWN remote is requested.)
+            return Ok(());
+        }
         let checkouts_dir = self.git_checkouts_dir();
         for (_remote, id) in self.state.git_remotes.iter_mut() {
             let repo = self.repos.get_mut(id).unwrap();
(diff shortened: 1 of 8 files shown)


msample:
This change seem like it's still allowlisting source downloads from our internal GCS?   So it feels like this should be some form of --trusted-source-domains [list,..]  style of option.

The original --no-fetch behaviour was nice for rebuilding the world locally on your laptop in a package repo (whether they came from GCS, github, or some mirror).  It's useful for any pkg repo (well, if/when the bin cache source can be overridden)

twitchyliquid64:
For rebuilds, we now have the `minimal package --rebuild` flag. Need to double check that rebuilds everything when no package is provided, but that should be the goto for the 'rebuild the world' use-case.

Can discuss separate to this PR, but ive been thinking about turning down the --no-cache flag, its too confusing + --rebuild should cover its intended use.

crates/checkouts/src/lib.rs
```diff
@@ -207,12 +207,28 @@ pub struct Manager {
...
+    /// [Error::OfflineCacheMiss] for any remote that would require a clone or fetch
+    /// rather than performing the network operation. Mirrors the value of
+    /// [`mctx::Config::use_remote_cache`] (i.e. `--no-fetch`).
+    offline: bool,
 }
 
 impl Manager {
     /// Initializes the checkouts manager in the given directory, returning a
     /// thread-safe, cloneable handle to the manager instance.
     pub fn new_in_dir<P: Into<PathBuf>>(base_dir: P) -> Result<ManagerHandle, Error> {
```
twitchyliquid64, line 220: Future nit: when this grows more configurations we should create an `Options` struct and take that for a `new` method, ala Builder pattern.

crates/checkouts/src/lib.rs
```diff
@@ -249,7 +266,16 @@ impl Manager {
     }
 
     /// Updates all repos to latest - does nothing for refs which arent symbolic (i.e. commits).
+    /// In offline mode, returns [Error::OfflineCacheMiss] for the first remote we'd
+    /// have to fetch.
     pub fn update(&mut self) -> Result<(), Error> {
+        if self.offline {
```
twitchyliquid64, line 273: This is called from things that really want to be fresh like `minimal update` - so prefer returning an error rather than silently lameducking; noop'ing here could lead to some hard-to-debug bugs.

Fwiw, `update()` shouldnt get called willy-nilly.

msample:
it seems cleaner to have --no-fetch-srcs for this.   The srcs are not diff levels of the binary/outputs cache or a caches at all - misses aren't tolerated.

bryan-minimal:
Thanks both — agreed on the separation. The conflation here was wrong: "rebuild from source locally" and "I'm fully offline" are distinct semantics that shouldn't share a flag, especially with `--rebuild` already covering the laptop-rebuild use case independently.

Refactor plan:
- Keep all the lower-level plumbing (`FileCache::offline`, `Manager::offline`, `OfflineCacheMiss` errors). Existing `--no-fetch` semantics restored exactly as before.
- Introduce a new `--offline` CLI flag with its own `Config::is_offline()` accessor. The trigger for `FileCache`/`Manager`'s offline mode moves from `!use_remote_cache()` to the new `is_offline()`.
- `use_remote_cache()` becomes `!no_fetch && !is_offline()` so `--offline` is a strict superset of `--no-fetch` (offline implies no artifact cache fetch either, since by definition we can't reach it).
- Add tests on the new offline path: `FileCache::download` returns `OfflineCacheMiss` on cache miss when offline; `Manager::checkout_of` returns `OfflineCacheMiss` for an unknown remote and skips `repo.fetch()` for a known one.

@msample re: the GCS-allowlist question — for the offline use case the URL doesn't matter because the cache is sha-keyed: pre-populate `~/.cache/minimal/downloads/<sha>/<filename>` and minimal finds bytes by hash without consulting the URL. So `gs://internal-bucket/...` vs `https://github.com/...` vs your own mirror all behave the same in offline mode. The `--trusted-source-domains` framing makes sense for a separate semi-online use case (allowlisted networks, blocklisted ones), but I'll keep that scope out of this PR.

Will push the updated branch shortly.

crates/checkouts/src/lib.rs
bryan-minimal: Good call — fixed in 048c6b1c. `update()` now returns `Error::OfflineCacheMiss` in offline mode (using the first known remote, or `"<no remotes>"` placeholder if state is empty). Test updated to assert the new behavior.

crates/checkouts/src/lib.rs
bryan-minimal: Acknowledged — agreed, will reach for an `Options` struct + builder when this constructor accumulates more flags. Tracking as future work; left the current shape as-is since it's only one new param.

crates/mctx/src/lib.rs
```diff
@@ -403,9 +406,16 @@ impl Context {
...
+        // `--offline` flips the underlying FileCache into offline mode so any
+        // source-URL cache miss surfaces as FileCacheError::OfflineCacheMiss
+        // rather than a silent network fetch.
+        let rs = common::RemoteStorage::new_with_offline(
+            self.config.downloads_dir(),
+            false,
+            self.config.is_offline(),
+        )
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
bryan-minimal: Fixed in 048c6b1c. Replaced `.unwrap()` with `.map_err(|e| Error::Other(anyhow!("initializing remote storage: {}", e)))?` so init failures propagate via the existing `Error` type instead of panicking the CLI.

bryan-minimal:
Stacked one more commit (9e6668b9): fixes a wording bug in the OfflineCacheMiss error messages. They were still saying `--no-fetch is set` instead of `--offline is set` — caught during a local smoke test where I would have pointed an operator at the wrong flag. Three sites updated (file_cache.rs + checkouts/error.rs + mctx/error.rs).

Also independently smoke-tested the end-to-end offline flow: `minimal --offline package <pkg>` correctly errors on any transitive prebuilt-dep cache miss, walking the full graph (not just the leaf source). Architecturally sound for the hermetic-builder use case — staging just needs to walk the same transitive graph.

bryan-minimal:
Pushed one more commit on top: `deb19a4e mctx: split offline from no_fetch (artifact cache stays usable in offline mode)`.

This refines the design from earlier review: the remote artifact cache is sha-verified on hydrate (`rcache::RemoteCache::materialize`), so cached pulls are hermetically clean even when `--offline` is set. Dropping the offline-implies-no_fetch coupling lets callers compose them independently:

- `--offline` alone: source/VCS fetches gate on cache; artifact cache still readable.
- `--no-fetch` alone: artifact cache off; source/VCS fetches still happen (existing meaning).
- `--offline --no-fetch`: both off, fully air-gapped (was the previous default of `--offline` alone).

**Motivating use case** (concrete): the minimalmertic hermetic-builder-rs sets `with_offline(true)` to surface mirror-source gaps as clear errors, but ALSO wants to read from a signer-attested build cache for expensive deps like llvm (3+ hour rebuild otherwise). Previously these were mutually exclusive; now they compose.

Non-breaking for callers that relied on the old coupling — they should set both flags explicitly. Default behavior preserved for `--offline` alone (still the safe choice).

Diff: `crates/mctx/src/config.rs` only (18 insertions, 5 deletions). No new tests because the existing offline-mode tests already cover the cache layers; this is purely a config-flag-semantics adjustment.

Sibling PR #228 carries the related sandbox2 + op changes that the hermetic-builder-rs use case needs alongside this — kept separate because that's CS-sandbox plumbing rather than offline-mode work.

twitchyliquid64 (approved):

PR #224, opened 2026-05-22
crates/sandbox2: Copy `Local` build_deps to avoid hardlinking errors

twitchyliquid64:
Fixes gominimal/inbox#129

## Summary by CodeRabbit

* **Chores**
  * Local build dependencies are now copied into sandboxes (with permissions) instead of being hardlinked, improving isolation and reliability of sandboxed builds.
* **Bug Fixes**
  * Sandboxing now returns a clear error when mapped files in the root filesystem are unsupported, producing more actionable messages for failing builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/224?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/mctx/src/error.rs
+++ b/crates/mctx/src/error.rs
@@ -249,6 +249,10 @@ impl From<sandbox2::Error> for Error {
                 }
             },
             sandbox2::Error::Output(e) => Self::Other(e.into()),
+            sandbox2::Error::MappedFile(p) => Self::Other(anyhow::anyhow!(
+                "internal: file {} mapped to rootfs",
+                p.display()
+            )),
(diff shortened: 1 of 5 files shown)


0chroma (approved):
Tested the fix by building packages from /var/home/.../pkgs (a btrfs subvolume separate from the cache mount point at /home/.cache/minimal). Successfully rebuilt tar, grep, and make with --rebuild — all completing without cross-device link errors. The FileCopy approach is the right default rather than a workaround since every package uses Local build deps, so there's nothing to warn about.

PR #228, opened 2026-05-25
sandbox2 + op: CS-compatible nested-sandbox plumbing

bryan-minimal:
## Summary

Three independent additions to minimal that unblock running `minimal package` inside Google Cloud Confidential Space as a nested sandbox (a top-level minimal invocation that itself shells out to `minimal package` inside hermetic-builder workloads).

All three are additive — no existing-caller behavior changes.

## Changes

**`sandbox2`: opt-in bind-mount /proc** (52996b5)

The default sandbox masks `/proc` to prevent introspection. Nested sandbox use (one `minimal` invocation spawning another `minimal package` underneath) needs the inner sandbox to read `/proc/self/*` for its own bookkeeping. Adds a `proc_bind_mount: bool` field on `Sandbox`. Default off (preserves existing behavior). When true, sandbox2 bind-mounts the host `/proc` into the sandbox.

**`op` + `orchestrator`: caller-supplied env_vars + rootfs for build sandboxes** (a3d35c1)

The build-spec sandbox config is currently fixed: caller can't add env vars or override the rootfs without recompiling minimal. Adds an `extra_env_vars: Option<...>` + `extra_rootfs: Option<PathBuf>` to `BackendShared`. Plumbed through `LocalBackend::build_in_sandbox` so existing callers see no change; the new fields default `None`.

Motivating use: hermetic-builder-rs runs in CS with a pre-staged rootfs (toolchain + libs) different from the minimal default, and needs to inject `BUILD_PKG`, `OUTPUT_DIR`, `STATE_DIR` env vars based on the queue task it's processing.

**`sandbox2`: capture stdout tail in InvocationFailed** (6c62ce9)

Build failures that print diagnostics to stdout (autotools `configure` for example) currently lose those bytes — `InvocationFailed` carries only stderr. Adds parallel stdout capture (same 4KiB rolling tail as stderr). Operators triaging hermetic build failures gain ~2x diagnostic surface for free.

## Test plan

- [x] `cargo fmt --check` clean (CI; macOS host can't run sandbox2 clippy — Linux-only deps)
- [x] Smoke-tested in production: all three are running in minimalmertic's hermetic-builder-rs (CS workload) and have processed 250+ successful builds since landing on the local branch.

## Relationship to #202

Lifted these out of [#202](https://github.com/gominimal/minimal/pull/202)'s branch where they were stacked. PR #202 keeps the `--offline` flag theme; this PR is purely about CS-compatibility sandbox plumbing. Reviewed separately on purpose — the changes are orthogonal.

## Summary by CodeRabbit

* **New Features**
  * Execution error messages now capture and display sandboxed process output (stdout) alongside existing error diagnostics, providing enhanced visibility into failed task execution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/228?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/sandbox2/src/lib.rs
+++ b/crates/sandbox2/src/lib.rs
@@ -400,6 +400,68 @@ impl<C: Channel> Sandbox<C> {
         Self::bind_mount(&self.state_dir, "/state", rec, &mut container)?;
         Self::bind_mount(&self.base_dir.join("run"), "/run", rec, &mut container)?;
 
+        // MINIMAL_INTERNAL_CS_BUILD: undocumented bundle of behaviors
+        // needed for running minimal as a library inside a GCP
+        // Confidential Space workload. Not for general callers; we
+        // expose it as a private "I am the trusted CS builder" flag so
+        // it can't accidentally enable non-hermetic behavior in normal
+        // `minimal package <pkg>` invocations.
+        //
+        // What the bundle does:
+        //
+        // 1. Bind-mount the outer /proc into the sandbox + enable
+        //    Runctl::MountFallback. CS workload containers get OCI
+        //    MaskedPaths (containerd's default — tmpfs over
+        //    /proc/{kcore,scsi,keys,...}). The Linux kernel's
+        //    anti-unmask guard then refuses any nested procfsmount
+        //    over a masked parent, so hakoniwa's implicit
+        //    `Container::new()::procfsmount("/proc")` returns EPERM.
+        //    Workaround: bind the outer (already-masked) /proc instead
+        //    of trying to mount a fresh procfs. The MountFallback
+        //    runctl is required for the same reason — hakoniwa emits
+        //    a mandatory MS_REMOUNT after every bind, which also
+        //    fails on the masked /proc until we let it retry with the
+        //    source mount's existing flags.
+        //
+        //    Caveat (noted by @twitchyliquid64 on the PR): MountFallback
+        //    is a container-global runctl, not per-mount. It applies
+        //    to every bind in the sandbox — a sandbox that wanted to
+        //    assert e.g. NOEXEC on a target may silently end up with
+        //    the source's existing flags instead. Acceptable inside
+        //    the CS attested boundary where outer isolation handles
+        //    the actual security property; the inner hakoniwa is for
+        //    build-script reproducibility, not isolation. Filed
+        //    upstream to see if hakoniwa would accept a per-mount
+        //    runctl that would let us scope this to /proc only.
+        //
+        // 2. Symlink the hermetic-builder's ecosystem cache paths.
+        //    hermetic-builder-rs stages caches at
+        //    <state>/cs-mirror/{cargo-vendor,npm-cache,...}/. The
+        //    state dir is bind-mounted at /state, so the caches are
+        //    already accessible there. Pkg build.shs in the
+        //    minimalmertic pkg set reference /cargo-vendor,
+        //    /npm-cache, etc. as hardcoded paths (per the
+        //    if-d-cargo-vendor offline-cache idiom); these symlinks
+        //    bridge the two without requiring every build.sh to know
+        //    about /state/cs-mirror/. Dangling symlinks (when a given
+        //    cache wasn't staged for the current pkg) are safe —
+        //    `[ -d /<name> ]` returns false through a dangling
+        //    symlink, so the build.sh's online-fallback branch is
+        //    taken correctly.
+        //
+        // Inert when the env var is unset; default `minimal package`
+        // invocations see no behavior change.
+        if std::env::var("MINIMAL_INTERNAL_CS_BUILD").as_deref() == Ok("1") {
+            container.bindmount_rw("/proc", "/proc");
+            container.runctl(hakoniwa::Runctl::MountFallback);
+            container.symlink("/state/cs-mirror/cargo-vendor", "/cargo-vendor");
+            container.symlink("/state/cs-mirror/npm-cache", "/npm-cache");
+            container.symlink("/state/cs-mirror/pnpm-store", "/pnpm-store");
+            container.symlink("/state/cs-mirror/bun-cache", "/bun-cache");
+            container.symlink("/state/cs-mirror/pip-wheels", "/pip-wheels");
+            container.symlink("/state/cs-mirror/rust-stage0", "/rust-stage0");
+        }
+
         if self.needs_bin_symlink()? {
             container.symlink("/usr/bin", "/bin");
         }
--- a/crates/sandbox2/src/lib.rs
+++ b/crates/sandbox2/src/lib.rs
@@ -621,35 +621,50 @@ impl<C: Channel> Sandbox<C> {
             let child_stdout = child.stdout.take();
             let child_stderr = child.stderr.take();
 
-            // Stdout thread
+            // Stdout thread — like stderr, capture a rolling tail of the
+            // last ~4 KiB so the caller can include it in InvocationFailed
+            // when a build script swallows its stderr (mesa's pip install
+            // 2>/dev/null pattern) but emits the real diagnostic to stdout.
             let stdout_file = self.stdout.take();
             let (stdout_tx, mut stdout_rx) = tokio::sync::mpsc::channel::<Vec<u8>>(32);
-            let stdout_thread = std::thread::spawn(move || -> Result<Option<fs::File>, Error> {
-                let mut file = stdout_file;
-                if let Some(mut pipe) = child_stdout {
-                    let mut buf = [0u8; 8192];
-                    loop {
-                        let n = pipe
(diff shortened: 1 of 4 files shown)


bryan-minimal:
Thanks for the careful review! Addressed all 4 comments — force-pushed `bryan/cs-sandbox-extensions`:

**1. Display impl rendering** (`crates/mctx/src/error.rs`) — Fixed. `Error::Execution`'s `Display` now prints the exit summary, then appends `stderr:` and `stdout:` blocks when non-empty. Folded into the sandbox2 stdout-tail commit so the field-add and the user-facing render land together.

**2. `drain(..)` consuming `extra_rootfs`** (`crates/op/src/specs.rs`) — Fixed. Changed to `for extra in &self.extra_rootfs { rootfs.insert(SandboxMapped::Dir(extra.clone())); }` so subsequent `run()` calls still see the caller-supplied paths. Added a comment explaining why.

**3. `MINIMAL_SANDBOX_BIND_PROC` truthy-check** (`crates/sandbox2/src/lib.rs`) — Fixed. Strict equality check now: `std::env::var("MINIMAL_SANDBOX_BIND_PROC").as_deref() == Ok("1")`. `=0` (and any other value) no longer enables the bind-mount — explicit `=1` only.

**4. 11-positional-arg constructor / builder pattern** (`crates/orchestrator/src/local_backend.rs`) — Documented the deliberate suppression but did NOT do the builder refactor in this PR. Reasoning: a typed `OrchestratorConfig` builder is the right shape, but pushing 11+ fields into a new struct is structurally larger than the env_vars+rootfs feature this PR adds. Tracked separately for a focused refactor PR. Comment added so future readers understand the lint allow.

Mirrored these same fixes on the local `bryan/fetcher-extra-env-vars` branch we vendor from in the downstream hermetic-builder-rs — keeps the downstream test-runs consistent with what's under review here.

twitchyliquid64:
Overall, if we need to make changes to support confidential computing, I would prefer to scope them to exactly just our builder (i.e. no new configurables, just an undocumented flag `MINIMAL_INTERNAL_CS_BUILD` or something), or use existing mechanisms (can we just set the environment variables in `build.sh` ?)

Or like if we need a new primitive (i.e. for loading go sources or whatever) we should figure out a clean way to plumb/scaffold it that isnt just env vars, so its less likely to accidentally be the source of non-hermetic behavior in the future (given our users wont be building in this mode).

> a top-level minimal invocation that itself shells out to minimal package inside hermetic-builder workloads

I'm confused about this, minimal-in-minimal? Shouldnt that not work at all because of NO_NEW_PRIVS?

> Motivating use: hermetic-builder-rs runs in CS with a pre-staged rootfs (toolchain + libs) different from the minimal default, and needs to inject BUILD_PKG, OUTPUT_DIR, STATE_DIR env vars based on the queue task it's processing.

Why is `OUTPUT_DIR` getting overridden? Staging output files in `OUTPUT_DIR` is such a core part of the build system this seems like a large structural change and not a small tweak.

> capture stdout tail in InvocationFailed

SGTM.

crates/sandbox2/src/lib.rs
```diff
@@ -400,6 +400,52 @@ impl<C: Channel> Sandbox<C> {
         Self::bind_mount(&self.state_dir, "/state", rec, &mut container)?;
         Self::bind_mount(&self.base_dir.join("run"), "/run", rec, &mut container)?;
 
+        // GCP Confidential Space (and other containerd-default runtimes)
+        // mask sub-paths of /proc (/proc/kcore, /proc/scsi, ...) via
+        // tmpfs overlays. The kernel then refuses any procfsmount in a
+        // nested namespace to prevent unmasking attacks, EPERM. The
```
twitchyliquid64: Is there no way to turn this off?

crates/sandbox2/src/lib.rs
```diff
@@ -400,6 +400,52 @@ impl<C: Channel> Sandbox<C> {
...
+        //
+        // Inert when /proc points at the outer container's own /proc,
+        // which the workload already controls. Acceptable inside a CS
+        // attested boundary where the outer scope is already isolated.
+        // Strict "1" check rather than is_some(): explicit opt-in.
+        // Avoids the footgun where MINIMAL_SANDBOX_BIND_PROC=0 (or
+        // any other value) would still enable the bind-mount.
+        if std::env::var("MINIMAL_SANDBOX_BIND_PROC").as_deref() == Ok("1") {
+            container.bindmount_rw("/proc", "/proc");
+            container.runctl(hakoniwa::Runctl::MountFallback);
```
twitchyliquid64, line 456: Also worth noting this applies to all mountpoints in the sandbox, not just `/proc`. Might not be avoidable if we cant turn off the weird behavior of the CS kernel.

bryan-minimal:
> Overall, if we need to make changes to support confidential computing, I would prefer to scope them to exactly just our builder (i.e. no new configurables, just an undocumented flag `MINIMAL_INTERNAL_CS_BUILD` or something), or use existing mechanisms (can we just set the environment variables in `build.sh` ?)
> 
> Or like if we need a new primitive (i.e. for loading go sources or whatever) we should figure out a clean way to plumb/scaffold it that isnt just env vars, so its less likely to accidentally be the source of non-hermetic behavior in the future (given our users wont be building in this mode).
> 
Thanks for the thorough review, putting behind a flag and reducing the changes sounds like a good plan. This was a bit of a mess to get the builder working, I'm going to redesign and put behind flag. I'll drop the extra_env_vars and extra_rootfs API additions

> > a top-level minimal invocation that itself shells out to minimal package inside hermetic-builder workloads
> 
> I'm confused about this, minimal-in-minimal? Shouldnt that not work at all because of NO_NEW_PRIVS?
> 
Yeah, this isn't accurate, the builder is using minimal as a library inside the Confidential Spaces setup, not shelling from minimal. This is a CS issue, not a minimal-in-minimal.

> > Motivating use: hermetic-builder-rs runs in CS with a pre-staged rootfs (toolchain + libs) different from the minimal default, and needs to inject BUILD_PKG, OUTPUT_DIR, STATE_DIR env vars based on the queue task it's processing.
> 
> Why is `OUTPUT_DIR` getting overridden? Staging output files in `OUTPUT_DIR` is such a core part of the build system this seems like a large structural change and not a small tweak.
> 
Yeah, this isn't accurate either, this was the original plan 1 I was using in hermetic builder, but my final design is just adding to the build except for some of the original Go work, but that came move into the pkgs changes. I was using some extra env vars to make the offline stuff work, but they can all be staged in /state instead of the top level extra bindings I was using when I was testing. I should be able to get Claude to redesign the hermetic builder to better work with what we have. 

> > capture stdout tail in InvocationFailed
> 
> SGTM.

I'll submit an updated patch soon, the /proc is a limitation of Confidential Spaces and hakoniwa doesn't support a per-mount runctl, so it is all or nothing. I'm going to move that part behind the the flag as suggested. If we want to lock it down more, I can see if there is a way to patch hakoniwa to be more flexible in that config since they have been pretty respsonsive.

crates/sandbox2/src/lib.rs
bryan-minimal: No, unfortunately part of Confidential Spaces, here is Claude's explanation, 

  Root cause: CS workload containers get OCI MaskedPaths (containerd's
  default — tmpfs masks /proc/{kcore,scsi,keys,...}). The Linux kernel's
  anti-unmask guard then refuses any nested procfsmount over a masked
  parent, so hakoniwa's implicit Container::new()::procfsmount returns
  EPERM. Not configurable from a CS workload — both the MaskedPaths
  layer and the kernel guard are sealed.

crates/sandbox2/src/lib.rs
bryan-minimal: Yeah, this is a hakoniwa limitation, all or nothing on the runctl mount options. I'm renmaing this behind the MINIMAL_INTERNAL_CS_BUILD flag.

twitchyliquid64:
Gotcha, so it sounds like the two broad strokes we need are:

1. The runctl for /proc
2. Some mechanism for plumbing the goproxy and stuff?

bryan-minimal:
Force-pushed v2 implementing the redesign:

- Dropped `extra_env_vars` + `extra_rootfs` + `new_orchestrator_with_env` public surface entirely.
- All CS-specific behavior (/proc bind-mount + MountFallback + ecosystem cache symlinks) now behind a single undocumented `MINIMAL_INTERNAL_CS_BUILD=1` env var.
- GOPROXY/GOSUMDB moved to each Go pkg's `build.sh` (minimal stays ecosystem-agnostic).
- Added a comment explicitly calling out your line-446 caveat about MountFallback's sandbox-wide scope.

PR collapses from 3 commits to 2; new public API surface is zero.

twitchyliquid64 (approved):

PR #230, opened 2026-05-25
crates/sessions: new crate for session primitives + lifecycle hooks primitive

evanspearman:
Adds a `sessions` library crate with two types:

  - `LifecycleHook` — three optional script slots (`on_activate`, `on_destroy`, `on_failure`), at least one required. Invariant enforced in the builder and reused on the serde path via `try_from`/`into`.
  - `HookScript` — adjacently tagged: `{ type = "inline" | "external", value = "..." }`.
  - `Loadout` — stub for now; will grow as `minimal2`/`minimald` need it.

Pedantic clippy on; tests + doctests cover the invariant and serde round-trip.

## Release Notes

* **New Features**
  * Added sessions crate to workspace.
  * Introduced lifecycle hooks for sessions with activation, destruction, and failure event handlers.
  * Added support for inline and external scripts in lifecycle hooks.
  * Introduced session loadouts with environment variable management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/sessions/src/lifecyclehook.rs
+++ b/crates/sessions/src/lifecyclehook.rs
@@ -1,0 +1,315 @@
+//! Lifecycle hooks: scripts run on activation, destruction, or failure of a host resource.
+//!
+//! A [`LifecycleHook`] groups up to three scripts (one per transition point) and
+//! guarantees that at least one is present — an empty hook is a no-op and is
+//! rejected at construction time. Build hooks programmatically through
+//! [`LifecycleHook::builder`] or deserialize them from configuration; both paths
+//! enforce the same invariant.
+
+use core::fmt;
+use std::path::PathBuf;
+
+/// Errors produced when constructing a [`LifecycleHook`].
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum Error {
+    /// No hook script was provided.
+    ///
+    /// Returned by [`LifecycleHookBuilder::build`] (and therefore by
+    /// deserialization) when none of `on_activate`, `on_destroy`, or
+    /// `on_failure` is set.
+    EmptyLifecycleHook,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::EmptyLifecycleHook => write!(
+                f,
+                "At least one of on_activate, on_destroy, or on_failure must be specified"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+/// A script executed by a lifecycle hook.
+///
+/// Either the script body provided directly ([`Inline`](Self::Inline)) or a
+/// path to a script file on disk ([`External`](Self::External)).
+///
+/// Serialized with adjacent tagging — the wire form is a table with `type` and
+/// `value` keys:
+///
+/// ```toml
+/// on_activate = { type = "inline",   value = "echo hi"   }
+/// on_failure  = { type = "external", value = "./run.sh"  }
+/// ```
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type", content = "value", rename_all = "lowercase")]
+pub enum HookScript {
+    /// Script body stored inline.
+    Inline(String),
+    /// Path to a script file on disk.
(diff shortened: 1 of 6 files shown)


crates/sessions/src/lifecyclehook.rs
```diff
@@ -0,0 +1,315 @@
...
+/// on_activate = { type = "inline",   value = "echo hi"   }
+/// on_failure  = { type = "external", value = "./run.sh"  }
+/// ```
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type", content = "value", rename_all = "lowercase")]
+pub enum HookScript {
+    /// Script body stored inline.
+    Inline(String),
+    /// Path to a script file on disk.
+    External(PathBuf),
```
twitchyliquid64, line 55: Over the time of building minimal Ive started to dislike `PathBuf`, because it leads to confusion/invariants/bugs based on:

1. Is this a path relative to the user system, the sandbox fs, or the minimald (in the future) ?
2. Is this meant to be interpreted relative to the cwd, if so which of the above contexts?
3. More of a grumble, but if you join an absolute path with cwd, it overwrites with the absolute path. I can feel a bug coming from this one day (i.e. if a user specifies an absolute path in the config).

Not sure what the fix is here, but we may want some newtype around paths. Food for thought.

crates/sessions/src/lifecyclehook.rs
```diff
@@ -0,0 +1,315 @@
...
+                f,
+                "At least one of on_activate, on_destroy, or on_failure must be specified"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+/// A script executed by a lifecycle hook.
```
twitchyliquid64, line 37: Whats the constraints aroud what a valid script is? Do we want to just say 'bash v4 or greater compatible?' (I would say POSIX shell except everyone ignores when thats the constraint and uses modern features lol)

crates/sessions/src/lifecyclehook.rs
evanspearman: Yeah, newtypes sound like the right way to go here. I'm using them for paths in the patches stuff I'm working on, where it becomes even more important to keep track of this stuff.

crates/sessions/src/lifecyclehook.rs
evanspearman: I think it depends on exactly when these run. If it's something like:

```
Install packages -> Activate Hook -> ... -> Destroy Hook -> Packages are gone
                                     !!! -> Failure Hook -> Destroy Hook -> Packages are Gone
```

Then we don't really need to constrain it. It could be fish, python, ruby, whatever you want as long as you have the package to run it. Just put in the right shebang.

If they're run when you don't have your packages available then I think some specific version of bash (as long as we can grantee that's available) makes sense.

crates/sessions/src/lifecyclehook.rs
evanspearman: Actually, they're probably not strictly newtypes as they'll probably need to contain a bit more context in some cases.

crates/sessions/src/lifecyclehook.rs
twitchyliquid64: Hmmm, at this stage i realize im not fully sure what the context is for lifecycle hooks. Are these basically just activation scripts for a session?

If theres one activation script then it makes sense to have a configurable shell like bash/fish etc. But if we are composing multiple scripts we need the env vars that get exported in one + the background tasks/jobs that get run to stay alive as all the activation scripts run (i.e. Karl's use case of launching postgres, we dont want postgres to die when the activation script finishes), in which case what shell is chosen for interpreting all those scripts?

evanspearman: Ah, right. In that case I think we just do bash unless we have a good reason not to. I don't think that needs to be encoded here though.

twitchyliquid64 (approved):

evanspearman:
Going to merge this now. I've been playing around with more contextful path types a bit, but I don't see much harm in just switching this to them later.

PR #235, opened 2026-05-27
crates/minimald: create scaffolding for building up SSH channels

twitchyliquid64:
Basically, the connection handler type wraps the connection handle, and very quickly reaches into it via a lock to tweak state when managing the ssh connection. As requests for pty, env vars etc come in, we mutate this pending state (owned by the Connection) to accumlate the configurables, sorta like the builder pattern.

When its time for the channel to actually be wired to something heavy, we take the `RuChannel<Msg>` out of the connection state, and it lives with the async context which actually does the work. The handler method returns quickly, freeing up the thread which services the connection once again. 

## Summary by CodeRabbit

* **New Features**
  * Per-channel SSH session lifecycle and per-channel configuration (env/PTY) handling
  * SSH one-shot RPC subsystem with a GetVersion RPC
  * Per-instance sockets via a CLI instance option for isolated daemon instances
  * Optional protocol tracing controlled via environment

* **Chores**
  * Separate state vs cache dirs and per-instance client directories
  * Persistent on-disk host key generation/usage and memoized server state

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/235?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/minimald/src/connection.rs
+++ b/crates/minimald/src/connection.rs
@@ -1,6 +1,29 @@
-use russh::server::{Config as RuConfig, RunningSession};
-pub use std::sync::{Arc, Mutex};
-use tokio::net::UnixStream;
+use russh::{
+    Channel as RuChannel, ChannelId,
+    server::{Config as RuConfig, Msg, RunningSession, Session},
+};
+use std::{
+    collections::BTreeMap,
+    sync::{Arc, LazyLock},
+};
+use tokio::{net::UnixStream, spawn, sync::Mutex};
+
+use crate::{
+    ChannelConfig, RequestedPty,
+    rpc::{self, OneshotSshRpc},
+    server::ServerStateHandle,
+};
+
+static PROTOCOL_TRACE_ENABLED: LazyLock<bool> =
+    LazyLock::new(|| std::env::var("PROTOCOL_TRACE").as_deref() == Ok("1"));
+
+macro_rules! protocol_trace {
+    ($($arg:tt)*) => {
+        if *PROTOCOL_TRACE_ENABLED {
+            tracing::info!($($arg)*);
+        }
+    };
+}
 
 /// The auth state of the SSH connection.
 #[derive(Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


twitchyliquid64:
Its not much, but this simple GetVersion RPC works:

```shell
xxx@lptp:~$ echo "null" | \
  ssh -sT -o ProxyCommand='socat - UNIX-CONNECT:/home/xxx/.local/state/minimal/providers/local-0/ssh.sock' \
  -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' local \
  minimald-v1-GetVersion
Warning: Permanently added 'local' (ED25519) to the list of known hosts.

{"version":"0.0.1","long_version":"0.0.1 (dirty b20ffb6a)","stdlib_version":"0.0.15"}
```

crates/minimald/src/connection.rs
```diff
@@ -1,6 +1,28 @@
...
+use russh::{
+    Channel as RuChannel, ChannelId,
+    server::{Config as RuConfig, Msg, RunningSession, Session},
+};
+use std::{
+    collections::BTreeMap,
+    sync::{Arc, LazyLock, Mutex},
+};
+use tokio::{net::UnixStream, spawn};
+
```
evanspearman: Since we're using these in a Tokio context we need to use the Tokio versions. Tokio is smart about scheduling around the tokio sync primitives so it can do things like run something else while waiting on a lock.

crates/minimald/src/connection.rs
```diff
@@ -10,22 +32,109 @@ pub enum Auth {
     Local,
 }
 
+/// Represents the parameters of a requested PTY.
+#[derive(Debug, Clone)]
+#[allow(dead_code)]
```
evanspearman: Instead of this, I'd recommend adding a lib.rs to the crate as well and putting the modules in there.

crates/minimald/src/connection.rs
```diff
@@ -10,22 +32,109 @@ pub enum Auth {
...
+pub struct RequestedPty {
+    char_sizes: (u32, u32),
+    pixel_sizes: (u32, u32),
```
evanspearman: Might be worth just defining a Vec2U32 or Size2U32 struct instead of bare tuples.

crates/minimald/src/connection.rs
```diff
@@ -10,22 +32,109 @@ pub enum Auth {
...
     pub auth: Auth,
-
     /// The username given via SSH. Not populated till the client
     /// authenticates the connection.
     pub ssh_username: Option<String>,
+    /// State specific to an SSH channel.
+    pub channels: BTreeMap<ChannelId, Channel>,
+
+    serv: ServerStateHandle,
 }
```
evanspearman, line 116: Are these all completely independent? If not we should make them private so we can enforce invariants.

crates/minimald/src/connection.rs
```diff
@@ -78,4 +240,140 @@ impl russh::server::Handler for ConnectionHandler {
             Ok(russh::server::Auth::reject())
         }
     }
+
+    async fn channel_open_session(
+        &mut self,
+        c: RuChannel<Msg>,
+        _: &mut Session,
+    ) -> Result<bool, Self::Error> {
+        let mut s = self.0.0.lock().unwrap();
```
evanspearman: I'd recommend using `expect` instead to document the invariant. Something seems not quite right about this though. I'm a bit confused as to why we have multiple layers of wrapper around `Arc<Mutex<Connection>>`.

Did some looking into this and Err happens on `lock` when:

> If another user of this mutex panicked while holding the mutex, then this call will return an error once the mutex is acquired. The acquired mutex guard will be contained in the returned error.

I'm not sure panicking ourselves when that happens is the correct behaviour here. Actually, `tokio::Mutext::lock` is infallible, so the unwrap just goes away when you switch to that.

crates/minimald/src/connection.rs
```diff
@@ -78,4 +240,140 @@ impl russh::server::Handler for ConnectionHandler {
...
+        if s.auth != Auth::Local {
+            return Ok(false); // indicate failure
```
evanspearman, line 236: Failure should probably just be another variant of the module's error type.

crates/minimald/src/rpc.rs
```diff
@@ -0,0 +1,61 @@
...
+        let request: Self::Request<'_> = serde_json::from_slice(&buf)?;
+        let response = handler(request)?;
+        let response_bytes = serde_json::to_vec(&response)?;
+
+        stream
+            .write_all(&response_bytes)
+            .await
+            .map_err(russh::Error::from)?;
+        //stream.flush().await.map_err(russh::Error::from)?;
+        //stream.shutdown().await.map_err(russh::Error::from)?;
```
evanspearman: Just double checking, were these intended to be left in?

crates/minimald/src/server.rs
```diff
@@ -2,70 +2,136 @@ use russh::keys::key::safe_rng;
 use russh::keys::{PrivateKey, ssh_key::Error as KeyError};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
evanspearman: Also should use the tokio version

evanspearman (changes requested):

crates/minimald/src/connection.rs
twitchyliquid64: My usual threshold for dedupe is when we have three or more instances, i think the pty config is going to be the only place we have width+height pairs.

crates/minimald/src/connection.rs
twitchyliquid64: The channel stuff ill defs make private, but for the others, im imagining a future where handlers may want to ask about connection state via `ConnectionHandle`.

crates/minimald/src/connection.rs
twitchyliquid64: Done!

crates/minimald/src/connection.rs
twitchyliquid64: Unfortunately this is the trait api that russh gives us: return false to decline the request for a channel, true to approve.

crates/minimald/src/rpc.rs
twitchyliquid64: I commented them out when debugging something. Ive put them back.

crates/minimald/src/server.rs
twitchyliquid64: Done!

crates/minimald/src/connection.rs
evanspearman: Oops, missed that this is a trait impl.

evanspearman (approved):

PR #237, opened 2026-05-27
feat(minvmd): scaffold crate with libkrun FFI wrappers and smoke test

norrietaylor:
## Summary

Lands the macOS-only `minvmd` crate as a scaffold + FFI surface + gated smoke test. Subsequent PRs add VM boot, the UDS↔vsock bridge, and the lifecycle daemon. 

- New `crates/minvmd` workspace member; `krun` module is `#[cfg(target_os = "macos")]`-gated. Linux ships only a runtime-bailing stub so existing Linux-only CI stays green.
- `build.rs` emits link search + rpath on macOS, defaulting to `/opt/homebrew/lib` with a `LIBKRUN_PREFIX` env override.
- libkrun FFI in `src/krun/raw.rs` covers the smoke-test surface (`create_ctx`, `set_vm_config`, `set_exec`, `start_enter`, `free_ctx`) plus the forward boot surface (`set_root`, `set_kernel`, `add_vsock_port`, `set_console_output`). One block-level `// SAFETY:` enumerates the pointer / NUL-termination / ownership-transfer invariants the `extern \"C\"` block inherits from the C ABI; per-function doc comments describe each function's contract.
- Safe wrappers in `src/krun/ctx.rs` expose an RAII `Context` (non-Clone / non-Copy, `Drop` calls `krun_free_ctx`) that validates inputs in safe Rust (NUL-termination via `CString`). Every unsafe call site carries its own `// SAFETY:` comment naming the invariants the wrapper enforces. `start_enter` consumes via `mem::forget` because libkrun documents the configuration as consumed unconditionally.
- Typed `VmError` matches the hand-rolled style in `sandbox2`: `Backend { op, code }` preserves errno magnitude; `NulInPath` / `NulInString` cover boundary validation; `StartEnterReturnedUnexpectedly { ret }` surfaces the libkrun-docs-violating path explicitly instead of synthesising a misleading \"errno 0\".
- `minvmd.entitlements` grants only `com.apple.security.hypervisor`. `justfile` `codesign-minvmd` recipe builds release and ad-hoc-signs.
- `tests/krun_smoke.rs` is `#[ignore]` and self-skips unless `MINVMD_E2E=1`. When opted in it spawns the auto-discovered `src/bin/krun_smoke_child` helper (separate process because `krun_start_enter` `exit()`s on success and never returns) which drives the safe wrappers end-to-end.

## Out of scope (intentional)

- VM boot with kernel + Alpine rootfs (follow-up).
- UDS↔vsock bridge to `minimald` (follow-up).
- Lifecycle daemon (run/status/stop, auto-spawn from `minimal2`) (follow-up).
- Networking (gvproxy / TSI).
- All forward FFI surface (`set_kernel`, `set_root`, `add_vsock_port`, `set_console_output`) is declared and wrapped but not yet exercised.

## Test plan

- [x] `cargo build -p minvmd` — green on aarch64-apple-darwin
- [x] `cargo clippy -p minvmd --all-targets -- -D warnings` — no issues
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p minvmd` — 9 passed, 1 ignored (the gated smoke test)
- [x] `MINVMD_E2E=1 cargo test -p minvmd --test krun_smoke -- --include-ignored` — 1 passed against real libkrun v1.18.1 on macOS
- [ ] Linux CI gate: the `krun` module is gated out, no libkrun linkage attempted
- [ ] `just codesign-minvmd` on a Mac — produces a release binary signed with the hypervisor entitlement

## Summary by CodeRabbit

* **New Features**
  * Added a macOS-only minvmd CLI with shell completion and macOS hypervisor entitlement.

* **Libraries / Runtime**
  * Introduced a macOS-only FFI-backed VM API with refined error handling.

* **Tests**
  * Added macOS-only, opt-in smoke tests and a helper to validate VM bring-up sequences.

* **Chores**
  * Registered minvmd in the workspace and added a release build + ad-hoc codesign recipe.

### DIFF
--- a/crates/minvmd/src/bin/krun_smoke_child.rs
+++ b/crates/minvmd/src/bin/krun_smoke_child.rs
@@ -36,8 +36,7 @@ fn main() {
 
 #[cfg(target_os = "macos")]
 fn run_macos() {
-    use minvmd::krun::Context;
-    use minvmd::krun::raw::{KRUN_KERNEL_FORMAT_IMAGE_BZ2, KRUN_KERNEL_FORMAT_IMAGE_GZ};
+    use minvmd::krun::{Context, KRUN_KERNEL_FORMAT_IMAGE_BZ2, KRUN_KERNEL_FORMAT_IMAGE_GZ};
 
     eprintln!("STAGE: create_ctx");
     let mut ctx = Context::create().expect("krun_create_ctx must succeed");
--- a/crates/minvmd/src/bin/krun_smoke_child.rs
+++ b/crates/minvmd/src/bin/krun_smoke_child.rs
@@ -21,35 +21,31 @@
 //!
 //! Linux is a no-op stub (exit 0); the integration test is `target_os = macos`.
 
-fn main() {
+fn main() -> Result<(), minvmd::VmError> {
+    #[cfg(target_os = "macos")]
+    return run_macos();
+
     #[cfg(not(target_os = "macos"))]
     {
         eprintln!("krun_smoke_child: macOS-only; nothing to do on this platform");
-        std::process::exit(0);
-    }
-
-    #[cfg(target_os = "macos")]
-    {
-        run_macos();
+        Ok(())
     }
 }
 
 #[cfg(target_os = "macos")]
-fn run_macos() {
+fn run_macos() -> Result<(), minvmd::VmError> {
     use minvmd::krun::{Context, KRUN_KERNEL_FORMAT_IMAGE_BZ2, KRUN_KERNEL_FORMAT_IMAGE_GZ};
 
     eprintln!("STAGE: create_ctx");
-    let mut ctx = Context::create().expect("krun_create_ctx must succeed");
+    let mut ctx = Context::create()?;
     eprintln!("STAGE: create_ctx ok ctx_id={}", ctx.id());
 
     eprintln!("STAGE: set_vm_config");
-    ctx.set_vm_config(1, 512)
-        .expect("krun_set_vm_config 1 vcpu / 512 MiB must succeed");
+    ctx.set_vm_config(1, 512)?;
     eprintln!("STAGE: set_vm_config ok");
 
     eprintln!("STAGE: set_exec");
-    ctx.set_exec("/bin/true", &[] as &[&str], None::<&[&str]>)
-        .expect("krun_set_exec /bin/true must succeed");
+    ctx.set_exec("/bin/true", &[] as &[&str], None::<&[&str]>)?;
     eprintln!("STAGE: set_exec ok");
 
     let kernel = std::env::var("MINVMD_KERNEL_PATH").ok();
@@ -61,11 +57,11 @@ fn run_macos() {
              start_enter requires a configured kernel + rootfs)"
         );
         drop(ctx); // RAII free_ctx
-        std::process::exit(0);
+        return Ok(());
     };
 
     eprintln!("STAGE: set_root path={rootfs}");
-    ctx.set_root(&rootfs).expect("krun_set_root must succeed");
+    ctx.set_root(&rootfs)?;
     eprintln!("STAGE: set_root ok");
 
     // Per-arch kernel format: the virtio-linux package ships `Image.gz` on
@@ -76,13 +72,7 @@ fn run_macos() {
         KRUN_KERNEL_FORMAT_IMAGE_BZ2
     };
     eprintln!("STAGE: set_kernel path={kernel} format={format}");
-    ctx.set_kernel(
-        &kernel,
-        format,
-        None::<&std::path::Path>,
-        None, // cmdline
-    )
-    .expect("krun_set_kernel must succeed");
+    ctx.set_kernel(&kernel, format, None::<&std::path::Path>, None)?;
     eprintln!("STAGE: set_kernel ok");
 
     eprintln!("STAGE: start_enter");
--- a/crates/minvmd/src/bin/krun_smoke_child.rs
+++ b/crates/minvmd/src/bin/krun_smoke_child.rs
@@ -1,0 +1,97 @@
+//! Test helper binary for `tests/krun_smoke.rs`.
+//!
+//! Exists as a separate process because `krun_start_enter`, on success, calls
+//! `exit()` with the guest workload's exit code and never returns. Driving it
+//! from inside the test process would tear down the test harness; running it
+//! in a child lets the parent test observe the outcome via the child's exit
+//! status.
+//!
+//! Behaviour:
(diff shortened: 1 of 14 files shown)


twitchyliquid64 (commented):
Did we try using the `libkrun` crate? would prefer that rather than rolling our own FFI if that works

https://crates.io/crates/libkrun corresponding to https://github.com/containers/libkrun/blob/main/src/libkrun/src/lib.rs

crates/minvmd/src/bin/krun_smoke_child.rs
```diff
@@ -0,0 +1,96 @@
...
+//!   2 — start_enter returned an error (libkrun-side failure before boot).
+//!   125/126/127 — libkrun init-time error codes per `krun_start_enter` docs.
+//!
+//! Linux is a no-op stub (exit 0); the integration test is `target_os = macos`.
+
+fn main() {
+    #[cfg(not(target_os = "macos"))]
+    {
+        eprintln!("krun_smoke_child: macOS-only; nothing to do on this platform");
+        std::process::exit(0);
```
evanspearman: At the moment, I don't think this call is necessary, but reasonable to keep in to be explicit.

norrietaylor:
@coderabbitai review

crates/minvmd/src/bin/krun_smoke_child.rs
```diff
@@ -0,0 +1,96 @@
...
+fn run_macos() {
+    use minvmd::krun::{Context, KRUN_KERNEL_FORMAT_IMAGE_BZ2, KRUN_KERNEL_FORMAT_IMAGE_GZ};
+
+    eprintln!("STAGE: create_ctx");
+    let mut ctx = Context::create().expect("krun_create_ctx must succeed");
+    eprintln!("STAGE: create_ctx ok ctx_id={}", ctx.id());
+
+    eprintln!("STAGE: set_vm_config");
+    ctx.set_vm_config(1, 512)
+        .expect("krun_set_vm_config 1 vcpu / 512 MiB must succeed");
```
evanspearman: Generally we don't want to use panics in place of normal error handling. I'd recommend only using expect when the invariant being broken is a bug rather than something that can be affected and runtime. For instance, suppose you have a struct that has a storage data structure and then one or more index data structure. If the invariant of a value being in one structure but not the other was broken, expect would be appropriate because it's caused by a bug in the code for keeping them in sync and not something the user is responsible for.

evanspearman: A better way of doing this would be to have an entry point like:

```rs
fn main() -> Result<(), WhateverErrorType> {
    // ...
    run_macos()?;
    // ...
}
```

And then:

```rs
fn run_macos() -> Result<(), WhateverErrorType> {
    // ...
    ctx.set_vm_config(1, 512)?;
    // ...
}
```

...And this is the point I realized this file is just for testing and not production code so probably not a big deal.

crates/minvmd/src/error.rs
```diff
@@ -0,0 +1,139 @@
...
+            Err(VmError::Backend {
+                op,
+                code: ret.unsigned_abs() as i32,
+            })
+        } else {
+            Ok(ret)
+        }
+    }
+}
+
```
evanspearman, line 75: A few things here:
- Generally better to just make whole modules `#[cfg(...)]` rather than scattering it about, goes throughout this PR
- Inline isn't useful here. What inline actually does is add the body to crate metadata so that the compiler can choose to inline it when called from other crates. It does nothing here since this method is `pub(crate)`. `#[inline(always)]` forces it, but generally modern compilers are good at knowing when and when not to inline so I would not do this unless we have an actual profiled performance issue
- Having this as a method on the error type itself rather than a free floating function in module with the ffi code feels weird to me, but it isn't per-say wrong to do it this way
- `ret.unsigned_abs() as i32`: I don't really see any reason to convert this back to `i32` since the value will never be negative, so no reason to need to represent negative values. The unsigned integer that `unsigned_abs` returns is more correct. Even better would be:

```rs
let errno = ret.checked_neg().unwrap_or(libc::EOVERFLOW);
Err(VmError::Backend { op, source: io::Error::from_raw_os_error(errno) })
```

evanspearman (changes requested):

norrietaylor:
@evanspearman thanks for the feedback. For now I am going to close this PR in favour of a krunkit exploration spike. I will incorporate your feedback for future PRs.

crates/minvmd/src/error.rs
norrietaylor: Addressed in f868ce41:
- Moved `check_backend` off `VmError` into the `krun` FFI module as a free function — it lives next to the bindings that produce the negative-errno convention.
- Dropped the `#[inline]`.
- Removed the scattered `#[cfg(target_os = "macos")]`: the helper now sits entirely inside the already-macOS-gated `krun` module, so there's nothing to gate in `error.rs` anymore.
- Switched `Backend` from `code: i32` to `source: io::Error` built via `io::Error::from_raw_os_error(ret.checked_neg().unwrap_or(libc::EOVERFLOW))`, and wired up `Error::source()`. No more `unsigned_abs() as i32`.

crates/minvmd/src/bin/krun_smoke_child.rs
norrietaylor: Fixed in f868ce41: `run_macos` and `main` now return `Result<(), VmError>` and every `Context` call propagates with `?` instead of `expect()`.

crates/minvmd/src/bin/krun_smoke_child.rs
norrietaylor: Kept as-is (explicit non-macOS early return), now via `return Ok(())` since `main` returns `Result`.

norrietaylor:
@twitchyliquid64 on the `libkrun` crate: I looked at it, but the crates.io `libkrun` **is** the VMM itself — it compiles libkrun from source (pulls `krun-hvf`, `kvm-bindings`, etc.) and exposes the C ABI. It's not a thin bindings crate, so depending on it would mean building the whole hypervisor backend into our tree rather than linking the prebuilt `libkrun.dylib` that ships via Homebrew.

The canonical `containers/krunkit` tool links the prebuilt dylib via `#[link(name = "krun")] extern "C"` (`src/context.rs`) with `build.rs` emitting `rustc-link-search={PREFIX}/lib` — no crate dep, no dlopen. `minvmd` now mirrors that exactly. The committed bring-up spec (`docs/specs/02-spec-minvmd-bringup/`) records this as the chosen FFI strategy.

Evan's review points are all addressed in the latest push; threads replied inline.

norrietaylor:
@coderabbitai review

evanspearman (approved):

PR #253, opened 2026-05-29
Minimald scaffolding, including sessions actor, session actor, session store + generics, and RPC tests 

twitchyliquid64:
## Summary by CodeRabbit

* **New Features**
  * Session management: create, list, lookup, and run per-session actors with workspace paths
  * Typed SSH RPCs: GetVersion, ListSessions, GetSessionRecord with async request/response handling
  * Test harness for end-to-end RPC and session integration tests

* **Infrastructure**
  * Consistent UTF‑8 realm-tagged path types for daemon state/cache and CLI helpers
  * Disk-backed session store with atomic on-disk persistence and safe rename behavior on Linux

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/253?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/minimald/src/session.rs
+++ b/crates/minimald/src/session.rs
@@ -1,0 +1,58 @@
+use sessions::{paths::DaemonAbsPath, store::SessionObject};
+use tokio::sync::{mpsc, oneshot};
+
+enum SessionMessage {
+    GetWorkspacePath(oneshot::Sender<DaemonAbsPath>),
+}
+
+/// Manages a running session.
+///
+/// Follows the actor pattern.
+#[derive(Debug)]
+pub struct Session<S: SessionObject> {
+    receiver: mpsc::Receiver<SessionMessage>,
+    session: S,
+}
+
+impl<S: SessionObject> Session<S> {
+    /// Launches the actor for the given session.
+    pub async fn run(session: S) -> Result<SessionHandle, std::io::Error> {
+        let (sender, receiver) = mpsc::channel(8);
+        let mngr = Self { receiver, session };
+
+        tokio::spawn(mngr.mainloop());
+        Ok(SessionHandle(sender))
+    }
+
+    /// The async task which handles interactions with the
+    /// session.
+    async fn mainloop(mut self) {
+        while let Some(msg) = self.receiver.recv().await {
+            self.handle_message(msg).await;
+        }
+    }
+    /// Handles a specific message recieved by the session.
+    async fn handle_message(&mut self, msg: SessionMessage) {
+        match msg {
+            SessionMessage::GetWorkspacePath(r) => {
+                let wsp = self.session.workspace_path();
+                std::fs::create_dir_all(&wsp).unwrap();
+                let _ = r.send(wsp);
+            }
+        }
+    }
+}
+
+/// The handle to the session.
+#[derive(Debug, Clone)]
+pub struct SessionHandle(mpsc::Sender<SessionMessage>);
+
+impl SessionHandle {
+    /// Returns the path where the session workspace is located.
+    pub async fn workspace_path(&self) -> DaemonAbsPath {
+        let (send, recv) = oneshot::channel();
+        // Ignore send errors - the recv will also fail.
+        let _ = self.0.send(SessionMessage::GetWorkspacePath(send)).await;
+        recv.await.expect("corresponding session is dead")
+    }
+}
--- a/crates/minimald/src/sessions.rs
+++ b/crates/minimald/src/sessions.rs
@@ -22,13 +22,32 @@ pub enum SessionKeyPredicate {
     Name(String),
 }
 
+/// Transport / internal error when communicating with the sessions actor.
+pub type ResponseError = std::io::Error;
+
+/// Encapsulates the return channel for messages back from the actor.
+struct Responder<T>(oneshot::Sender<Result<T, ResponseError>>);
+
+impl<T> Responder<T> {
+    /// Constructs both ends of the return channel.
+    pub fn channel() -> (Self, oneshot::Receiver<Result<T, ResponseError>>) {
+        let (send, recv) = oneshot::channel();
+        (Self(send), recv)
+    }
+
+    /// Awaits the provided future, transmitting its result to the caller.
+    pub async fn handle<F>(self, fut: F)
+    where
+        F: Future<Output = Result<T, ResponseError>>,
+    {
+        let _ = self.0.send(fut.await);
+    }
+}
+
 enum ManagerMessage {
-    List(oneshot::Sender<Vec<SessionInfo>>),
-    GetRecord(
-        SessionKeyPredicate,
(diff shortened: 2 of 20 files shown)


crates/minimald/src/session.rs
```diff
@@ -0,0 +1,58 @@
...
+        while let Some(msg) = self.receiver.recv().await {
+            self.handle_message(msg).await;
+        }
+    }
+    /// Handles a specific message recieved by the session.
+    async fn handle_message(&mut self, msg: SessionMessage) {
+        match msg {
+            SessionMessage::GetWorkspacePath(r) => {
+                let wsp = self.session.workspace_path();
+                std::fs::create_dir_all(&wsp).unwrap();
```
evanspearman, line 39: I think we need to make this method fallible and propagate errors. We're definitely going to have other message handlers that can fail.

crates/minimald/src/sessions.rs
```diff
@@ -0,0 +1,161 @@
...
+                        .collect(),
+                );
+            }
+            // Gets the record for a specific session.
+            ManagerMessage::GetRecord(pred, r) => {
+                let _ = r.send(match pred {
+                    SessionKeyPredicate::Id(id) => self
+                        .store
+                        .find_by_uuid(&id)
+                        .unwrap()
```
evanspearman: If the unwraps in here are due to an invariant enforced elsewhere we should document that using expect() rather than unwrap. It's not clear to me that these are invariants though, so it might make more sense to do something like:

```rs
﻿﻿fn key_for(&self, pred: &SessionKeyPredicate) -> Result<Option<Key>, StoreError> {
      match pred {
          SessionKeyPredicate::Id(id)     => self.store.find_by_uuid(id),
          SessionKeyPredicate::Name(name) => self.store.find_by_name(name),
      }
}

//...

ManagerMessage::FindRecord(pred, r) => {
      let record = self
          .key_for(&pred)?
          .map(|k| self.store.get(&k).map(|e| e.record().clone()))
          .transpose()?;
      let _ = r.send(record);
  }

  ManagerMessage::GetSession(pred, r) => {
      let handle = match self.key_for(&pred)? {
          None => None,
          Some(k) => Some(match self.running.get(&k) {
              Some(h) => h.clone(),
              None => {
                  let entry = self.store.get(&k)?;
                  let h = Session::run(entry).await?;
                  self.running.insert(k, h.clone());
                  h
              }
          }),
      };
      let _ = r.send(handle);
  }
```

crates/sessions/src/paths.rs
```diff
@@ -194,6 +194,29 @@ impl<R: Realm> AbsPath<R> {
         }
     }
 
+    /// Joins a given name *in the same realm*, producing a new
+    /// [`AbsPath<R>`].
+    ///
+    /// We intentionally take a &'static str to make it difficult
+    /// to call this method with anything but a single dir/file.
```
evanspearman: I don't really understand why that would be the case. `&'static str` in theory makes it less likely that violating this invariant would happen outside of development time, although, doesn't prevent it.

crates/sessions/src/paths.rs
```diff
@@ -194,6 +194,29 @@ impl<R: Realm> AbsPath<R> {
...
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the directory name includes
+    /// nested directories, or attempts to traverse.
+    pub fn sub_path(&self, dir: &'static str) -> AbsPath<R> {
```
evanspearman: I don't really like doing these as runtime checks since I think the intention is that we want an easy way to do this for sub-paths that we know are okay at compile time. An alternative approach that is more idiomatic:

```rs
  #[doc(hidden)]
  pub const fn _validate_subdir(s: &str) {
      let bytes = s.as_bytes();
      if bytes.is_empty() {
          panic!("`sub_path` component must not be empty");
      }
      let mut i = 0;
      while i < bytes.len() {
          if bytes[i] == b'/' {
              panic!("`sub_path` component must not contain `/`");
          }
          i += 1;
      }
      if bytes.len() == 2 && bytes[0] == b'.' && bytes[1] == b'.' {
          panic!("`sub_path` component must not be `..`");
      }
  }

  impl<R: Realm> AbsPath<R> {
      /// Joins a single sub-component. Trusts the caller; prefer the
      /// [`sub_path!`] macro at literal call sites so the constraint is
      /// enforced at compile time.
      pub fn sub_path_unchecked(&self, dir: &str) -> AbsPath<R> {
          AbsPath {
              inner: self.inner.join(dir),
              _realm: PhantomData,
          }
      }
  }

  #[macro_export]
  macro_rules! sub_path {
      ($base:expr, $dir:literal) => {{
          const _: () = $crate::paths::_validate_subdir($dir);
          ($base).sub_path_unchecked($dir)
      }};
  }
```

You would then use it like:

```rs
let sessions = sub_path!(state_dir, "sessions");   // OK
let bad      = sub_path!(state_dir, "a/b");        // compile error
let bad      = sub_path!(state_dir, "..");         // compile error
let bad      = sub_path!(state_dir, "");           // compile error
```

crates/sessions/src/paths.rs
```diff
@@ -235,6 +258,23 @@ impl<R: Realm> AbsPath<R> {
     pub fn components(&self) -> Utf8Components<'_> {
         self.inner.components()
     }
+
+    /// Returns an absolute path representing the current
```
evanspearman, line 281: Going to put up a PR soon based on the need for this.

crates/sessions/src/store.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// Returns an I/O error if the session directory, record, or index
+    /// cannot be written.
+    fn create(&mut self, record: Record) -> Result<Self::Key, std::io::Error>;
+}
+
+/// The concrete key used to identify sessions from [`DiskLoader`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DiskSessionKey {
+    session_uuid: Uuid,
+    dir_key: String,
```
evanspearman: Probably better to just store `DaemonRelPath` here which removes the unwrap later on.

evanspearman (changes requested):

crates/sessions/src/paths.rs
twitchyliquid64: 100%, you could still use unsafe or `Box::leak` but should be hard to do accidentally. In any case, I'll switch over to the const approach you used below.

crates/minimald/src/sessions.rs
twitchyliquid64: I was just being lazy - will properly handle those cases.

twitchyliquid64: I ended up revamping the model a bit by having a generic type that can carry these errors, ptal.

evanspearman (approved):

PR #256, opened 2026-05-29
feat(minvmd): crate skeleton + CLI

norrietaylor:
Stack 1/3 for the `minvmd` macOS host broker.

Adds `crates/minvmd`: workspace member, clap CLI (`up`/`down`/`status`/`debug-shell`), local user-facing error type, daemon `Config` (cpus/memory/state-disk), and `SPEC.md` (landed shape). Subcommands are stubbed; image/lifecycle/broker land in 2/3 and 3/3. macOS-gated with a Linux no-op shim; unit tests run on Linux CI.

`cargo {build,test,clippy,fmt} -p minvmd` green (4 tests).

## Summary by CodeRabbit

* **New Features**
  * Introduced `minvmd`, a new macOS-based tool for virtual machine management, providing CLI commands for VM lifecycle operations (boot, shutdown, status monitoring, debugging).
  * Added configurable VM resource settings: CPU count, memory size, and storage allocation with sensible defaults.

* **Documentation**
  * Published specification detailing the minvmd integration with VM image artifacts and operational workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gominimal/minimal/pull/256?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/minvmd/src/config.rs
+++ b/crates/minvmd/src/config.rs
@@ -1,0 +1,108 @@
+//! `Config` — minvmd daemon configuration.
+//!
+//! Reads `~/.minimal/vm/minvmd.toml`; falls back to defaults when the file is
+//! absent. Unknown keys are silently ignored (forward-compatible).
+
+use std::path::Path;
+
+use serde::Deserialize;
+
+// TODO(unit-2): wire into lifecycle::up. `Config` is consumed by
+// image::ensure_image in Sub-task 2.1 and lifecycle in Sub-task 3.1. Until
+// those land, only tests construct it, so the bin target flags dead code.
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+#[serde(default)]
+pub struct Config {
+    pub cpus: u32,
+    pub memory_mib: u64,
+    pub state_size_gib: u32,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            cpus: 2,
+            memory_mib: 2048,
+            state_size_gib: 10,
+        }
+    }
+}
+
+#[allow(dead_code)]
+impl Config {
+    /// Load config from `path`. If the file does not exist, returns defaults.
+    pub fn load(path: &Path) -> anyhow::Result<Self> {
+        match std::fs::read_to_string(path) {
+            Ok(contents) => Ok(toml::from_str(&contents)?),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(Self::default()),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    /// Load from the canonical path `$MINIMAL_HOME/vm/minvmd.toml`.
+    /// `$MINIMAL_HOME` defaults to `~/.minimal` when unset.
+    pub fn load_default() -> anyhow::Result<Self> {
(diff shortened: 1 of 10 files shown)


crates/minvmd/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+// image::ensure_image in Sub-task 2.1 and lifecycle in Sub-task 3.1. Until
+// those land, only tests construct it, so the bin target flags dead code.
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+#[serde(default)]
+pub struct Config {
+    pub cpus: u32,
+    pub memory_mib: u64,
+    pub state_size_gib: u32,
+}
```
evanspearman, line 20: The coderabbit suggestion here is valid, but an even better solution (by a lot) would be to use newtypes to wrap the types to enforce the invariants at that level. The nutype crate is really nice for trivial invariants like this as it lets you do it in a declarative way with macros.

Number of CPUs could in theory be a u16 safely in theory, _but_ alignment is also a thing, so there's not much point unless there is another 16-bit field we might want to add at some point. Either way, not going to make a material difference.

crates/minvmd/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+        Self {
+            cpus: 2,
+            memory_mib: 2048,
+            state_size_gib: 10,
+        }
+    }
+}
+
+#[allow(dead_code)]
+impl Config {
```
evanspearman, line 33: coderabbit suggestion is valid here

crates/minvmd/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Load config from `path`. If the file does not exist, returns defaults.
+    pub fn load(path: &Path) -> anyhow::Result<Self> {
```
evanspearman, line 35: We can always change this later, but since this is a user-facing CLI we may want to adopt color_eyre instead of anyhow since you get better error messages from it.

crates/minvmd/src/error.rs
```diff
@@ -0,0 +1,50 @@
...
+//! shared one.
+//!
+//! Use `anyhow::Error::from(UserFacing::new("message"))` for errors that should
+//! be printed verbatim to the user. For internal/system errors, propagate with
+//! bare `?`.
+
+use std::fmt;
+
+#[derive(Debug)]
+pub struct UserFacing {
```
evanspearman, line 14: Not sure I really understand how this is supposed to be used.

crates/minvmd/src/test_support.rs
```diff
@@ -0,0 +1,44 @@
...
+//! pieces they use are inlined here rather than pulled from a shared crate,
+//! compiled only under `cfg(test)`.
+
+#![cfg(test)]
+
+use std::path::Path;
+use std::sync::Mutex;
+use tempfile::TempDir;
+
+/// Re-export so test modules can write `test_support::serial_test::serial`.
```
evanspearman, line 13: I think we should be writing testable units in such a way that we just pass the minimal home into them, and then we don't need this.

evanspearman (changes requested):

PR #262, opened 2026-05-29
docs: add session domain model diagram

norrietaylor:
Adds `docs/session-domain-diag.md`: domain model for the session architecture.

Covers:
- Session (workload) / Provider (creator + host) vernacular
- Class diagram of `Minimal` client → `Socket` discovery → `Provider` implementations (`minimald`, `minvmd`, hosted, cloud)
- Local Linux deployment flow
- Local macOS deployment flow (no native `minimald`; sessions reached via VM-hosted `minimald` proxied to host)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Session domain documentation: clear terminology (Session, Provider), how discovery and connections work, architecture diagrams, platform deployment scenarios (Linux/macOS), VM/socket proxying and lifecycle behavior, startup/bootstrap rules (scan/start providers), stale-socket handling, and a glossary clarifying hosting lifecycles and future backend placeholders.

### DIFF
--- a/docs/session-domain-diag.md
+++ b/docs/session-domain-diag.md
@@ -14,7 +14,6 @@ classDiagram
 
     class Socket {
         <<connection>>
-        +path: $HOME/.minimal/local/[pid].sock
     }
 
     class Provider {
@@ -51,10 +50,10 @@ classDiagram
         <<workload>>
     }
 
-    Minimal "1" --> "*" Socket : discovers via local dir
+    Minimal "1" --> "*" Socket : discovers
     Socket "1" --> "1" Provider : connects to
     Minimal "1" ..> "*" Provider : connects to many
-    Provider "1" --> "1" Socket : creates [pid].sock
+    Provider "1" --> "1" Socket : creates
 
     Provider <|.. Minimald : implements
     Provider <|.. MinVMD : implements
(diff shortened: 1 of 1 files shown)


docs/session-domain-diag.md
```diff
@@ -0,0 +1,174 @@
...
+classDiagram
+    class Minimal {
+        <<client process>>
+        +instantiate()
+        +discoverSocket()
+    }
+
+    class Socket {
+        <<connection>>
+        +path: $HOME/.minimal/local/[pid].sock
```
0chroma: On linux at least, might be worth considering putting this in `XDG_RUNTIME_DIR` to be XDG compliant, see https://wiki.archlinux.org/title/XDG_Base_Directory

since that dir is created with strict perms and is usually on tmpfs, runtime sockets won't risk polluting the filesystem or leaking to other users, will get cleaned up if the user's session dies. Also more predictable, in ~/.minimal someone may delete it thinking it's a stray config file.

norrietaylor: Good call. XDG_RUNTIME_DIR is the place for this on linux. On macOS it is less clear.

For now, I am going to remove references to definitive socket paths in this PR. It is an implementation detail and specifying it is not my goal right now

twitchyliquid64 (approved):

docs/session-domain-diag.md
twitchyliquid64: In the current code the socket lives with the provider config at `$XDG_STATE_DIR/minimal/providers/<name>/ssh.sock` Now would be the best time to change it!

https://github.com/gominimal/minimal/blob/3d4ce47849a41b5c6513f98ff5a0cff84a2bd79c/crates/minimald/src/main.rs#L27-L64

PR #265, opened 2026-05-29
feat(minimald): implement RPC for creating sessions, index by session name also

twitchyliquid64:
* Session names are now unique, and `DiskLoader` indexes by name also
 * Implements `CreateSession` RPC.

This is technically breaking the format of `index.json`, except there was never a path to write it before, so it should be empty.

## Summary by CodeRabbit

* **New Features**
  * Sessions can now be created remotely via the SSH RPC interface for programmatic session management.

* **Improvements**
  * Session names must be unique; creating a session with a duplicate name is rejected.
  * Session storage and indexing improved for more reliable listing and lookups.
  * Session records missing an ID are now accepted and given a default ID.

* **Chores**
  * Added a dependency used by session handling.

* **Tests**
  * New/updated tests validate session creation, storage, and listing behavior.

### DIFF
--- a/crates/sessions/src/store.rs
+++ b/crates/sessions/src/store.rs
@@ -2,6 +2,7 @@
 
 use std::collections::BTreeMap;
 
+use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
 use crate::paths::{DaemonAbsPath, DaemonRelPath};
@@ -106,6 +107,21 @@ impl SessionObject for DiskSession {
     }
 }
 
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+struct Index {
+    short_to_uuid: BTreeMap<String, Uuid>,
+    name_to_uuid: BTreeMap<String, Uuid>,
+}
+
+impl Index {
+    pub fn insert(&mut self, short: String, uuid: Uuid, name: Option<String>) {
+        self.short_to_uuid.insert(short, uuid);
+        if let Some(name) = name {
+            self.name_to_uuid.insert(name, uuid);
+        }
+    }
+}
+
 /// A loader of session state based on <minimal-state-dir>/sessions.
 ///
 /// ./index.json maps short directory names to session UUIDs. Typically
@@ -115,7 +131,7 @@ impl SessionObject for DiskSession {
 /// ./<short-dir-name>/record.json is the session record.
 pub struct DiskLoader {
     minimal_dir: DaemonAbsPath,
-    index: BTreeMap<String, Uuid>,
+    index: Index,
 }
 
 impl DiskLoader {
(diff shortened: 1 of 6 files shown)


crates/sessions/src/store.rs
```diff
@@ -200,15 +225,19 @@ impl Loader for DiskLoader {
...
-        self.index.iter().map(|(short, id)| Self::Key {
-            session_uuid: *id,
-            dir_key: DaemonRelPath::try_new(short).unwrap(),
-        })
+        self.index
+            .short_to_uuid
+            .iter()
+            .map(|(short, id)| Self::Key {
+                session_uuid: *id,
+                dir_key: DaemonRelPath::try_new(short).unwrap(),
```
evanspearman, line 233: I suspect this is something that's going to keep coming up. We probably want to newtype `Uuid` as SessionId (this would be good regardless) and either implement a `to_daemon_rel_path` on it.

We should probably also add `new_unchecked` constructors (with big warnings on them) for the path types and just be really careful about where we use them.

Don't think we can do `From<SessionId> for DaemonRelPath` since we were talking about moving the `paths` module to its own crate and the `sessions` crate will be downstream of that.

evanspearman: Slightly out of scope but we probably want to rename this method to `keys` which is more standard.

crates/sessions/src/store.rs
```diff
@@ -115,7 +131,7 @@ impl SessionObject for DiskSession {
 /// ./<short-dir-name>/record.json is the session record.
 pub struct DiskLoader {
     minimal_dir: DaemonAbsPath,
-    index: BTreeMap<String, Uuid>,
+    index: Index,
```
evanspearman, line 134: Just to clarify, is this going to be always kept in sync or is it going to be lazy? If lazy, we probably want to wrap it in `RefCell` so any future read-only methods don't need to take `&mut self`. Looks like it's kept in sync so far, just wanted to check.

crates/sessions/src/store.rs
```diff
@@ -106,6 +107,21 @@ impl SessionObject for DiskSession {
...
 }
 
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+struct Index {
+    short_to_uuid: BTreeMap<String, Uuid>,
+    name_to_uuid: BTreeMap<String, Uuid>,
+}
+
+impl Index {
+    pub fn insert(&mut self, short: String, uuid: Uuid, name: Option<String>) {
```
evanspearman, line 117: Minor, but do we want search to live in the index type as well?

evanspearman (approved):

twitchyliquid64:
Thanks! Given the dep from a later PR I'm going to merge this as-is and send a follow-up PR for the changes (newtype `SessionId`, keys() etc)

PR #273, opened 2026-06-01
chore(sessions): Implemented var and var policy primitive types + code ergonomics

evanspearman:
- Dropped the provenance stuff from primitives, that stuff should be injected later
- Added implementation for variables
    - They can be either: specified, inherited, or inherited with a default (inherited will error if it is not defined)
    - Two versions: strict and lenient: This is because POSIX is strict about variable names but Linux is more lenient. In most cases we want the POSIX checking as this helps produce better error messages for certain types of typos, but we still need the lenient escape hatch
    - The different shape is to make it serialize, deserialize to more ergonomic toml
- Bunch of other tweaks to other primitives for code ergonomic reasons (mostly builder pattern stuff)
- Added a primitive for user policies since we have policies for vars and patches now

## Release Notes

* **New Features**
  * Sessions now support package installation alongside existing features
  * Added optional descriptions to lifecycle hooks and session loadouts
  * Introduced flexible variable handling with strict and lenient variable types
  * New policy system for controlling environment variable and file patch configurations
  * Enhanced builder-style APIs for constructing and modifying session configurations

### DIFF
--- a/crates/sessions/src/loadout.rs
+++ b/crates/sessions/src/loadout.rs
@@ -1,13 +1,350 @@
+//! User-level loadout: packages, variables, patches, and lifecycle
+//! hooks the user wants in their session.
+//!
+//! # Example: helix + zellij with the user's dotfiles
+//!
+//! Top-level keys (`packages`, `patches`) come first because every key
+//! following a `[section]` header in TOML is scoped to that section
+//! until the next header; an empty line is *not* a scope reset.
+//!
+//! ```toml
+//! packages = ["helix", "zellij"]
+//!
+//! # `dest` paths are inside the sandbox; `source` paths are on the host.
+//! # `~` on either side is expanded at apply time (sandbox-home for dest,
+//! # host-home for source).
+//! patches = [
+//!     # Helix: single config files plus a themes directory.
+//!     { dest = "~/.config/helix/config.toml",
+//!       source = "~/dotfiles/helix/config.toml" },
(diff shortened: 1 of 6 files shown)


crates/sessions/src/loadout.rs
```diff
@@ -1,13 +1,350 @@
...
+//!
+//!     # Zellij: single config file plus a layouts directory.
+//!     { dest = "~/.config/zellij/config.kdl",
+//!       source = "~/dotfiles/zellij/config.kdl" },
+//!     { dest = "~/.config/zellij/layouts/",
+//!       source = { base = "~/dotfiles/zellij/layouts", patterns = ["**/*.kdl"] } },
+//! ]
+//!
+//! [vars]
+//! EDITOR    = "hx"
```
twitchyliquid64, line 33: Some of these are kinda special (EDITOR, TERM, PAGER etc). Do we want to elevate them to their own thing in the config? Do we want to be able to set a default (separate to the `default` def in this file) if not set?

twitchyliquid64 (approved):

crates/sessions/src/loadout.rs
evanspearman: I don't think that's necessary. Home-manager just treats them like normal environment variables, so I think doing the same is fine for us. Typically applications use fallbacks if these are not set. I don't think treating these like other environment variables precludes us from setting our own defaults either though.

PR #275, opened 2026-06-01
feat!(sessions): newtype session ID, refactor store index

twitchyliquid64:
Implements some followups from #265, most notably a newtype for session IDs.

## Summary by CodeRabbit

* **Refactor**
  * Session identifiers unified to a domain-specific SessionId across RPC, session manager, SFTP, and storage for stronger typing and consistent behavior.
* **Bug Fixes**
  * Improved session creation, lookup, and SFTP handling with clearer failure logging and more robust ID resolution.
* **Tests**
  * Updated tests to cover the new SessionId behavior, including creation, listing, lookup, and persistence.

### DIFF
--- a/crates/sessions/src/lib.rs
+++ b/crates/sessions/src/lib.rs
@@ -2,6 +2,7 @@
 //! shape of a Minimal session.
 
 use std::collections::BTreeMap;
+use std::fmt;
 
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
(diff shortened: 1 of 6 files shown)


crates/sessions/src/lib.rs
```diff
@@ -15,12 +16,45 @@ pub mod paths;
...
+    }
+
+    /// Parses the given UUID as a session ID.
+    pub fn parse_str(s: &str) -> Result<Self, uuid::Error> {
+        Uuid::parse_str(s).map(Self)
+    }
+
+    /// Constructs a string representing the given UUID.
+    #[must_use]
+    pub fn to_string(&self) -> String {
```
evanspearman: FYI: when you implement `Display`, `ToString` is implemented automatically (and is actually the preferred way to implement `ToString` over manually doing a `impl ToString for MyType`.

evanspearman (approved):

crates/sessions/src/lib.rs
twitchyliquid64: Made `Display` be a transparent impl of UUID's display, and removed `to_string()`

PR #281, opened 2026-06-02
feat(minimald): implement scaffolding for execution over ssh

twitchyliquid64:
The main piece is the new `ExecTask` struct, which represents and async task & encapsulates the logic gluing a process to an ssh channel. `ExecTask` is generic over a `Spawnable` trait, which is something that can build a process (in the future, this can also be sandboxed processes ala minimal tasks).

For now I've just got it running unsandboxed processes as a simple way to test - we won't be doing this long term.

I'm also thinking in the long term `ExecTask` can probably be an actor, and recieve messages to do things like kill its process etc. And we could have the closure in `spawn()` register the `ExecTask` with a session and/or connection.

## Summary by CodeRabbit

* **New Features**
  * Implemented SSH exec request handling: remote command execution with stdin/stdout/stderr streaming, working-directory and environment support, and exit status reporting.

* **Tests**
  * Added unit and end-to-end tests covering stdio bridging, exit-code behavior, error handling, and a test harness helper to exercise exec requests in tests.

### DIFF
--- a/crates/minimald/src/exec.rs
+++ b/crates/minimald/src/exec.rs
@@ -1,0 +1,757 @@
+//! SSH `exec` request handling: spawns a process and bridges its stdio
+//! over an SSH channel.
+//!
+//! Process spawning is abstracted via the [`Spawnable`] / [`Process`]
+//! traits so the bridge logic can be exercised with a mock in tests
+//! without going through `tokio::process` or russh.
+
+use std::collections::BTreeMap;
+use std::fmt::Display;
+use std::io;
+use std::process::Stdio;
+
+use paths::DaemonAbsPath;
+use russh::{
+    Channel, ChannelId,
+    server::{Msg, Session},
+};
+use sessions::SessionId;
+use tokio::{
+    io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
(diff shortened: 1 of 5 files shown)


crates/minimald/src/exec.rs
```diff
@@ -0,0 +1,757 @@
...
+
+    /// Returns all three stdio handles the first time it's called,
+    /// `None` thereafter (per-handle `Option::take` semantics, matching
+    /// `tokio::process::Child`).
+    fn take_stdio(&mut self) -> Option<(Self::Stdin, Self::Stdout, Self::Stderr)>;
+
+    /// Resolves when the process exits. `Ok(None)` means the process
+    /// was terminated by a signal rather than exiting with a numeric
+    /// code.
+    async fn wait(&mut self) -> io::Result<Option<i32>>;
```
evanspearman, line 62: Would there be any reason we might want to capture the specific signal?

evanspearman (approved):

crates/minimald/src/exec.rs
twitchyliquid64: Yes, on one side we need to communicate the exit status down the ssh channel, but also, every process needs to be wait()’ed on to avoid zombie processes.

PR #292, opened 2026-06-02
feat(mctx)!: configure separate state and cache directories

twitchyliquid64:
Under Minimal One, we are going to have minimal state (like sandboxes, sessions etc) live in one directory (typically `$XDG_STATE_DIR/minimal` or `$HOME/.local/state/minimal`) and cached objects (like the artifact cache, fetched sources etc) live in another (`$XDG_CACHE_DIR/minimal` or `$HOME/.local/cache/minimal`). This separation better follows the XDG specification.

The default remains the same when the directories are not specified via the `mctx::ConfigBuilder`. We should pick a date in the future to make a hard switch.

## Summary by CodeRabbit

* **Refactor**
  * Separated state and cache storage into distinct base directories and updated local built-artifact cache handling. This clarifies where state vs. cache data live, improves isolation of cached artifacts, and makes cache initialization and access more predictable.

### DIFF
--- a/crates/mctx/src/config.rs
+++ b/crates/mctx/src/config.rs
@@ -49,9 +49,11 @@ pub struct ConfigBuilder {
     offline: Option<bool>,
     num_parallel_builds: Option<usize>,
 
-    minimal_dir: Option<PathBuf>,
+    minimal_state_dir: Option<PathBuf>,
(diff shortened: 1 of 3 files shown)


norrietaylor (approved):
LGTM

PR #362, opened 2026-06-07
[WIP] feat(minvmd,minimald): Stage 2 — minimald as guest pid-1 (e2e integration)

norrietaylor:
**Draft integration branch** combining #359 (block-root boot + rootfs package + RAW kernel) and #361 (vsock-native minimald), running `minimald` as the guest pid-1 with a full session over the libkrun bridge. Lands after #359 and #361.

## Working end-to-end (Rust e2e test, exit-status accurate)
`minimald_exec_over_bridge` (gated `MINVMD_E2E=1`) drives a real **russh client** over a `UnixStream` to the bridge UDS and exercises the whole stack:
```
auth_none → CreateSession (SSH subsystem RPC) → exec "echo …" in the session
→ stdout="MINIMALD_SESSION_OK\n"  exit=Some(0)
```
i.e. host UDS → libkrun bridge → guest vsock → socat relay → minimald SSH → session exec, with correct stdout **and exit status**.

Foundation (all console/test-validated):
- `minimald` cross-compiles to `aarch64-unknown-linux-musl` (static).
- minvmd provisions + attaches a persistent rw data disk (`/dev/vdb`, sparse, guest-formatted); rootfs bakes the `/var/lib/minimal` mountpoint.
- minimald boots as guest pid-1: mounts `/proc`+`/sys`, formats+mounts the data disk, emits `READY`, serves SSH over a `socat` vsock→UDS relay into `run_on_uds`.

## Bugs found + fixed via the e2e
- **pid-1 reaper stole exec exit status**: the eager `waitpid(-1)` SIGCHLD reaper raced tokio's process reaping (ECHILD → every command reported exit 1). Dropped it; tokio reaps its own children. Hakoniwa orphan reaping deferred (spec).
- **"bridge interop bug" was a test artifact**: CLI tools (`ssh`/`socat`/`nc`) don't round-trip the bridge, and there's a post-READY startup race. A Rust client that retries works reliably — no libkrun/tokio-vsock defect.

## Transport
Guest serves a UDS on the data disk fronted by a `socat` vsock→UDS relay (the `bridge_e2e`-proven vsock-listen + minimald's native `run_on_uds`). Direct `tokio_vsock` (`run_on_vsock`, retained as pub API) is unproven against this bridge and could replace the relay later.

## Remaining (smaller now)
- Orphan reaping compatible with tokio (for `min run`/hakoniwa under net=none).
- Client/RPC wiring on the host CLI (`minimal2 ls` is a placeholder).
[...]

PR #367, opened 2026-06-08
feat(minvmd): boot a microVM from kernel + rootfs built as minimal packages

norrietaylor:
> **Stage 1** — the guest runs a socat bring-up stub (`/sbin/microvm-init`), not minimald. minimald-as-pid-1 is Stage 2 ([#362](https://github.com/gominimal/minimal/pull/362)).

## What changed

- **Boots a real Linux microVM from artifacts the build system produces itself** — removes the `gs://minimal-shim` CLI download and the pinned-Alpine apk overlay.
- **Kernel + rootfs are now upstream `minimal` packages** (`virtio-kernel-raw` + `microvm-rootfs`), cache-pulled by the pinned `locked_commit`.
- **Ext4 block root instead of virtiofs** — loaded via `krun_add_disk2` (`/dev/vda`) with a kernel `init=` cmdline (a block root has no libkrun `/init.krun`).
- **Faster boot** — ships the kernel uncompressed (raw `Image`), skipping libkrun's in-VMM gzip decompress (~77 ms); boot-to-READY ~113 ms median on CI.

## How to run on Mac

Prereqs: Apple Silicon · `brew install slp/krun/libkrun` · the `minimal` shim on `PATH`.

```sh
mkdir -p .scratch
minimal materialize --output .scratch/vmlinuz    --arch aarch64 virtio-kernel
minimal materialize --output .scratch/rootfs.img --arch aarch64 minvmd-rootfs

cargo build -p minvmd --bin minvmd -p minimal2 --bin minimal2
# codesign LAST — a later cargo run/test relinks + unsigns it (krun_start_enter -> EINVAL)
codesign --entitlements crates/minvmd/minvmd.entitlements --force -s - target/debug/minvmd

export PATH="$PWD/target/debug:$PATH"
export MINVMD_KERNEL_PATH="$PWD/.scratch/vmlinuz"
export MINVMD_ROOTFS_PATH="$PWD/.scratch/rootfs.img"
minimal2 ls        # cold: auto-spawns minvmd, boots the VM, prints []
minvmd status      # running
minvmd stop

# boot-to-READY latency (min/median/max over N runs, default 10)
scripts/bench-minvmd-boot.sh
```

`--output` must be under the repo (the shim only syncs the project dir back to the host).


crates/minvmd/src/krun/ctx.rs
```diff
@@ -214,6 +214,36 @@ impl Context {
         Ok(())
     }
 
+    /// Add a disk image as a virtio-blk block device backing `block_id`.
+    /// `disk_format` is one of `raw::KRUN_DISK_FORMAT_*`.
+    pub fn add_disk(
```
evanspearman, line 219: Not requesting a change here, just want to point out a convention that's common for these sorts of interfaces in Rust that you might want to consider in the future.

Instead of something like:

```rs
let mut thing = Thing::new();
thing.set_whatever(whatever)?;
thing.set_otherthing(otherthing)?;
thing.set_infallible(infallible);
```

Something like the following usage is preferred:

```rs
let thing = Thing::new()
    .with_whatever(whatever)?
    .with_otherthing(otherthing)?
    .with_infallible();
```

Or in cases where there are invariants that involve multiple fields:

```rs
let thing = Thing::builder()
    .with_whatever(whatever)
    .with_otherthing(otherthing)
    .with_infallible(infallible)
    .build()?;
```

There are a few different ways to implement this

Consuming immutable form:

```rs
pub fn with_whatever(self, whatever: Whatever) -> Result<Self, Error> {
    Self {
        whatever,
        ..self
    }
}
```

Mutable reference form:

```rs
pub fn with_whatever(&mut self, whatever: Whatever) -> Result<&mut Self, Error> {
    self.whatever = whatever;
    self
}
```

And other forms like consuming mutable, etc. that can be inferred.

The immutable form may seem less efficient, but the compiler actually ends up doing the same thing under the hood for these, so generally I recommend the consuming immutable form unless there's another reason not to.

Again, not asking for a change here, just wanted to point out this convention.

crates/minvmd/src/krun/ctx.rs
```diff
@@ -214,6 +214,36 @@ impl Context {
...
+        &mut self,
+        block_id: &str,
+        disk_path: impl AsRef<Path>,
+        disk_format: u32,
+        read_only: bool,
+    ) -> Result<(), VmError> {
+        let id_cstr = cstring_from_str(block_id, "block_id")?;
```
evanspearman, line 226: FYI: you can do:

```rs
let id_cstr = c"block_id";
```

Or in cases where you need an owned type rather than a static reference:

```rs
let id_cstr = CString::from(c"block_id");
```

crates/minvmd/src/krun/ctx.rs
```diff
@@ -214,6 +214,36 @@ impl Context {
         Ok(())
     }
 
+    /// Add a disk image as a virtio-blk block device backing `block_id`.
+    /// `disk_format` is one of `raw::KRUN_DISK_FORMAT_*`.
+    pub fn add_disk(
+        &mut self,
+        block_id: &str,
+        disk_path: impl AsRef<Path>,
```
evanspearman, line 222: Minor, but for paths we should move towards using the path newtypes in our paths crate which enforce some invariants at compile time. They're also based on `camino::Utf8Path` so we don't need to worry about non-utf8 characters.

crates/minvmd/src/krun/raw.rs
```diff
@@ -22,10 +22,16 @@ use crate::error::VmError;
 // - x86_64  `virtio-linux` ships `bzImage`  → `KRUN_KERNEL_FORMAT_ELF`. libkrun
 //   loads bzImage as ELF; `IMAGE_BZ2` (=3) is bzip2 blobs, not the bzImage
 //   container format.
+pub const KRUN_KERNEL_FORMAT_RAW: u32 = 0;
```
evanspearman, line 25: Just wanted to point out that you can do:

```rs
#[repr(i32)]
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum MyType {
    Whatever = 0,
    Variants = 1,
    You = 2,
    Need = 3,
}
```

Not sure if that's the right call for this case or not, but in cases where you want some additional type safety this is an available tool.

evanspearman (approved):

PR #372, opened 2026-06-09
feat(sessions): patch source/policies home expansion

evanspearman:
Patch source `FileSet` patterns and `PatchPolicy` allow/deny/ignore
  patterns are commonly written `~/`-relative (`source = "~/dotfiles/..."`,
  `deny = ["~/.ssh/**"]`). The walker doesn't interpret `~`, so today
  those silently match nothing — a dotfile patch contributes zero files,
  a `~/.ssh/**` deny rule never fires.

  The resolver now expands `~` against a configurable host home
  (`HomeLookup` — `dirs::home_dir` by default) before walking and before
  policy `check`. Both halves keep their `~/`-relative form in the
  returned policy, so save/load is lossless.

  Behavior:
  - Lookup is lazy: invoked only when at least one `~`-prefixed pattern
    is in scope. A loadout with no `~` anywhere never asks for home.
  - Lookup is re-attempted mid-resolution if the hook adds a `~`-prefixed
    rule and home wasn't previously needed.
  - `PatchDest` is unchanged — destinations are implicitly relative to the
    sandbox user's home and reject `~` / absolute paths at construction.

  Failures surface as `ResolveError::HomeUnresolved(HomeResolutionFailure)`
  distinguishing three causes (`Unavailable`, `NotUtf8`, `NotAbsolute`) so
  the user knows whether to set `$HOME`, fix encoding, or report a buggy
  lookup. No silent fallback to root or to an empty path.

  Type choices that came out of review along the way:
  - `resolve_home` returns `HostAbsPath`, surfacing relative results as
    `NotAbsolute` instead of letting them flow through.
  - `expand_home` and `expand_policy_home` take `&HostAbsPath`
    unconditionally; the `Option` lives at the caller, so the
    "tilde + no home" combination is impossible at the type level rather
    than enforced by `expect`.
  - `compute_dest`'s strip_prefix is camino's path-component-aware
    variant, so `/etc/xdg` no longer falsely matches `/etc/xdgfoo`.

  Tests cover all three home-failure modes, the lazy-lookup precondition,
  hook-introduced tilde rules, and the path-component-aware strip-prefix
  regression. 127 lib tests pass.

Resolves https://github.com/gominimal/inbox/issues/196

## Summary by CodeRabbit

* **New Features**
  * Patch source and policy patterns with leading `~` are expanded against the host home during session resolution; raw `~` is preserved in saved/returned policies.
  * Resolver skips host-home lookup when no `~` patterns exist.
  * Policies returned/modified by hooks that introduce `~` rules are re-expanded and enforced.

* **Documentation**
  * Clarified `~` semantics for sources vs destinations, the resolution stages, and explicit failure cases for unresolved home paths.

### DIFF
--- a/crates/sessions/src/composable.rs
+++ b/crates/sessions/src/composable.rs
@@ -548,11 +548,16 @@ pub enum ResolveError {
     /// At least one patch source or policy pattern starts with `~`,
     /// but resolving the host home directory failed.
     ///
-    /// Both failure modes (no home available; non-UTF-8 home path)
-    /// live under one variant because the home lookup happens once at
-    /// the start of resolution — not per pattern — so the error has
-    /// no pattern context to surface. Match on the inner
-    /// [`HomeResolutionFailure`] to distinguish causes.
+    /// All three failure modes (no home available, non-UTF-8 home
+    /// path, non-absolute home path) live under one [`HomeUnresolved`]
+    /// variant because the home lookup happens once at the start of
+    /// resolution — not per pattern — so the error has no pattern
+    /// context to surface. Match on the inner [`HomeResolutionFailure`]
+    /// to distinguish [`Unavailable`][HomeResolutionFailure::Unavailable],
+    /// [`NotUtf8`][HomeResolutionFailure::NotUtf8], and
+    /// [`NotAbsolute`][HomeResolutionFailure::NotAbsolute].
+    ///
+    /// [`HomeUnresolved`]: Self::HomeUnresolved
     #[error("a `~`-prefixed pattern requires home expansion, but {0}")]
     HomeUnresolved(#[from] HomeResolutionFailure),
 }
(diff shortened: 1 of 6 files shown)


twitchyliquid64 (approved):

crates/sessions/src/composable.rs
```diff
@@ -969,14 +1079,161 @@ fn enumerate_patch_files(
...
+/// preserved as a literal `~` character.
+///
+/// Returns the original `FileSet` borrowed (zero alloc) when the
+/// pattern doesn't begin with `~`. Returns a fresh `FileSet` (one
+/// re-parse) when expansion happens.
+///
+/// The home argument is unconditional: callers are expected to gate
+/// the call on whether any tilde pattern is in scope, so by the time
+/// this function runs we always know a usable home is available.
+fn expand_home<'a>(
```
norrietaylor, line 1097: `format!("{home}/{suffix}")` feeds the home into a glob unescaped — a home containing `*?[]{}\ (e.g. /home/u[1])` reinterprets the home segment, so walk_root() truncates at the metachar and the patch silently matches nothing.

norrietaylor (approved):

crates/sessions/src/composable.rs
evanspearman: Thanks. Good catch!

This is actually being completely replaced in the var expansion stuff I'm currently working on and the bug should no longer be present. I'll add an explicit test for this case in that change though.

PR #375, opened 2026-06-10
feat(minimald): scaffolding for the attachable session process

twitchyliquid64:
The beginnings of attachable sessions.

Remaining TODOs:
 - [x] Fix the lifecycle of the sandboxed environment, which is being torn down even with the process still in it
 - [x] Fix job control + Ctrl-C not working
   - needs https://github.com/souk4711/hakoniwa/pull/174
 - [x] Add a million tests for all the state/race/sequence invariants
 - [x] Implement detaching via a key-chord

Given an existing session with a `minimal.toml` in it, you should be able to spawn or reattach to the session by doing:

```shell
$> MINIMAL_SESSION_ID=<session-uuid> ssh -o SendEnv=MINIMAL_SESSION_ID \
    -o ProxyCommand='socat - UNIX-CONNECT:/home/xxx/.local/state/minimal/providers/local-0/ssh.sock' \
    -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' local
```

AKA a 'request shell' ssh session - one with no command trailing.

## Summary by CodeRabbit

* **New Features**
  * Added interactive SSH shell support with PTY-backed terminal I/O, window-size handling, attach/detach and resume behavior preserving terminal state.
  * Exposed session host controls enabling programmatic attach and kill of session hosts.
  * Added API to set sandboxed container as session leader.

* **Refactor**
  * Reworked startup to build a custom Tokio runtime and moved async startup into a separate entry function.

* **Chores**
  * Updated workspace dependencies.

### DIFF
--- a/crates/minimald/src/session_host.rs
+++ b/crates/minimald/src/session_host.rs
@@ -331,6 +331,53 @@ pub struct Host {
 
     // Temporary buffer for reading from the pty master (i.e. 'stdout').
     stdout_buf: Vec<u8>,
+
+    // Keeps the session's `Env` — and the `Context` and `Graph` it borrows —
+    // alive for as long as this host (and thus the session process) lives.
+    // Declared last so it is dropped after `process`: the process is torn down
+    // before the sandbox files backing its rootfs are removed. Never read; held
+    // purely for its `Drop`.
+    _session_env: SessionEnv,
+}
+
+/// Owns the [`mctx::Env`] for a session, along with the [`mctx::Context`] and
+/// [`graph::Graph`] it borrows, keeping all three alive for the session.
+///
+/// `Env<'a>` borrows the context and graph mutably and owns the temp dirs and
+/// sandbox base directory that hold the running process's rootfs. We can't
+/// store the env next to the values it borrows (that would be a self-referential
+/// struct), so the context and graph are leaked to obtain `'static` borrows for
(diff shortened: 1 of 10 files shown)


crates/minimald/src/session_host.rs
```diff
@@ -0,0 +1,700 @@
...
+                    },
+                    Err(_would_block) => {},
+                }
+            },
+            // Read from remote (ssh channel) - these bytes need writing to session process stdin
+            Some(msg) = self.remote_rx.recv() => {
+                match msg {
+                    Either::Left(b) => {
+                        let mut b = &b[..];
+                        while !b.is_empty() {
```
norrietaylor: Since this loop holds the select! arm until all of b is written, the read arm can't run — could that deadlock against a child that's blocked writing output? A pending-stdin buffer with its own arm might be safer.

twitchyliquid64: Yeah, its possible for a program to block on writing to stdout and not consume stdin until unblocked, which would lead to a deadlock. So some amount of buffering here would be reasonable.

twitchyliquid64: Should be fixed by: https://github.com/gominimal/minimal/pull/375/changes/34ffff5e691a536f4c12f09add8ab9550e08a872

norrietaylor (approved):

PR #389, opened 2026-06-12
feat(sessions): var expansion and path canonicalization

### LINKED ISSUE #384: Expand `$VAR` in patch source patterns
## Context

`~` expansion in patch sources and policy lives in the resolver
(#07), but `$VAR` expansion in patch *source* patterns is still
deferred. The patches module doc has been carrying a TODO since the
original commit, and a user writing
`source = "$XDG_CONFIG_HOME/nvim/**"` today gets a silent empty
match — the OS doesn't interpret `$VAR`, and we never expand it.

Scope-narrowing decisions from design discussion:

- **Source paths only.** `$VAR` does not affect policy patterns
  (`allow` / `deny` / `ignore` stay literal) and does not affect
  `PatchDest` (dests are always implicitly sandbox-home-relative;
  the apply layer doesn't expand vars there). Policy patterns
  staying literal keeps the security story simple: an unset env var
  can never silently disable a deny rule.
- **Host env at resolution time.** Same shape as the home-expansion
  flow — expand before the walker runs, error if a required var is
  unset.

### Where `$VAR` expansion happens

| Context | Expanded? |
|---|---|
| Patch `source` `FileSet` | yes |
| `PatchPolicy` `allow/deny/ignore` | **no** — patterns stay literal |
| Patch `dest` (`PatchDest`) | **no** — already sandbox-home-relative |
| Var values (`VarValue::Specified`) | already handled by `ResolvedVar` |

### Syntax

- `$NAME` and `${NAME}` both recognized.
- `\$` escapes a literal `$`.
- `$` not followed by a valid name char is a literal `$` (no
  error).
- Names follow `[A-Z_][A-Z0-9_]*` (matches `StrictVarName`). Names
  that look like lowercase / lenient identifiers are left literal.

### Behavior on lookup failure

- **Var unset in host env**: error.
  `ResolveError::VarUnset { name, in_pattern }` with the pattern
  text quoted so the loadout author can find the call site.
- No silent empty-string substitution. `$XDG_CONFIG_HOME/nvim`
  silently becoming `/nvim` is a real footgun; better to fail loud.

### Resolution layer

- Mirrors the home-expansion shape (`expand_home`):
  - `expand_vars(fs, env_lookup) -> Result<Cow<FileSet>, ResolveError>`
- Lazy: only invoked when a pattern contains a `$` that introduces
  a real name.
- `env_lookup: &dyn Fn(&str) -> Option<String>` — caller-supplied,
  same pattern as `HomeLookup`. Production wires it to
  `std::env::var`; tests pass fakes.
- Called from `enumerate_patch_files` alongside `expand_home`. Both
  fire per FileSet; `~` is expanded first, then `$VAR` (order
  doesn't matter today, but pinning the order keeps future edits
  predictable).
- Caches per-name lookups inside a single resolution so the same
  `$XDG_CONFIG_HOME` reference across N patches reads env once.

### Round-trip

Returned policy retains source patterns in `$VAR` form? **No** —
sources don't get round-tripped through resolution; they live on
the `Patch` objects in the original loadout. The expanded form
exists only inside the resolver and the resulting `SessionPatch`'s
`host_path` (which is the fully-resolved file). So this falls out
for free.

## Acceptance criteria

- [ ] `source = "$XDG_CONFIG_HOME/nvim/**"` walks the right tree
  (test pins `$XDG_CONFIG_HOME` via the env lookup fake).
- [ ] `source = "$NOT_SET/foo"` errors with
  `ResolveError::VarUnset { name: "NOT_SET", … }`.
- [ ] Policy patterns containing `$VAR` are passed to the matcher
  literally (the policy is checking against the
  post-walker, post-expansion *source paths*, so `$` in a deny rule
  matches a `$` in a real file name — vanishingly rare in
  practice).
- [ ] `dest = "$HOME/foo"` is rejected at `PatchDest::try_new`
  (already true — `$` isn't a valid leading char and absolute
  paths are already rejected; this is a regression-prevention
  test).
- [ ] `\$` escape produces a literal `$`.
- [ ] Lowercase / malformed names (`$foo`, `$1bar`) are left
  literal, no error.

## Open questions

- Should we ever expand `$VAR` in policy patterns? The argument for:
  user can write `deny = ["$HOME/.ssh/**"]` in `~/`-rejected form.
  The argument against: `~` already does this cleanly, and `$VAR`
  in policy opens the door to "unset var silently disables a deny
  rule" footguns. Decision: stay literal. If a user wants the
  effect, `~/.ssh/**` works.
- Vars-in-paths gating (separate policy listing which vars may be
  substituted) — punted from this issue. Easy to add later as a
  layered concern if a real use case appears; today the user is
  the only one who writes patch source patterns (project/package
  contributors not yet wired in via gominimal/inbox#8), and their loadout is
  trusted.

## Notes

The patches module's old "subject to a separate policy" doc claim
about `$VAR` is intentionally not implemented here. If
project-/package-origin patches start writing `$VAR`-prefixed
sources after gominimal/inbox#8 lands, revisit whether a vars-in-paths gate is
needed for non-user contributions.

## Blocks

- None hard, but cleans up a `// TODO`-shaped gap and unblocks
  `$XDG_*` patterns in real loadouts, which is the most common
  reason users will reach for env-based sources.

evanspearman:
## Summary

  - Unifies tilde + `$VAR` expansion behind one `expand_source` call,
    driven by resolved session vars
  - Adds path canonicalization + dual-path policy check for symlinks
  - Rips out the old `HomeLookup` / `HomeResolutionFailure` /
    `expand_home` machinery

  ## Behavior changes

  - Patterns must be absolute (`~/x` or `$HOME/x` for home-relative).
    Relative → `ExpandError::NotAbsolute`.
  - `..` rejected everywhere (raw or substituted).
  - Substituted glob metas now escaped (fixes the parent PR's
    unescaped-home regression for `~/`).
  - `follow_symlinks: true` checks both link + canonical target; deny
    on either wins.
  - `PatchPolicy` setters are infallible; validation moves to
    `expand_with` at resolve time.

  ## Test plan

  - [x] `cargo test -p sessions` (174 lib + 5 doc)
  - [x] `cargo clippy -p sessions --all-targets -- -D warnings`

Resolves https://github.com/gominimal/minimal/issues/384

## Summary by CodeRabbit

* **New Features**
  * Patch source patterns now expand `~` and `$VAR`/`${VAR}` using session-resolved values (with clearer errors when expansion fails).
  * Patch policy rules now evaluate against pre-expanded patterns with improved `allow`/`deny`/`ignore` precedence and more accurate matching when symlinks are involved.

* **Bug Fixes/Improvements**
  * Safer path handling: rejects non-absolute patterns and prevents traversal via `..` from expanded inputs.
  * Patch destinations are normalized at creation (removes redundant slashes and `.` components).

* **Tests / CI**
  * Expanded coverage for expansion failures, symlink dual-path behavior, and policy re-expansion; macOS CI now runs the sessions test suite.

### DIFF
--- a/crates/sessions/src/composable.rs
+++ b/crates/sessions/src/composable.rs
@@ -307,24 +307,6 @@ impl Contribution {
 /// [`std::env::var`]; tests pass a synthetic closure.
 pub type EnvLookup<'a> = &'a dyn Fn(&str) -> Result<String, std::env::VarError>;
 
-/// A lookup function for resolving the host user's home directory.
-///
-/// Threaded through [`Composer::resolve`] so the patch resolver can
-/// expand leading `~` in source patterns. Returns `None` when no home
-/// is available (e.g. `$HOME` unset and no `passwd` entry); the
-/// resolver surfaces that as [`ResolveError::HomeUnresolved`] when a
-/// pattern actually requires it.
-///
-/// Invoked at most twice per resolution: once at the top of
-/// [`Composer::resolve`] if any `~`-prefixed pattern exists, and once
-/// more if a hook returns an `updated_policy` introducing new
-/// `~`-prefixed rules. A `~`-free policy in a `HOME`-less environment
-/// never triggers the lookup.
-///
-/// Production callers should pass [`dirs::home_dir`]; tests can pin a
-/// synthetic closure via [`Composer::with_home`].
-pub type HomeLookup<'a> = &'a dyn Fn() -> Option<std::path::PathBuf>;
-
 /// Anything that can contribute primitives (vars, patches, packages,
 /// lifecycle hooks) to a [`Composer`] during session construction.
 ///
@@ -530,57 +512,19 @@ pub enum ResolveError {
     /// batch so the message points somewhere concrete.
     #[error("policy hook contract violation: {kind} ({context})")]
     HookContract { kind: &'static str, context: String },
-    /// A patch source has an unwalkable configuration — typically a
-    /// pattern with no literal path prefix. The user must fix their
-    /// loadout/project config; retrying without changes will not help.
-    ///
-    /// Surfaced separately from [`PatchWalk`](Self::PatchWalk) because
-    /// configuration errors and transient IO failures have different
-    /// audiences (the loadout author vs. the operator).
-    #[error("patch source configuration is invalid ({} error(s)):{}", sources.len(), DisplayJoin(sources))]
-    PatchConfig { sources: Vec<crate::patches::Error> },
     /// One or more patch source filesystem walks failed with IO-level
     /// errors (permission denied, non-UTF-8 paths, etc.). All errors
     /// surfaced by every `FileSet::resolve` invocation are accumulated
     /// — none are discarded.
     #[error("patch resolution failed ({} error(s)):{}", sources.len(), DisplayJoin(sources))]
     PatchWalk { sources: Vec<crate::patches::Error> },
-    /// At least one patch source or policy pattern starts with `~`,
-    /// but resolving the host home directory failed.
-    ///
-    /// All three failure modes (no home available, non-UTF-8 home
-    /// path, non-absolute home path) live under one [`HomeUnresolved`]
-    /// variant because the home lookup happens once at the start of
-    /// resolution — not per pattern — so the error has no pattern
-    /// context to surface. Match on the inner [`HomeResolutionFailure`]
-    /// to distinguish [`Unavailable`][HomeResolutionFailure::Unavailable],
-    /// [`NotUtf8`][HomeResolutionFailure::NotUtf8], and
-    /// [`NotAbsolute`][HomeResolutionFailure::NotAbsolute].
-    ///
-    /// [`HomeUnresolved`]: Self::HomeUnresolved
-    #[error("a `~`-prefixed pattern requires home expansion, but {0}")]
-    HomeUnresolved(#[from] HomeResolutionFailure),
-}
-
-/// Why home-directory resolution failed during patch resolution.
-///
-/// All variants describe operator/environment problems, not
-/// loadout-config problems: the loadout author can't fix any of them.
-#[non_exhaustive]
-#[derive(Debug, thiserror::Error)]
-pub enum HomeResolutionFailure {
-    /// No home directory was found (e.g. `$HOME` unset and no
-    /// `passwd` entry).
-    #[error("no home directory was found")]
-    Unavailable,
-    /// The home directory's path is not valid UTF-8.
-    #[error("home directory is not valid UTF-8: `{lossy}`")]
-    NotUtf8 { lossy: String },
-    /// The home directory's path is not absolute. Home is expected to
-    /// be absolute by POSIX convention; a relative result indicates a
-    /// misbehaving `HomeLookup`.
-    #[error("home directory is not absolute: `{0}`")]
-    NotAbsolute(Utf8PathBuf),
+    /// Expanding `~/` or `$VAR` references in a patch source or policy
+    /// pattern failed. Surfaces every failure mode of
+    /// [`expand_source`](crate::expansion::expand_source): malformed
+    /// syntax, a referenced var that is not in the resolved-vars set,
+    /// or a post-expansion string that fails to parse as a glob.
+    #[error("patch source expansion failed: {0}")]
+    Expansion(#[from] crate::expansion::ExpandError),
 }
 
 /// Render a slice of `Display`-able errors as one indented bullet per
@@ -609,6 +553,14 @@ pub struct SessionVar {
 }
 
 impl SessionVar {
+    /// Construct directly. Outside the resolver, the only callers
+    /// expected to use this are tests building synthetic sessions.
+    #[cfg(test)]
+    #[must_use]
+    pub fn new(var: ResolvedVar, source: Source) -> Self {
+        Self { var, source }
+    }
+
     /// The resolved variable that survived policy resolution.
     #[must_use]
     pub fn var(&self) -> &ResolvedVar {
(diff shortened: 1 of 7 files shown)


norrietaylor (changes requested):
This patch is breaking local macos builds, this is my fault as the `ci-macos.yml` path filter I added omits `crates/sessions/**`, so the workflow never runs for this PR. Even when it runs, the macOS jobs only test `minvmd` (`cargo test -p minvmd`, `clippy -p minvmd` at `ci-macos.yml:360-367`). Adding sessions to the paths would trigger the workflow but still never execute a sessions test.

Two edits to `.github/workflows/ci-macos.yml` are needed

 1. Add sessions to both path filters (push and pull_request):

```
      paths:
        - "crates/minvmd/**"
        - "crates/minimal2/**"
        - "crates/sessions/**"     # platform-sensitive: canonicalization, symlinks
        ...

```
  Run the sessions tests in the existing clippy+test job (the one at line ~336 on the self-hosted runner) — one line, reuses the job's build cache rather than adding a new job to the bottlenecked single runner:

```
        - run: cargo clippy -p minvmd --all-targets -- -D warnings
        - run: cargo test -p minvmd
        - run: cargo test -p sessions
```

crates/sessions/src/expansion.rs
```diff
@@ -0,0 +1,650 @@
...
+}
+
+/// Drop `.` and empty components, reject any `..` component.
+///
+/// Preserves a leading `/` (absolute paths) and a trailing `/` (the
+/// "this names a directory" affordance some patterns rely on).
+/// Glob metacharacters (`**`, `*`, `?`, `[...]`, `{...}`) are
+/// path components in their own right and pass through unchanged —
+/// only literal `.` and `..` components are touched.
+fn normalize_path(s: &str) -> Result<String, ExpandError> {
```
twitchyliquid64, line 182: Path traversal is blanket illegal?

Seems like a safe starting point, but i wonder if we want to loosen this in the future.  Handling path traversal is a mess tho (absolutize() isnt good enough if the user can do symlinks - canonicalize is best but requires files to exist + eats IOPS).

crates/sessions/src/expansion.rs
```diff
@@ -0,0 +1,650 @@
...
+
+fn lookup<'a>(vars: &'a [SessionVar], name: &str) -> Option<&'a str> {
+    vars.iter()
+        .find(|v| v.var().name() == name)
+        .map(|v| v.var().value())
+}
+
+/// Parse a `$VAR` or `${VAR}` reference starting at `bytes[at]` (which
+/// is the `$`). Returns `(name, bytes_consumed_including_dollar)`.
+fn parse_var_ref(bytes: &[u8], at: usize) -> Result<(&str, usize), ExpandError> {
```
twitchyliquid64, line 220: Might be worth adding a comment here that escaping variable declarations (i.e. `$$var`) must be handled earlier

twitchyliquid64 (approved):
Approved tho note Norries comment above about fixing CI

crates/sessions/src/expansion.rs
evanspearman: I'm just having a hard time seeing a good reason to allow `..`, but I may be missing something and there may be workflows that would benefit from it. I don't think it would be that hard to support in the future if we need to though.

crates/sessions/src/expansion.rs
evanspearman: Good catch!

crates/sessions/src/expansion.rs
evanspearman: A note on symlinks: We do need to support symlinks because home-manager (nix) uses them extensively and I want to support home-manager generated files for admittedly selfish reasons. That said to mitigate security concerns around them:
- Policies apply to both the link and the file it's linking to. If one is ignored or denied it treats them as both being such and if one isn't allowed (outside of user loadouts which bypass the user allowlist for ergonomic reasons) neither are allowed
- There will be a user config option to turn off resolving symlinks entirely

twitchyliquid64: If we dont support `..`, i dont think its possible to use symlinks to admit sensitive files in a way that wasnt possible already without this primitive. Because if someone can create symlinks on the host, they can already access files on the host.

crates/sessions/src/composable.rs
```diff
@@ -870,8 +974,12 @@ impl Composer {
     ) -> Result<(Resolution, UserPolicy), ResolveError> {
         let (vars_policy, patches_policy) = policy.into_parts();
         let (vars, vars_policy) = resolve_vars(self.vars, vars_policy, hooks)?;
+        // Patch sources and policy patterns expand against the
+        // resolved vars produced by `resolve_vars`. Anything they
+        // reference must already be in `vars` — there is no fallback
+        // to process env.
         let (patches, patches_policy) =
-            resolve_patches(self.patches, patches_policy, hooks, options, &*self.home)?;
+            resolve_patches(self.patches, patches_policy, hooks, options, &vars)?;
```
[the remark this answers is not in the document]
evanspearman: Fixed

crates/sessions/src/composable.rs
```diff
@@ -1284,52 +1384,53 @@ fn resolve_patches(
...
-    //     patterns retain their `~` form so the policy round-trips.
-    //   - `expanded`: home-expanded copy used for the actual `check`
+    //     patterns retain their `~/` and `$VAR` form so the policy
+    //     round-trips through serialization.
+    //   - `expanded`: var-expanded copy used for the actual `check`
     //     calls — patterns actually match the absolute paths the
     //     walker yields. Re-derived whenever the hook updates the
     //     policy.
-    let mut expanded = expand_policy_home_opt(&policy, home.as_ref())?;
+    let mut expanded = policy.expand_with(resolved_vars)?;
```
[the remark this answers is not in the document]
evanspearman: Fixed

crates/sessions/src/expansion.rs
```diff
@@ -0,0 +1,650 @@
...
+            if bytes.get(i + 1) == Some(&b'$') {
+                out.push('$');
+                i += 2;
+                continue;
+            }
+            let (name, consumed) = parse_var_ref(bytes, i)?;
+            let value = lookup(resolved_vars, name).ok_or_else(|| ExpandError::UndefinedVar {
+                name: name.to_owned(),
+            })?;
+            escape_glob_metas(value, &mut out);
```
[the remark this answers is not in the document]
evanspearman: Fixed

crates/sessions/src/expansion.rs
```diff
@@ -0,0 +1,650 @@
...
+            i += consumed;
+            continue;
```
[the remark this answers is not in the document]
evanspearman: This was an intentional design choice

crates/sessions/src/expansion.rs
evanspearman: Fixed

evanspearman:
> This patch is breaking local macos builds, this is my fault as the `ci-macos.yml` path filter I added omits `crates/sessions/**`, so the workflow never runs for this PR. Even when it runs, the macOS jobs only test `minvmd` (`cargo test -p minvmd`, `clippy -p minvmd` at `ci-macos.yml:360-367`). Adding sessions to the paths would trigger the workflow but still never execute a sessions test.
> 
> Two edits to `.github/workflows/ci-macos.yml` are needed
> 
>     1. Add sessions to both path filters (push and pull_request):
> 
> 
> ```
>       paths:
>         - "crates/minvmd/**"
>         - "crates/minimal2/**"
>         - "crates/sessions/**"     # platform-sensitive: canonicalization, symlinks
>         ...
> ```
> 
> Run the sessions tests in the existing clippy+test job (the one at line ~336 on the self-hosted runner) — one line, reuses the job's build cache rather than adding a new job to the bottlenecked single runner:
> 
> ```
>         - run: cargo clippy -p minvmd --all-targets -- -D warnings
>         - run: cargo test -p minvmd
>         - run: cargo test -p sessions
> ```

Fixed

norrietaylor (commented):
LGTM

norrietaylor (approved):

PR #423, opened 2026-06-15
feat(minimald): implement populating session files using a tarball

twitchyliquid64:
Lets you populate workspace files by streaming a tarball. You can drive it manually like this:

```shell

$> tar -v -cf - -C 'my/workspace' ./ | zstd -c -T0 | \
    MINIMAL_SESSION_ID=aaaa019eb7a0-36d9-7d41-8286-265d76e9b3ab ssh -s -o SendEnv=MINIMAL_SESSION_ID \
    -o ProxyCommand="socat - UNIX-CONNECT:$XDG_STATE_DIR/minimal/providers/local-0/ssh.sock" \
    -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' local-0 \
    minimald-v1-WorkspaceFilesTarZst
```

## Release Notes

* **New Features**
  * Added a new SSH subsystem that streams a compressed tar archive and unpacks it into a target session’s workspace directory.

* **Chores**
  * Updated workspace dependencies to enable async streaming and compressed archive handling.

* **Tests**
  * Added coverage for successful streamed unpacking and for handling an unknown session identifier.

### DIFF
--- a/crates/minimald/src/rpc.rs
+++ b/crates/minimald/src/rpc.rs
@@ -1,18 +1,21 @@
 use minimald_rpc::{
     CreateSession, CreateSessionResponse, Errorable, GetSessionRecord, GetSessionRecordRequest,
     GetSessionRecordResponse, GetVersion, GetVersionResponse, ListSessions, ListSessionsEntry,
-    ListSessionsResponse, OneshotSshRpc,
+    ListSessionsResponse, OneshotSshRpc, RPC_SUBSYSTEM_PREFIX,
 };
 use russh::{
     Channel as RuChannel, ChannelId,
     server::{Msg, Session},
 };
+use sessions::SessionId;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::task::spawn;
 
 use crate::{
+    ChannelConfig,
     connection::{ConnectionError, ConnectionHandle},
     server::ServerStateHandle,
+    sessions::SessionKeyPredicate,
 };
 
 /// Server-side serving glue for [`OneshotSshRpc`]s.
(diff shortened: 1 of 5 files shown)


crates/minimald/src/rpc.rs
```diff
@@ -148,6 +151,66 @@ async fn serve_create_session(s: ServerStateHandle, c: RuChannel<Msg>) {
...
+    let reader = async_compression::tokio::bufread::ZstdDecoder::new(tokio::io::BufReader::new(
+        c.make_reader(),
+    ));
+    if let Err(e) = async_tar::Archive::new(reader)
+        .unpack(workspace_path.as_utf8_path())
+        .await
+    {
+        let _ = c
+            .extended_data_bytes(1, format!("unpack failed: {e}"))
+            .await;
```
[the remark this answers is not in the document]
twitchyliquid64: Nah, would rather do it live / non-atomically

norrietaylor (approved):
Pretty slick!

crates/minimald/src/rpc.rs
```diff
@@ -148,6 +151,66 @@ async fn serve_create_session(s: ServerStateHandle, c: RuChannel<Msg>) {
...
+            let _ = c
+                .extended_data_bytes(1, format!("session UUID lookup failed: {e}"))
+                .await;
+            let _ = c.close().await;
+            return;
+        }
+    };
+    let workspace_path = session_handle.workspace_path().await;
+
+    let reader = async_compression::tokio::bufread::ZstdDecoder::new(tokio::io::BufReader::new(
```
evanspearman, line 193: Probably out of scope, just something we should think about: this compression has a cost. Given that we're most likely sending data locally rather than over the network, it may actually be faster to *not* compress it. Most file sync will happen at session start, which we want to feel snappy, so it may be worth doing a comparison of a few different workloads to see which is faster.

crates/minimald/src/rpc.rs
```diff
@@ -203,12 +271,131 @@ pub async fn handle_ssh_rpc(
...
+            let mut header = async_tar::Header::new_gnu();
+            header.set_size(contents.len() as u64);
+            header.set_mode(0o644);
+            tar.append_data(&mut header, path, *contents).await.unwrap();
+        }
+        let tar_bytes = tar.into_inner().await.unwrap();
+
+        let mut encoder = async_compression::tokio::write::ZstdEncoder::new(Vec::new());
+        encoder.write_all(&tar_bytes).await.unwrap();
+        encoder.shutdown().await.unwrap();
```
evanspearman, line 289: Do any of these have failure modes we need to care about? It would be ideal to document why it's okay to unwrap here with an expect.

crates/minimald/src/rpc.rs
```diff
@@ -203,12 +271,131 @@ pub async fn handle_ssh_rpc(
...
+
+    /// Creates a session through the public RPC and returns its id.
+    async fn fresh_session(client: &mut TestClient) -> SessionId {
+        client
+            .call::<CreateSession>(&CreateSessionRequest {
+                record: sessions::Record {
+                    id: SessionId::nil(),
+                    name: Some("stream-test".to_string()),
+                    username: None,
+                    project_path: HostAbsPath::try_new("/tmp").unwrap(),
```
evanspearman, line 301: Just FYI we have unchecked versions of the constructors now for scenarios like this.

crates/minimald/src/test_harness.rs
```diff
@@ -179,6 +179,22 @@ impl TestClient {
...
+    pub(crate) async fn open_subsystem(
+        &mut self,
+        subsystem: &str,
+        env: &[(&str, &str)],
+    ) -> russh::Channel<russh::client::Msg> {
+        let channel = self.handle.channel_open_session().await.unwrap();
+        for (name, value) in env {
+            channel.set_env(true, *name, *value).await.unwrap();
+        }
+        channel.request_subsystem(true, subsystem).await.unwrap();
```
evanspearman, line 194: Similarly, what are the failure modes here?

evanspearman (changes requested):

crates/minimald/src/rpc.rs
twitchyliquid64: This is the ideal case for zstd compression with level 0 or less, its usually faster than I/O itself (colossus for example compresses all stripes with snappy, which is googles zstd equivalent). On the client I was thinking we would keep the compression level really low for this case.

crates/minimald/src/rpc.rs
twitchyliquid64: I imagine the failure modes are:
1. Broken connection
2. Bad framing: either zstd or tarball. In any case we should probably not panic and report an error before closing the connection.

Will refactor to handle these.

crates/minimald/src/rpc.rs
twitchyliquid64: 👀

crates/minimald/src/test_harness.rs
twitchyliquid64: These will fail if the server misbehaves, but this is all test wiring so im okay with unwrap() everywhere

crates/minimald/src/rpc.rs
twitchyliquid64: Oh, actually this is all test code. Im okay with unwrapping here for brevity.

crates/minimald/src/rpc.rs
evanspearman: Sounds good

crates/minimald/src/rpc.rs
evanspearman: Oops, sorry about that

evanspearman (approved):

PR #425, opened 2026-06-16
feat: session-specific sandbox wiring

twitchyliquid64:
* Sandbox env:
   * Session-specific home at `/home`
   * CWD / 'repo-dir' at `/workbench`, configurable
   * XDG_* variables point into `/home` except cache, which points into `/state` (we should probably rename that /cache at this point lol)

 * Duplicated `mctx::Env` into minimald and started specializing it for sessions, as well as removed some of the lifetime crimes we were doing with `&mut Context` / `&mut Graph`.

NB: The duplication is intentional: we want to support the old `minimal run <task>` semantics as long as possible, so duplicating will let us cook on the sessions stuff separately.

## Summary by CodeRabbit

* **New Features**
  * Added a new “Session” sandbox layout with configurable home and working directories, including optional working-directory naming.
  * Introduced a session-owned runtime environment and a new `min` helper to drive session actions (add/search/run/check/build and related operations).

* **Improvements**
  * Updated session path handling so exec, SFTP, and workspace file unpacking consistently use the session working directory.
  * Session attachment now propagates the authenticated SSH username for the launched session.

* **Bug Fixes**
  * Fixed command execution, repository receive operations, and unpacked file locations to match the new session working-directory behavior.


crates/sandbox2/src/config.rs
```diff
@@ -66,6 +66,14 @@ pub enum WdSetup {
         read_only: bool,
         fs_mappings: Vec<common::FsMapping>,
     },
+    /// The layout used for a minimal session.
+    ///
+    /// The homedir is at /home, and the working directory is at /workbench (unless overridden).
+    Session {
+        home: PathBuf,
```
evanspearman, line 73: Any reason not to use `camino::Utf8PathBuf` or one of our newtypes here?

twitchyliquid64: Sandbox2 as a crate needs to be lifted to use the path newtypes as a whole, but Ive adjusted all the usage in minimald to use the specific `paths::` type.

crates/minimald/src/rpc.rs
```diff
@@ -188,13 +188,13 @@ async fn unpack_workspace_files(
...
+    let paths = session_handle.paths().await;
 
     let reader = async_compression::tokio::bufread::ZstdDecoder::new(tokio::io::BufReader::new(
         c.make_reader(),
     ));
     async_tar::Archive::new(reader)
-        .unpack(workspace_path.as_utf8_path())
+        .unpack(paths.working.as_utf8_path())
         .await
         .map_err(|e| format!("unpack failed: {e}"))?;
```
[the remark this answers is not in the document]
twitchyliquid64: Version `0.6.1` (which we use) has a fix for https://github.com/advisories/GHSA-j5gw-2vrg-8fgx.

crates/minimald/src/env.rs
```diff
@@ -0,0 +1,964 @@
...
+        } else {
+            None
+        };
+
+        let result: Result<(), Error> = async {
+            let mut env = build_ctx
+                .make_env(
+                    task_name,
+                    &mut graph,
+                    task.inherit_cwd.then(|| std::env::current_dir().unwrap()),
```
norrietaylor: `inherit_cwd` resolves to the daemon's cwd, not the session's.

`std::env::current_dir()` here runs inside the daemon actor, so it returns the daemon process working directory rather than the user's session working dir (`/workbench`). This was copied from `mctx/src/env.rs:332`, where the call runs in the CLI process and is correct; in the daemon it is wrong, and `.unwrap()` panics if the daemon cwd is unreadable.

Thread the session working dir through instead of reading the process cwd.


(claude assist on this one)

crates/minimald/src/env.rs
```diff
@@ -0,0 +1,964 @@
...
+/// Two writers can share one connection (via `Arc<Mutex<_>>`) so stdout and
+/// stderr don't interleave mid-line.
+struct StreamWriter {
+    stream: Arc<Mutex<UnixStream>>,
+    buf: Vec<u8>,
+}
+
+impl StreamWriter {
+    fn pair(stream: &UnixStream) -> (Self, Self) {
+        let shared = Arc::new(Mutex::new(stream.try_clone().unwrap()));
```
norrietaylor: `try_clone().unwrap()` panics the session command actor.

Reached by `min run` / `min patched-pkg`. On fd exhaustion the spawned actor task (`tokio::spawn(channel.run())`) panics and dies, killing the `min` channel for the whole session. `BridgeChannel::handle` (env.rs:358) already handles the identical `try_clone` failure gracefully — return an error here too instead of unwrapping.

(claude assist on this one)

crates/minimald/src/env.rs
```diff
@@ -0,0 +1,964 @@
...
+        for bsr in transitives.keys() {
+            if self.has_packages.insert(*bsr) {
+                let src = match cache.read_dir(&new_graph.spec_hash(bsr)) {
+                    Ok(entry) => entry,
+                    Err(e) => {
+                        let _ = writeln!(stream, "error: {e}");
+                        return;
+                    }
+                };
+                if let Err(e) = common::hardlink_dir_contents(
```
norrietaylor: Blocking `hardlink_dir_contents` runs on the async runtime.

This does recursive `read_dir`/`metadata`/`hard_link` synchronously inside the async `install` handler (same for `create_dir_all`/`install_min_helpers` in `Env::build`). `docs/rust-coding-standards.md`: "No blocking in async context. No std::fs ... use tokio::fs ... spawn_blocking." Hardlinking a large package tree stalls a runtime worker thread. Wrap the heavy I/O in `tokio::task::spawn_blocking`.

(claude assist on this one)

crates/minimald/src/env.rs
```diff
@@ -0,0 +1,964 @@
...
+use ot::OpTracker;
+use paths::DaemonAbsPath;
+use sandbox2::config::{Config, SandboxMapped};
+use sandbox2::{Container, Sandbox};
+use tempfile::TempDir;
+use tokio::sync::mpsc;
+use tokio::task::JoinHandle;
+
+/// The min helper script installed at `/usr/bin/min` inside the sandbox.
+const MIN_SCRIPT: &str = include_str!("env_min_helper.sh");
```
norrietaylor, line 51: `env.rs` and `env_min_helper.sh` duplicate `mctx::Env` wholesale.

`env_min_helper.sh` is byte-identical to `crates/mctx/src/min_helper.sh`. `MIN_SCRIPT`, `StreamWriter` + its `AsyncWrite` impl, and the `run_check`/`run_patched_pkg`/`run_task`/dispatch handlers mirror `mctx/src/env.rs` — the only real difference is sync-nested-runtime vs. async actor. The `min` protocol and handler logic now have to be maintained in two places and will drift.

The doc comment justifies owning vs. borrowing; that argues for making `mctx::Env` generic over ownership and sharing the helper/protocol layer, not forking ~964 lines.

crates/minimald/src/env_min_helper.sh
```diff
@@ -0,0 +1,156 @@
...
+    fi
+
+    if [[ "$error" == "true" ]]; then
+        return 1
+    fi
+}
+
+min_search() {
+    local term="$1"
+    if [[ -z "$term" || -z "$term" ]]; then
```
norrietaylor: Redundant duplicate condition: `[[ -z "$term" || -z "$term" ]]` checks the same variable twice, so the OR branch is dead. Same pattern at line 58. Collapse to `[[ -z "$term" ]]`.

(Pre-existing in `mctx/src/min_helper.sh`, propagated here by the byte-identical copy.)

crates/minimald/src/env.rs
twitchyliquid64: I wanted to duplicate it because at this stage mctx::Env is deprecated code to support the old minimal CLI, and  we wanted the flexibility to be able to tune the sessions sandbox how we want.

crates/minimald/src/env.rs
twitchyliquid64: Good catch!

crates/minimald/src/env.rs
twitchyliquid64: Fixed - plumbed error.

crates/minimald/src/env.rs
twitchyliquid64: Fixed.

crates/minimald/src/env.rs
norrietaylor: Makes sense - will note for next time.

norrietaylor (approved):

PR #427, opened 2026-06-16
chore: re-arranged the code in the sessions module to prepare for theclient/daemon split

evanspearman:
This is a refactor to prepare for the client/daemon split in session/loadout logic. There are some other minor changes that were required due to some of the moves, but no functional changes.

## Summary by CodeRabbit

- **New Features**
  * Added a policy hooks API that can externally decide or approve undecided variables and patch-source items during session resolution.

- **Refactor**
  * Consolidated policy, provenance, and expansion behavior into shared core types for consistent attribution of decisions and errors.
  * Improved patch source enumeration, including optional symlink-aware canonicalization and provenance-carrying results.

- **Breaking Changes**
  * Public module/API layout was reorganized; the previous policy/vars surfaces were removed and replaced with new client/core/wire exports.
  * Error types for variable/patch handling were updated to the new core primitives model.


twitchyliquid64 (approved):
I think this is good to go after a `cargo clippy --all-targets --fix --allow-dirty -- -D warnings && cargo fmt`

PR #508, opened 2026-06-21
ci(macos): self-provision libkrun via brew

norrietaylor:
Make the macOS CI install `libkrun` itself instead of depending on hand-provisioned self-hosted-runner state.

## Why
The three macOS jobs (`build-macos`, boot/autospawn e2e, minvmd FFI smoke) only **verified** `/opt/homebrew/lib/libkrun.dylib` existed and errored otherwise — libkrun was expected to be pre-provisioned on the self-hosted Apple Silicon runner. That runner state silently went missing, so **`build-macos` fails on every Rust code PR** (`libkrun.dylib not found`). `build-macos` is on the critical path of the minimald networking epic — it blocks #507 and all of #496–#502.

History note: there was never a libkrun install/fetch step in `ci-macos.yml` (`scripts/fetch-libkrun.sh` exists but fetches the **Linux** `.so` package and is only wired into `ci-linux-kvm.yml`). So this is a runner-provisioning gap, not a regression — and the durable fix is to stop depending on runner state.

## Change
Add an idempotent provision step before each Verify guard:
```yaml
- name: Provision libkrun (slp/krun tap)
  run: brew install slp/krun/libkrun
```
- Same `slp/krun` tap the runner setup already used → supply-chain surface unchanged.
- `brew install` is a no-op when already present.
- Drops the now-inaccurate "CI does not install it" comment.

## Self-validating
This PR's own `build-macos` (+ boot/autospawn/FFI) jobs run the new step first, so a green run proves the fix end-to-end.

## Summary by CodeRabbit

* **Chores**
  * Updated macOS CI workflow to automatically self-provision required dependencies across multiple test jobs on ARM64 self-hosted runners, improving build automation.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -128,6 +128,12 @@ jobs:
           persist-credentials: false
       - name: Toolchain
         uses: dtolnay/rust-toolchain@stable
+      - name: Provision libkrun (slp/krun tap)
+        # Self-install libkrun so macOS jobs no longer depend on hand-provisioned
+        # runner state — a missing libkrun.dylib silently broke every code PR.
+        # Idempotent (a no-op when already present); same slp/krun tap the runner
+        # setup used, so the supply-chain surface is unchanged.
+        run: brew install slp/krun/libkrun
       - name: Verify libkrun is available
         run: |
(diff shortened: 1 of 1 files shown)


bryan-minimal (approved):
LGTM

PR #543, opened 2026-06-23
docs(spikes): restore dns-hostname-mechanism spike doc lost from main

norrietaylor:
The `docs/spikes/2026-06-20-dns-hostname-mechanism.md` spike (the `*.localhost` + host-proxy DNS decision) merged in #487 is **missing from `main`** — a later sdd-spec stale-base derivation reverted it (same class as norrietaylor/spectacles#317; cf. the #529 cleanup). #500 adopts this spike's finding, so the doc should be present.

Restored verbatim from the #487 merge commit `6cc2dde`. Docs-only.

## Summary by CodeRabbit

* **Documentation**
  * Updated hostname resolution format to `.localhost` with systemd-resolved configuration guidance for Linux distributions.


docs/spikes/2026-06-20-dns-hostname-mechanism.md
```diff
@@ -0,0 +1,308 @@
...
+  - localhost
+  - rootless
+---
+
+## Question
+
+Does `*.localhost` wildcard hostname resolution work rootlessly (no per-invocation
+root) on common Linux distributions targeted by `minimald` users — Ubuntu 22.04+,
+Fedora 38+, Debian 12 (stable), and Arch Linux? Or does the alternative
+`*.min.local` domain approach require a one-time resolver configuration step
```
msample, line 27: *.min.internal may be better unless we are specifically hoping to integrate to mDNS/Bonjour?

bryan-minimal (approved):
LGTM

docs/spikes/2026-06-20-dns-hostname-mechanism.md
norrietaylor: Great catch and just in the nick of time. 

The bot fleet had diverged from the spec. 


Claude says

> The networking spec's plan is **R5 "PTask hostnames in host DNS" → B5 "`*.localhost` + host-side proxy (or one-time `/etc/resolver`)"** (plus a named gvproxy-webproxy + macOS PAC-file path), tagged "Needs decision (DNS) — zero-root path exists".
The load-bearing piece is the **host-side proxy** doing hostname→PTask routing, with resolution **host-side** — so it's systemd-agnostic (our sandboxes and microVMs have no systemd, and per the spec they never need to resolve anything; the host does). The TLD (`.localhost` vs `.min.internal`) is then just proxy/resolver config, not the mechanism.
We diverged in #500/#546 by narrowing B5 down to a systemd-resolved probe (R3.4), which both drops the host-side proxy and breaks in the no-systemd runtime. Re-aligning #500 to the spec's B5 host-side-proxy model — `.min.internal` is fully viable under it if preferred.

norrietaylor: Proceeding on **same-component**: #500 lands `net/proxy.rs` as the routing core and #502 extends it with TLS/mTLS — one proxy, not two. Defaulting client→proxy to **proxy-env/PAC** (zero-root, cross-platform); shout if you want `/etc/resolver` instead. Re-aligning #546 to this now.

norrietaylor:
__

docs/spikes/2026-06-20-dns-hostname-mechanism.md
norrietaylor: Adopting `.min.internal` per your call. Under the B5 host-side egress proxy the TLD is a free label (proxy routes by `Host:` header, no resolver), so no mDNS/Bonjour tie-in — `.min.internal` is the cleaner, semantically-honest choice for internal PTask services. Swapping it across the proxy/registry/tests/spec in #546 now.

Issue #572, opened 2026-06-25
DM1/3/4: run gvproxy on the host with a per-PTask vsock shuttle (minvmd-owns-vm-gvproxy)


norrietaylor:
## Problem
Wherever **minimald runs inside a minvmd-managed libkrun VM** (DM1 macOS/HVF, DM3/DM4 Linux/KVM), the networking data plane is wired to the wrong side of the VM boundary. Per the spec, gvproxy must run on the **host** (minvmd-owned) with a per-PTask **vsock shuttle** inside the guest; egress is a single NAT through that host gvproxy to the host network. The current implementation instead spawns gvproxy **inside the guest**, which has no host uplink — so an OwnIp PTask gets a `100.64.x.x` switch IP but **no real egress** (DNS / outbound fail).

## Scope (not mac-specific)
The gap depends on the VM boundary, not the host OS:
- **DM1** (macOS + libkrun VM) — affected.
- **DM3 / DM4** (Linux host running minvmd's libkrun VM) — **same gap**: minvmd and the guest minimald are the same binaries; minimald-in-guest spawns gvproxy in-guest identically.
- **DM2** (native Linux, minimald on the host, no minvmd VM) — **NOT affected, correct as-is**: minimald spawns gvproxy locally and PTasks attach via SCM_RIGHTS fd-pass. No vsock shuttle needed (no VM boundary).

## Spec (intended VM-boundary topology)
`docs/specs/03-spec-networking/networking-with-diagrams.md` (DM1 diagram) and `architecture.md`:
- "One gvproxy per host" — on a VM-boundary host it runs on the **host**, rootless, outside the libkrun VM.
- Guest runs only a per-PTask shuttle: `tap fd ↔ vsock frame` (raw L2 frames, **not** a second TCP/IP stack).
- "Egress is a single NAT through gvproxy to the host network — one TCP/IP stack in the path on either platform."
- Decisions: `shared-gvproxy-per-host`, `minvmd-owns-vm-gvproxy` (on DM1/3/4 minvmd owns the VM's gvproxy; minimald owns it only on DM2), `ownip-ptask-fd-pass` (fd-pass on DM2, vsock shuttle on DM1/3/4).

## Current behavior (the gap)
[...]

PR #570, opened 2026-06-24
feat(minimal2): add --network and --ingress flags to activate

norrietaylor:
## Summary
`minimal activate` hardcoded `NetworkMode::default()` (HostNet) and an empty policy, so `NoNet`/`OwnIp` and ingress port mappings had no CLI surface — they were reachable only via the netns proofs that construct a `Record` directly (`crates/minimald/src/rpc.rs:675`). This adds the activate flags.

## Changes
`crates/minimal2/src/main.rs`:
- `--network <no-net|host-net|own-ip>` (default `host-net`; no-flag behavior unchanged) via a local `CliNetworkMode` `ValueEnum` + `From<CliNetworkMode> for sessions::NetworkMode`, keeping the `sessions` crate free of a clap dependency.
- `--ingress EXT:INT[/PROTO]` (repeatable; PROTO defaults `tcp`) parsed into `sessions::PortMapping`. Malformed specs and non-tcp/udp protocols are rejected at parse time.
- Surface the daemon's typed `CreateSession` validation error (e.g. ingress on a non-`own-ip` session, privileged host port) instead of the generic `returned an error from the daemon` line.

```
Options:
  -n, --name <NAME>                Optional session name
      --network <NETWORK>          Network mode: no-net, host-net (default), or own-ip [default: host-net] [possible values: no-net, host-net, own-ip]
      --ingress <EXT:INT[/PROTO]>  Static ingress port mapping `EXT:INT[/PROTO]` (PROTO = tcp|udp, default tcp). Repeatable. Requires `--network own-ip`
```

## Scope
- Default behavior is unchanged: a flag-less `minimal activate` still creates a `HostNet` session with empty policy.
- Daemon-side `Record::validate_policy` is the enforcement point and is unchanged — the CLI builds the `Record` and surfaces the daemon's verdict.
- No egress flag (egress allow-list enforcement is deferred to #553); `--ingress` sets static `port_mappings` only, not `dynamic_allowed_range`.

## Verification
Docker `rust:1.95` + `protobuf-compiler`, `--locked`, `CARGO_INCREMENTAL=0`:
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p minimal2 --all-targets -- -D warnings` — clean
- `cargo test -p minimal2` — 6 passed (3 new parser tests: tcp default, explicit proto, malformed/bad-proto rejection)
- `minimal activate --help` renders the flags above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

[...]

PR #575, opened 2026-06-25
[WIP] feat(minimald): run attach --command inside the session PTask sandbox (DM2)

norrietaylor:
`minimal2 attach <session> --command <cmd>` ran `<cmd>` as a plain host process (`TokioExec` → `/bin/sh -c` on the host), ignoring the session entirely — so it never entered the PTask sandbox or its network namespace. This routes it through a real sandbox2/hakoniwa container built with the session's `NetworkMode`, mirroring the interactive-shell launch path.

## Scope: DM2 (native Linux) only
This is the **native minimald** exec-into-PTask path (DM2), where minimald spawns gvproxy locally and the OwnIp PTask attaches via fd-pass. The VM-boundary deployments (DM1 macOS, DM3/DM4 Linux+minvmd) need the host-gvproxy + per-PTask vsock shuttle from **#572** — out of scope here. DM2 is "correct as-is" per #572.

## Changes
- **`exec.rs`** — new `SandboxedExec` (replaces `TokioExec` for the plain-command branch under `cfg(not(test))`); builds an `Env` from the session record with `.with_network_mode(...)`, runs `/bin/sh -c <cmd>` in the container, bridges stdio through the existing `bridge()`/`Process` machinery. For `NetworkMode::OwnIp` it wires the namespace to the per-host switch via `attach_own_ip` and holds the `(env, own_ip)` guard for the command's lifetime. `min run` and git-receive paths unchanged.
- **`session_host.rs`** — factored the launcher's env construction into a shared `pub(crate) build_session_env` (exec and the interactive shell now build identical sandboxes); made `attach_own_ip` `pub(crate)` so exec reuses the exact tap/relay/ingress logic (no duplication).
- **`sessions.rs`** — `pub(crate) ManagerHandle::net_switch()` accessor so the exec path reaches the one per-host `GvproxySwitch`.

Two incidental fixes found while verifying: a piped one-shot command must not `set_session_leader()` (ENOTTY on a pipe), and the OwnIp startup race (command exits before the tap is in the netns) is gated behind a one-byte stdin handshake released after the attach completes.

## Coordination with #572 — shared seam
This and #572 both touch the OwnIp attach path. **Overlap is in `session_host.rs` around `attach_own_ip`** (here: made `pub(crate)` + factored `build_session_env`; #572: adds the DM1/3/4 tap→shuttle branch). Otherwise orthogonal (DM2 local-spawn+fd-pass vs DM1/3/4 host-gvproxy+vsock-shuttle). Whoever lands second rebases the `attach_own_ip` seam.

## Dependencies for end-to-end CLI testing
The exec code is independent and compiles/tests on `main`, but driving it end-to-end from `minimal2` needs:
- **#570** (`--network`/`--ingress`) to create own-ip/ingress sessions from the CLI, and
- **#573** (workspace upload) to populate the session worktree — without it a freshly-activated session is empty and compose fails at `minimal.toml not found`.

## Verification
- `cargo build/test/clippy/fmt -p minimald` — clean (incl. the existing `exec.rs` e2e tests + a new `plain_command_carries_session_network_mode`).
- Empirical, native daemon (DM2): **no-net** `attach --command` → only `lo`, no `eth0`, curl fails; **own-ip** → switch tap `mtap0_2` + `100.64.x.x` IP (daemon log: "attached OwnIp PTask to gvproxy switch ip=100.64.0.3"). Before: both showed the host's `eth0`/`192.168.5.x`. (Worktree hand-seeded for the test pending #573.)

## Summary by CodeRabbit

* **New Features**
  * Plain `attach --command` requests now run in the target session’s sandboxed environment.
  * Commands can now use the session’s network setup, including isolated “own IP” mode.

* **Bug Fixes**
  * Improved reliability for short-running commands in isolated networking mode.
  * Prevented commands from exiting before networking is fully ready, reducing intermittent failures.

### DIFF
--- a/crates/minimald/src/exec.rs
+++ b/crates/minimald/src/exec.rs
@@ -374,6 +374,245 @@ impl Process for TokioProcess {
     }
 }
 
+/// Production [`Exec`] for a plain `attach --command <cmd>` request: runs the
+/// argv through `/bin/sh -c` *inside the session's PTask sandbox* — a
+/// `sandbox2`/`hakoniwa` container honouring the session's [`NetworkMode`] —
+/// rather than as a bare host process.
+///
+/// The sandbox is built exactly like the interactive shell launcher
+/// ([`crate::session_host::SandboxLauncher::launch`]) via the shared
+/// [`build_session_env`](crate::session_host::build_session_env) helper, so a
+/// command sees the same rootfs and network namespace an interactive attach
+/// would. A fresh sandbox is built per command (there is no persistent PTask to
+/// reuse yet); that is acceptable for the one-shot exec path.
+///
+/// For [`NetworkMode::OwnIp`] the freshly-unshared network namespace is wired
+/// onto the per-host gvproxy switch (the same `attach_own_ip` the launcher
+/// uses), and the resulting attachment is held alongside `env` for the
+/// command's whole lifetime.
+///
+/// The struct is compiled in all configs so its record→sandbox wiring can be
+/// unit-tested; only the [`Exec`] impl (which builds a real sandbox) is gated
+/// to production, since the sandbox package set is unavailable under test.
+// Under test the `Exec` impl below is not compiled, so the fields the producer
+// reads (argv/name/username/ingress/net_switch) look unused; the wiring is
+// still exercised by `plain_command_carries_session_network_mode`.
+#[cfg_attr(test, allow(dead_code))]
+pub(crate) struct SandboxedExec {
+    /// The shell command line, run as `/bin/sh -c <argv>`.
+    pub argv: String,
+    /// Display name for the sandbox env (derived from the session record).
+    pub name: String,
+    /// Username for the sandbox env; `None` falls back to the env default.
+    pub username: Option<String>,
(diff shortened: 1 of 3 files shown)


crates/minimald/src/exec.rs
```diff
@@ -374,6 +374,245 @@ impl Process for TokioProcess {
...
+/// command's whole lifetime.
+///
+/// The struct is compiled in all configs so its record→sandbox wiring can be
+/// unit-tested; only the [`Exec`] impl (which builds a real sandbox) is gated
+/// to production, since the sandbox package set is unavailable under test.
+// Under test the `Exec` impl below is not compiled, so the fields the producer
+// reads (argv/name/username/ingress/net_switch) look unused; the wiring is
+// still exercised by `plain_command_carries_session_network_mode`.
+#[cfg_attr(test, allow(dead_code))]
+pub(crate) struct SandboxedExec {
```
twitchyliquid64, line 401: The exec impl type for running a task already exists, see TaskExec

norrietaylor:
Closing as will not do.

PR #581, opened 2026-06-26
feat(minimald,minvmd,minimal2): end-to-end VM/guest/session networking integration

norrietaylor:
End-to-end networking across the VM, guest-daemon, and session layers: a
session opened through `minimal2` now gets working egress, managed DNS, an mTLS
reverse proxy, and static ingress, with the host-side proxies reachable on the
host loopback.

## What this wires together

The networking primitives existed in pieces (gvproxy switch [#561](https://github.com/gominimal/minimal/pull/561),
mTLS reverse proxy [#554](https://github.com/gominimal/minimal/pull/554), static
ingress + session policy [#556](https://github.com/gominimal/minimal/pull/556)).
This connects them into one path that works from `minimal2` down to an
in-sandbox shell.

### VM layer (`minvmd`)
- Spawns the **one** host gvproxy switch and registers a **single** vsock shuttle
  port (`add_vsock_port2(.., listen=false)`, port 1024) that libkrun dials
  through to the host gvproxy `-listen` socket. Registered for every boot (not
  only own-ip), so the daemon's root-netns tap and each own-ip PTask tap can
  attach. There is **no in-VM bridge**: each tap opens its own vsock connection
  over that one port and relays its raw L2 frames to the host gvproxy, which is
  the switch (so two own-ip PTasks on the same host talk via the host switch).
- Own-ip boot now hard-errors if the gvproxy binary is missing instead of
  silently booting a session with no network (`MINVMD_GVPROXY_BIN`); a
  present-but-unlaunchable binary only degrades a non-own-ip boot.
- Guest RAM default is arch-conditional — 2048 MiB on x86_64 (hole-safe: a 4 GiB
  guest straddles the 32-bit MMIO hole and panics the kernel before `/init`),
  4096 on aarch64 — and overridable via `MINVMD_VM_RAM_MIB`; sized so the guest
  tmpfs rootfs does not hit `StorageFull`.

### Guest layer (`minimald`)
- **Root-netns egress** (`guest.rs`): brings up `eth0`, configures the interface
  over AF_INET ioctls (no `ip` dependency), installs `resolv.conf`, and attaches
  to the gvproxy switch over vsock — this is what gives the daemon itself egress
  (e.g. to clone an upstream repo during scaffold).
- **In-sandbox interactive session** now works inside the microVM: switch_root
  via `MS_MOVE`+`chroot(".")` instead of a bare `chroot`, fixing the
  userns-vs-chroot `EPERM` (hakoniwa exit 125); devpts mounted with a ptmx
  instance; `/tmp` tmpfs; shell exec'd at `/usr/bin/bash`.
- **Own-ip ingress** applied over the gvproxy control channel
  (`ControlChannel::Unix` / `::Vsock`).
- **Host-side proxies started in-VM and published on the host loopback**
  (`server.rs`): egress/DNS proxy (`:7654`) always, mTLS reverse proxy (`:7655`)
  under `networking-proxy`. In-VM they bind `0.0.0.0` (decoupled from `eth0`
  readiness) and only a port whose listener actually bound is exposed on the
  host via the gvproxy forwarder. DM2 (native) binds host loopback directly.

### Session / CLI layer (`minimal2`, `sessions`)
- `activate --network {no-net|host-net|own-ip}` and `--ingress HOST:GUEST[/proto]`.
- Session policy round-trips ingress mappings; validation rejects ill-formed
  policies at the boundary.

### Networking abstraction (`sandbox2`)
Following review feedback, the per-sandbox network wiring lives behind a
`sandbox2::Network` trait (`HostNet`/`NoNet` built-ins + a `GvproxyNetwork` impl
for own-ip), configured on the sandbox `Config` and applied by `sandbox2` itself,
so **tasks** — not only interactive sessions — get networking through one seam; a
task now runs under its session's network mode (a no-net session's task no longer
leaks host egress). Teardown of the switch attach is explicit. The shared switch
primitives (subnet/MAC/`-config` rendering) moved into a dedicated `switch` crate,
and the two same-named `GvproxySwitch` types were renamed `GvproxySupervisor`
(minvmd, owns the host gvproxy process) / `SwitchClient` (minimald, the
allocator + switch client) to make the per-model ownership clear. The gvproxy
**process** lifecycle stays in minimald (DM2) / minvmd (DM1) — `sandbox2` only
wires an already-running switch into a sandbox netns.

## Temporary scaffolding (call out)

`crates/mctx/src/scaffold.rs` is a **temporary, self-contained copy** of the
`minimal.toml` generator (`generate_mfile` / `scaffold_default_mfile`, packages
`["base"]`), used to scaffold a default `minimal.toml` for an empty session so an
attach does not fail with "minimal.toml not found". It is marked temporary in the
code. The intended end state is to share the real generator rather than copy it;
this copy avoids a larger refactor inside this integration PR and should be
removed when that refactor lands.

The scaffold clone runs inside the async session-actor; the blocking clone is
fenced with `tokio::task::block_in_place` on the daemon's multi-thread runtime
(guarded by runtime flavor so it falls back to a plain call rather than panicking
on a current-thread caller).

## Test plan & status

Full plan + executable runner committed under
`docs/specs/03-spec-networking/` (`test-plan.md`, `test-plan.sh`). Every
per-session assertion runs **inside the sandbox via the full interactive session
path** (`minimal2 attach <sess>`, no `-c`); `attach -c` is never used because it
runs on the daemon host in the guest root netns and bypasses the sandbox.

## Deployment Mode (DM) Key 

- DM1 — macOS (Apple Silicon) + Linux VMs (minimald in each VM, UDS-over-vsock)
- DM2 — native Linux, a single minimald running directly on the host (no microVM), reachable over a UDS or TCP-localhost; multi-user hosts may run one per user
- DM3 — native Linux + Linux VM(s) with minimald in each
- DM4 — DM2 + DM3 combined (host minimald and VM minimald)
- DM5 — any of the above, network-exposed with authN/authZ

Re-run on the pinned rootfs that now carries `ip`/`nsenter` (own-ip PTasks
launch). Verdicts are per **deployment model**: DM1 (macOS + VM, HVF), DM2
(native Linux, host-native daemon, **no VM**), DM3 (native Linux + VM, KVM).

| TC | DM1 (macOS+VM) | DM2 (native, no VM) | DM3 (Linux+VM) |
|----|----------------|---------------------|----------------|
| TC1 no-net isolation  | **PASS**       | **PASS** ⁴            | **PASS** ²      |
| TC1b host-net egress  | **PASS**       | **PASS** ⁴            | **PASS** ²      |
| TC2 own-ip egress     | **PASS**       | **FAIL** ³            | **PASS** ²      |
| TC2 same-host peer    | **BLOCKED** ¹  | **FAIL** ³            | **BLOCKED** ¹   |
| TC3 managed DNS :7654 | **BLOCKED** ¹  | **FAIL** ³            | **BLOCKED** ¹   |
| TC4 static ingress    | **PARTIAL**    | **PARTIAL** ³         | **PARTIAL** ²   |
| TC5 policy validation | **PASS**       | **PASS** ⁴            | **PASS** ⁴      |
| TC6 policy round-trip | **PASS**       | **PASS** ⁴            | **PASS** ⁴      |
| TC7 mTLS :7655        | **PARTIAL**    | **PARTIAL** ³         | **PARTIAL** ²   |
| TC8 ssh-forward       | **BLOCKED** ¹  | **FAIL** ³            | **BLOCKED** ¹   |
| TC9 mesh CLI          | **PASS**       | **PASS** ⁴            | **PASS** ⁴      |

DM1 is the authoritative macOS interactive run (evidence: TC1 no ifaces;
TC1b/TC2 in-sandbox curl→200; TC4/TC6 ingress mapping round-trips; TC5 all 3
rejected; TC7 no-cert→**401**, with-cert data path on ¹).

¹ Blocked by [#588](https://github.com/gominimal/minimal/issues/588) — the
  **first** own-ip attach per boot succeeds, every subsequent attach wedges. Hits
  the microVM models (DM1/DM3) identically; the same `BLOCKED`/`PARTIAL` rows are
  the own-ip *data path* (TC2-peer, TC3, TC4-data, TC7-data, TC8).

² DM3 shares DM1's microVM/own-ip path (minimald-in-VM over the vsock shuttle);
  the first own-ip egress completes and the rest hit ¹, reproduced on Linux/KVM.
  Not re-run TC-by-TC — identical guest/session code to DM1.

³ **DM2 own-ip before this fix.** A host-native daemon is unprivileged, so own-ip
  failed three independent ways: tap setup `EPERM` (`ip`/`nsenter` don't inherit
  `setcap +ep`); with `setcap`, sandbox `uid_map` `EPERM` (dumpable→`SUID_DUMP_ROOT`);
  and no in-netns DNS (`127.0.0.53` stub is unreachable). Fixed in the stacked
  **[#589](https://github.com/gominimal/minimal/pull/589)** (rootless in-process tap +
  dumpable reset + gateway DNS): DM2 own-ip egress and `curl example.com`→200 now
  pass. TC4/TC7 are `PARTIAL` here for the same reason as DM1 (CLI/policy half
  passes; own-ip data path was the failing half).

⁴ DM2/DM3 CLI & policy rows (TC5/6/9) and the no-net / host-net sandbox (TC1/1b)
  share the host-agnostic / shared sandbox path; the mTLS-401 and managed-DNS
  proxies bind host loopback directly on DM2 (no gvproxy forwarder).

### Known gap — own-ip PTask attach reliability ([#588](https://github.com/gominimal/minimal/issues/588))
The earlier rootfs gap is **resolved**: `.minimal/minimal.toml` `locked_commit` is
pinned to a rootfs build that carries `ip`/`nsenter`, so `move_tap_into_netns`
succeeds, an own-ip PTask launches, and in-sandbox egress over the gvproxy shuttle
returns 200 (TC2).

The remaining own-ip data-path TCs are blocked by a **separate** reliability bug
([#588](https://github.com/gominimal/minimal/issues/588)): the **first** own-ip
attach per boot succeeds, but every **subsequent** own-ip attach in the same boot
wedges (the vsock shuttle / switch per-tap state is not released on session
destroy). Because each data-path TC destroys and re-activates a fresh own-ip
session, only the first (TC2 egress) completes; TC2-peer, TC3, TC4-data,
TC7-data, and TC8 hang. This is not introduced by this PR — the vsock shuttle
landed in [#561](https://github.com/gominimal/minimal/pull/561) — and is tracked
separately. The proxy infrastructure itself is verified (`:7655` → 401, `:7654`
reachable on host loopback). The same attach issue is why own-ip **tasks**
currently fall back to HostNet.

## Platform caveat

The full interactive test plan was run on **macOS** (HVF, DM1 / microVM path).
**Linux/KVM** (DM1 on Linux) is exercised in CI: the arch-conditional RAM fix
clears the x86_64 MMIO-hole boot panic, the `minvmd-linux-kvm-e2e` lane boots the
VM, and `netns-integration` exercises the Linux sandbox-netns path. DM2 (native
Linux, no VM) own-IP is delivered separately by the stacked
[#589](https://github.com/gominimal/minimal/pull/589) (rootless host-native
own-IP), where it is verified end to end. The host-specific bit (mTLS cert dir)
is already OS-aware in the runner.

## Summary by CodeRabbit

* **New Features**
  * Added `minimal2` Activate networking controls: `--network` and repeatable `--ingress EXT:INT[/PROTO]` for static TCP/UDP mappings.
  * Introduced configurable “own-IP” guest networking with host gvproxy integration and guest egress support; micro-VM mode now applies it automatically.
* **Bug Fixes**
  * Auto-scaffolds a default `minimal.toml` for newly attached empty session workspaces.
  * Improved guest rootfs prep (writable `/tmp`, better `devpts`/PTY handling) and interactive SSH attach (`ssh -tt`).
* **Documentation**
  * Added a networking CLI test plan and companion script.


crates/minimald/src/session_host.rs
```diff
@@ -654,12 +685,15 @@ async fn attach_own_ip(
...
+    use crate::net::SwitchTransport;
+    use crate::net::switch::{
+        attach_to_switch, attach_to_switch_vsock, move_tap_into_netns, open_tap,
+    };
 
     // Allocate a lease and ensure gvproxy is running, snapshotting the control
-    // socket and subnet under one lock; the slow tap/relay work runs unlocked so
-    // concurrent attaches don't serialize on the namespace plumbing.
-    let (lease, sock, subnet) = {
+    // socket, subnet, and transport under one lock; the slow tap/relay work runs
```
twitchyliquid64: Network initialization and wiring with the switch should be part of the sandbox2 crate, this stuff being configured via the Config builder & Sandbox struct. 

Otherwise networking will only work in minimald sessions and not tasks or any other minimal sandbox.

twitchyliquid64 (commented):
More generally, what’s the lifecycle of the gvproxy objects?

twitchyliquid64 (commented):
I think we are missing a ‘network’ abstraction that ties in with the sandbox2 crate to do the network configuration, has multiple impls for the different kinds of network (host vs gvproxy etc), and keeps networking logic away from sandbox consumers like minimald/mctx.

This is a bunch of rework obvi, so down if you want to record this for later and just get this in now.

norrietaylor:
Addressed the review — built the sandbox2 network abstraction now rather than deferring.

## gvproxy lifecycle (your question)

One gvproxy **process** per host; the owner depends on the deployment model:

(as per spec https://github.com/gominimal/minimal/blob/main/docs/specs/03-spec-networking/networking-with-diagrams.md) 

- **native Linux:** `minimald` owns it. 
- **libkrun VM:** `minvmd` owns it on the host, outside the guest. 

The two same-named `GvproxySwitch` types were the main source of confusion — renamed to `GvproxySupervisor` (minvmd, the process owner) and `SwitchClient` (minimald, the allocator/relay client) in `9e0a8b4c`.

note: I have not tried deploying minimald and minvmd in a parallel configuration on the same host. I am sure that will be it's own adventure

crates/minimald/src/session_host.rs
norrietaylor: Done in 62fe1c4d / c6c2b534

norrietaylor:
> I think we are missing a ‘network’ abstraction that ties in with the sandbox2 crate to do the network configuration, has multiple impls for the different kinds of network (host vs gvproxy etc), and keeps networking logic away from sandbox consumers like minimald/mctx.
> 
> This is a bunch of rework obvi, so down if you want to record this for later and just get this in now.

I'm glad to do the work, but you are gonna have to review it :) 

edit: this joke used to be funnier before claude

norrietaylor:
CodeRabbit review — all findings addressed in the latest push. Triage + dispositions:

**Fixed (correctness):**
- `switch.rs` host-shuttle connect/write had no timeout → 5s bound on the vsock connect + `/connect` upgrade (`daff5d91`).
- `sessions.rs` HostNet hostname route stranded on rename → `RenameSession` re-points it (`daff5d91`).
- `sandbox2` run_with_cancel leaked the sandbox/switch on `Network::attach`/writer/wait failure → kill+reap the child on attach failure and tear the `NetGuard` down before propagating any error (`a202aa38`). Covers the `gvproxy_network.rs:63`, `lib.rs:786`, and `lib.rs:947` findings.
- `switch::SwitchSubnet::new` stored host bits verbatim → canonicalize `base` to the network address (`f12f48bc`).

**Fixed (docs/tooling):**
- `justfile codesign-minvmd` signed the debug binary while building release → signs `target/release/minvmd` (`79accea7`).
- `OWN_IP` env doc now lists all accepted truthy values (`1`/`true`/`yes`/`on`) (`79accea7`).
- test-plan: companion script name corrected to `test-plan.sh`; stale per-TC statuses refreshed (TC1/TC1b now PASS; own-ip TCs cite the real `ip`/`nsenter` rootfs blocker); trap-based cleanup added so an interrupted run tears down sessions + the TC8 ssh-forward (`79accea7`).

**Note:** the `mctx::scaffold` unbounded-blocking-clone nitpick is the explicitly-temporary scaffold (removed once sessions receive their project files); its `block_in_place` is now flavor-guarded so it can't panic on a current-thread runtime (`daff5d91`).

Verified: cross-clippy (minimald+mctx, aarch64-musl), native clippy (minvmd+switch), switch + minvmd net tests, fmt — all clean.

twitchyliquid64 (approved):

Issue #588, opened 2026-06-26
own-ip attach intermittently hangs forever on AF_VSOCK missed wakeup


norrietaylor:
## Symptom

- Own-ip session attach (and same-host peer-over-switch) intermittently wedges with no timeout, no error, no progress. Reproduced: the TC2 peer step in `docs/specs/03-spec-networking/test-plan.sh` hung ~27 min until killed; recurs on the second own-ip attach.
- Single-attach own-ip egress succeeds reliably (HTTP 200) across multiple runs. The hang is timing-dependent, surfacing under a second attach / two taps on one host.

## Hypothesis

- Missed wakeup at the libkrun AF_VSOCK transport readiness layer (guest↔host shuttle, port 1024, `add_vsock_port2(.., listen=false)`).
- The `relay.rs` frame pump is correct — it uses the `AsyncFd::readable()/writable()` + `try_io` idiom that re-arms readiness on `WouldBlock`. The suspect is below it: the vsock fd never signals readable when bytes arrive, so `readable().await` parks forever.
- Consistent with the prior finding that a blocking-I/O shuttle (socat) is reliable (the kernel wakes a blocking read directly) while the async readiness-tracked vsock path intermittently misses the wakeup.

## Investigation

- Instrument the vsock relay: log readiness events vs bytes-available on the shuttle fd at hang time.
- Compare blocking vs `AsyncFd` shuttle read under the two-attach repro.
- Check libkrun vsock notification semantics (edge vs level) for the `add_vsock_port2(listen=false)` dial path.

## Fix (direction, TBD by investigation)

- Add a connect/attach watchdog so a wedged vsock can't hang an attach indefinitely (bounded retry + surfaced error), and/or correct the readiness registration that drops the wakeup.

## Acceptance

- Own-ip attach + same-host peer reachability completes within bound across ≥20 consecutive runs.
- A wedged vsock fails the attach with an error inside a bounded timeout instead of hanging.

## References

- `crates/minvmd/src/net/relay.rs`, `crates/minvmd/src/net/shuttle.rs`
- Shuttle origin: #561
- Repro harness: `docs/specs/03-spec-networking/test-plan.sh` (TC2 peer)

norrietaylor:
Sharper repro from two clean runs on the new rootfs (own-ip attach now launches — ip/nsenter present in the guest):

[...]

PR #590, opened 2026-06-28
[WIP] fix(minimal2): serialize concurrent attach establishment (#588)

norrietaylor:
Mitigation for [#588](https://github.com/gominimal/minimal/issues/588): a second overlapping `minimal2 attach` to an own-ip session wedges (no shell).

Stacked on [#581](https://github.com/gominimal/minimal/pull/581) (base branch `feat/networking-host-exposure`) because it depends on that PR's own-ip attach path.

## Root cause
Two own-ip session SSH handshakes racing through **libkrun's host→guest virtio-vsock muxer** leave one session wedged — its `RequestShell` is never delivered to `session_host`, so the second concurrent attach hangs. Verified the defect is in libkrun's vsock device under concurrent connections (not `tokio-vsock`, which is correct; not minimald logic). libkrun 1.19.0 fixed only the *sequential* direct-vsock case. Full investigation in [#588](https://github.com/gominimal/minimal/issues/588).

## Fix (in-repo workaround, pending the upstream libkrun fix)
`minimal2 attach` spawns a short-lived detached `attach-lock` helper that holds an exclusive `flock` for a bounded establishment window (default 4000ms, `MINIMAL_ATTACH_ESTABLISH_WINDOW_MS`) before exec-ing ssh:

- A concurrent attach blocks on the same lock and starts its SSH handshake ~one window later, so the racy handshake/`RequestShell` phase is **serialized** instead of overlapping the muxer.
- Single attaches are uncontended — the parent only waits for the lock-acquired signal, not the full window — so steady-state latency is unchanged.
- `exec ssh` is preserved (clean interactive TTY); the helper holds the lock in the background and releases on window expiry.

## Verification (macOS/HVF, DM1)
- Two concurrent own-ip attaches both reach a shell with working egress: **3/3** rounds (HTTP 200 each), **20/20** establish-only. Daemon boot.log confirms both get `RequestShell` + `attached OwnIp PTask`, with session B attaching ~4s after A (the flock window).
- Pre-fix: first-spawned wedged **3/3**.
- `cargo fmt` / `cargo clippy -p minimal2 --all-targets -D warnings` / `cargo test -p minimal2` clean.

## Notes
- Unblocks concurrent own-ip sessions (e.g. `docs/specs/03-spec-networking/test-plan.sh` TC2 same-host peer) without waiting on the upstream libkrun fix.
- The upstream defect remains tracked in [#588](https://github.com/gominimal/minimal/issues/588); a draft upstream issue (with the build/repro notes) is prepared for containers/libkrun.

References: [#588](https://github.com/gominimal/minimal/issues/588)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attach reliability by reducing race conditions when starting interactive sessions.
  * Added a short coordination window during connection setup to help prevent intermittent failures in environments with socket concurrency issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3486,6 +3486,7 @@ dependencies = [
  "clap",
  "clap_complete",
  "dirs",
+ "libc",
  "minimald-rpc",
  "minvmd",
  "ot",
--- a/crates/minimal/Cargo.toml
+++ b/crates/minimal/Cargo.toml
@@ -10,6 +10,7 @@ clap_complete.workspace = true
 camino.workspace = true
 chrono.workspace = true
 dirs.workspace = true
+libc.workspace = true
 minimald-rpc.workspace = true
 paths.workspace = true
 russh.workspace = true
--- a/crates/minimal/src/main.rs
+++ b/crates/minimal/src/main.rs
@@ -38,6 +38,11 @@ enum Command {
     /// Proxy stdio to a daemon UDS socket (used as an SSH ProxyCommand).
     #[command(hide = true)]
     Proxy(ProxyArgs),
+    /// Hold the attach-establishment lock briefly, then exit. Internal: spawned
+    /// by `attach` to serialize concurrent session handshakes around the libkrun
+    /// vsock concurrency bug (#588). Prints `ok` once the lock is held.
+    #[command(hide = true)]
+    AttachLock(AttachLockArgs),
     /// Forward a local TCP port to a remote address inside a PTask via SSH
     /// (R4.8, R4.9).
     ///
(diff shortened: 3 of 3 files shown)


crates/minimal2/src/main.rs
```diff
@@ -600,6 +659,39 @@ async fn cmd_attach(global: &GlobalArgs, args: AttachArgs) -> Result<(), ()> {
...
+fn cmd_attach_lock(args: AttachLockArgs) -> Result<(), ()> {
+    use std::io::Write;
+
+    let file = std::fs::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .truncate(false) // the file is only a flock target; never written
+        .open(&args.lock_path)
+        .map_err(|e| eprintln!("attach-lock: open {}: {e}", args.lock_path.display()))?;
+    let mut lock = fd_lock::RwLock::new(file);
```
twitchyliquid64: Can we use flock instead like elsewhere? https://github.com/gominimal/minimal/blob/98b7ae988835178362d651b9a788aa0279a4685b/crates/lcache/src/read_tracker.rs#L127-L131

Or does that not work on darwin

norrietaylor: I will investigate as I clean up these conflicts

norrietaylor: Done — switched to `libc::flock(LOCK_EX)` matching lcache's read-tracker (dropped the fd-lock crate). Blocking, minus `LOCK_NB`, so a contender waits instead of failing. Works on Darwin: two concurrent holders serialize ~window apart, and the mitigation prevents the wedge. Also rebased onto main (past the crate rename), so the conflict's resolved.

Side note from re-verifying: #588 is a real but *rare* libkrun race (~0.4%/attach), not the deterministic bug I first reported — details on the issue.

twitchyliquid64 (approved):

PR #597, opened 2026-06-29
feat(sessions): update session status

evanspearman:
Storage scaffolding for the daemon-side draft-session flow.

- New `SessionStatus { Draft, Active }` on `Record` (defaults to
  `Active`, `#[serde(default)]` for backward-compat with on-disk
  records predating the field). Store records the value verbatim;
  state-machine rules belong to higher layers.
- New `Loader::save(&mut self, key, record)`. Overwrites the row
  in place; refuses id changes (`InvalidInput`), stale keys
  (`NotFound`), and real name collisions (`AlreadyExists`).
  Same-name re-save is a no-op. Atomic via the existing
  `temp + renameat2` plumbing.
- `Loader::rename` is now sugar over `save` (one write path).
- `Record` derives `PartialEq + Eq`; trivially supported by
  every field.

## Summary by CodeRabbit

* **New Features**
  * Sessions now have a persisted lifecycle state (**draft/active**) that is saved and restored across reloads.
  * For older saved sessions that don’t include this field, the lifecycle state automatically defaults to **active**.

* **Bug Fixes**
  * Session creation/activation requests now include the lifecycle state consistently across exec/SFTP/RPC flows.
  * Session save/rename now applies stricter validation to prevent stale data “resurrection,” handle overwrites correctly, and resolve duplicate-name collisions consistently.

### DIFF
--- a/crates/sessions/src/lib.rs
+++ b/crates/sessions/src/lib.rs
@@ -264,8 +264,34 @@ impl fmt::Display for SessionId {
     }
 }
 
+/// Lifecycle status of a [`Record`].
+///
+/// `Draft` covers a session whose composition is still in flight —
+/// an id has been allocated and a stub record persisted, but the
+/// composition pipeline hasn't yet produced a finalized
+/// `Composition`. `Active` is the finalized, ready-to-use state.
+///
+/// Defaults to `Active` so on-disk records predating this field
+/// (which were always finalized at create time) deserialize
+/// correctly.
+///
+/// The store layer doesn't enforce transitions between states — it
+/// records the status verbatim. State-machine rules (e.g. "only
+/// the composition pipeline can promote `Draft` → `Active`") live
+/// in the manager actor.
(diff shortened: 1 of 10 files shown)


crates/sessions/src/lib.rs
```diff
@@ -264,8 +264,34 @@ impl fmt::Display for SessionId {
     }
 }
 
+/// Lifecycle status of a [`Record`].
```
twitchyliquid64, line 267: `SessionStatus` implies its the state machine for the session- given its more the loadout vibes, do we want to call it `Loadout` or `LoadoutStatus` ?

evanspearman: I think it is actually part of the session lifecycle itself. A session is created as a `Draft` once the client sends the CreateSession request, along with the loadout information, and then it becomes active once the client gets back to it with the verdict on package and project contributions. It's not really a loadout specific thing.

twitchyliquid64 (approved):

PR #602, opened 2026-06-30
feat(sessions): tighten daemon composer pending/ready routing

evanspearman:
Depends on https://github.com/gominimal/minimal/pull/600

- Wires up daemon-side Phase 2 composition routing: `SessionComposer::compose` now returns a `ComposeOutcome::{Ready, Pending}` instead of a finalized `Composition`, and the manager handler in `minimald` consumes the new shape.
- Daemon-collected vars and patches route back to the client as a `ContributionResponse` so the client's `policy` and `PolicyHooks` gate them — the daemon never runs user policy. Daemon-collected packages and lifecycle hooks have no per-item verdict slot in the wire schema, so they pass through into the assembled `Composition` on the all-decided path.
- Today every internal `CreateSession` caller still sends an empty `WireContribution`, so the manager hits the Ready fast path. The Pending branch is guarded with a defensive `InvalidInput` error in the manager until the `SubmitVerdict` handler lands (next step).
- New `PendingComposeState` retains the daemon-side packages, hooks, and the client's wire contribution across a Pending outcome so Phase 4 can finalize after the client's verdict comes back.
- Doc + test coverage: `crates/sessions/docs/COMPOSITION.md` describes the new 4-phase flow; tests cover the empty-input fast path, the packages/hooks-only fast path, the var → Pending routing, and that the client's wire contribution merges into the composition byte-for-byte on the all-decided path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session creation handling for client-supplied contributions, preventing submitted details from being silently ignored.
  * Clearer rejection of unsupported in-progress/pending session outcomes with an explicit input error.
  * Reduced session name collisions in repeated end-to-end runs by generating a unique session name each time.
* **Documentation**
  * Updated session composition documentation to clarify phase responsibilities and the current on-wire behavior for session responses (with `Pending` not reachable yet).

### DIFF
--- a/crates/minimald/src/sessions.rs
+++ b/crates/minimald/src/sessions.rs
@@ -7,6 +7,8 @@ use crate::{
 use paths::DaemonAbsPath;
 use sessions::{
     SessionId,
+    core::compose::ComposeOptions,
+    daemon::composer::{ComposeOutcome, SessionComposer},
     store::{DiskLoader, Loader, SessionKey, SessionObject},
     wire::request::WireContribution,
 };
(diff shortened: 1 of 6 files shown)


twitchyliquid64:
minimal/minimal2 crates were moved, but you can get a clanker to rebase with this prompt:

```
If you git fetch origin/main and merge it into the current branch, theres a conflict because
i renamed crates/minimal to crates/mip and crates/minimal2 to crates/minimal.

Can you do the fetch + rebase to fix the conflict?
```

evanspearman:
@twitchyliquid64 Thanks, it's rebased now

crates/minimald/src/sessions.rs
```diff
@@ -290,29 +295,80 @@ impl<L: Loader> Manager<L> {
...
+                        // the response so the client can correlate a
+                        // future `SubmitVerdict`). The `Ready` outcome
+                        // ignores it. We pass `nil` because the manager
+                        // doesn't allocate the id until `store.create`
+                        // below — and the `Pending` outcome is unreachable
+                        // today since no daemon-side contributors are
+                        // wired in. The defensive guard below catches
+                        // it anyway.
+                        let composer = SessionComposer::new(contribution);
+                        match composer.compose(SessionId::nil(), ComposeOptions::default()) {
```
twitchyliquid64, line 331: Hmm, thoughts on the contribution being passed through to the `Session` actor and keeping `Sessions` mostly as a dumb router? We would need to let the `Session` handle storage of the composition, implying its separate to the session record itself, but that seems doable (we already let the `Session` actor manage the dirs for the working tree + session "homedir")

twitchyliquid64: Oh I guess we manage the state of the composition in the session record.. hmmmm

twitchyliquid64 (approved):

PR #613, opened 2026-07-01
feat(minimald): implement Shutdown RPC

twitchyliquid64:
Tells the minimald to shutdown.

`force=false` - will error if there are active sessions.
`force=true` - will kill active sessions/connections, drain them, then exit.

```shell
$> echo "{\"force\": true}" | ssh -s  \
    -o ProxyCommand='socat - UNIX-CONNECT:/home/xxx/.local/state/minimal/providers/local-0/ssh.sock' \
    -o 'UserKnownHostsFile=/home/xxx/.local/state/minimal/providers/local-0/known_hosts' \
    local-0 minimald-v1-Shutdown
```

## Summary by CodeRabbit

* **New Features**
  * Added a SSH-based one-shot `Shutdown` RPC with a `force` option.
  * Server now stops accepting new connections and performs shutdown with a short grace period before terminating remaining session work.
  * Shutdown responses indicate whether termination started or was refused due to active sessions.

* **Bug Fixes**
  * New session operations are blocked once shutdown begins.
  * Non-forced shutdown is refused when sessions are live; forced shutdown proceeds and then rejects further session work.

* **Tests**
  * Added coverage for the above shutdown scenarios.

* **Chores**
  * Added a Unix-domain socket test client helper and updated workspace dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3549,6 +3549,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tokio-vsock",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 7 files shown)


0chroma (approved):
LGTM! well thought out ^^

PR #643, opened 2026-07-06
feat(rcache,mctx): read cache over HTTPS via MINIMAL_REMOTE_CACHE_URL (R2 read path)

bryan-minimal:
## Why

First code step of the GCS→R2 cache migration ([build-servers#145](https://github.com/gominimal/build-servers/issues/145)): let the cache **reader** fetch artifacts over plain unauthenticated **HTTPS** (from a Cloudflare R2 custom domain, or a GCS public URL) instead of the GCS client — so CI reads stop incurring GCS egress (~$460/mo, ~99% of the bill). This is the **read path only**; writes are unchanged (they stay on GCS, per the read-first plan).

## What

Reads are gated on a new **`MINIMAL_REMOTE_CACHE_URL`** env var (or `ConfigBuilder::with_remote_cache_read_url`). When set, the reader does plain HTTPS GETs against that base; when unset, it reads through the GCS client exactly as before (**default behaviour unchanged**).

- **`common/fetchers`** — new `AnyBackend`: a runtime-selectable `FetchBackend` that is *either* the GCS `Storage` client or a reqwest HTTPS `Client`, chosen at construction. This keeps a single `RemoteCache<AnyBackend>` type so the backend choice doesn't thread a type parameter through every consumer (orchestrator, bin-provider, materialize). A url/backend mismatch is a construction bug and panics rather than fetching the wrong thing.
- **`rcache/remote`** — `RemoteCache<AnyBackend>` constructors `new_any_https(url)` / `new_any_gcs(storage, bucket)`; hardened two progress-bar `content_length().unwrap()`s to `unwrap_or(0)` (a plain-HTTPS backend may omit `Content-Length`).
- **`mctx/config`** — `remote_cache_read_url` field + builder + a **pure** `resolve_remote_cache_read_url` (builder value wins, else `MINIMAL_REMOTE_CACHE_URL`, empties ignored) + getter.
- **`mctx/lib`** — `remote_cache()` returns `RemoteCache<AnyBackend>`: HTTPS when a read URL is configured, else the GCS client.
- **`orchestrator`** — `RemoteCache<GcsStorage>` → `RemoteCache<AnyBackend>`.

Verified that `auth=true` is unused anywhere (all callers read the public bucket anonymously) and that `remote_cache()` isn't called from build-servers, so this is self-contained.

## Correctness note

The base URL **must end in `/`** — otherwise `Url::join` replaces the last path segment instead of appending object names under it. There's a dedicated test for this footgun.

## Tests

- **`common`** (unit): `AnyUrl` join/filename for both variants incl. the trailing-slash footgun; `AnyBackend` get-dispatch happy path; and the url/backend **mismatch panics** (never a silent wrong fetch). *(5 tests, pass on host.)*
- **`mctx`** (unit): the pure read-URL resolver — builder-over-env precedence, env fallback, empty/whitespace ignored — plus a builder→config→getter round-trip. *(No process-env flakiness.)*
- **`rcache`** (integration): a real local HTTP server proving `new_any_https` **fetches + parses the index over the actual reqwest transport** (`AnyBackend::Https` → get/execute/chunk/content_length), that a **404 index yields an empty (not errored) cache** like the GCS path, and that a **malformed URL is rejected** cleanly.

*On property tests:* the risk here is I/O-forwarding + URL edge cases, not algebraic invariants, so targeted edge-case units + the end-to-end integration test give better signal than a proptest over a match arm.

## Not in scope

Writes (the `remote_writer.rs` S3 port) — deferred / folded into the signed-index work, per #145. This lands the read cutover that captures the egress savings.

## Summary by CodeRabbit

* **New Features**
  * Remote cache access now supports both GCS buckets and HTTPS mirrors.
  * Configuration can now use a unified remote-cache location, with separate read and write behavior.
  * Added backend-aware cache creation for flexible remote cache setup.

* **Bug Fixes**
  * Improved handling for remote cache reads when response sizes are unavailable.
  * Writing now clearly fails when configured with a read-only HTTPS mirror.

### DIFF
--- a/crates/mctx/src/config.rs
+++ b/crates/mctx/src/config.rs
@@ -4,15 +4,16 @@ use std::{
 };
 
 use checkouts::ManagerHandle;
+use common::fetchers::{AnyUrl, GcsUrl, ReqwestUrl};
 use ot::OpTracker;
 
 /// The errors possible in configuration or building a configuration.
 #[derive(Debug)]
 pub enum ConfigError {
     IO(&'static str, PathBuf, std::io::Error),
-    /// Configured remote cache bucket name is empty or whitespace-only.
+    /// Configured remote cache location (`remote_cache_url`) is empty or unparseable.
     /// Carries the offending value for diagnostics.
-    InvalidRemoteCacheBucket(String),
+    InvalidRemoteCacheUrl(String),
 }
 
 impl fmt::Display for ConfigError {
@@ -21,8 +22,8 @@ impl fmt::Display for ConfigError {
             ConfigError::IO(ctx, path, e) => {
                 write!(f, "{} I/O error at path {}: {}", ctx, path.display(), e)
             }
-            ConfigError::InvalidRemoteCacheBucket(bucket) => {
-                write!(f, "invalid remote cache bucket: {:?}", bucket)
+            ConfigError::InvalidRemoteCacheUrl(url) => {
+                write!(f, "invalid remote_cache_url: {:?}", url)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


crates/mctx/src/config.rs
```diff
@@ -57,6 +68,7 @@ pub struct ConfigBuilder {
     vcs_manager: Option<ManagerHandle>,
     ot: Option<OpTracker>,
     remote_cache_bucket: Option<String>,
```
twitchyliquid64: Could we use this opportunity to simply by killing `remote_cache_bucket` and using `remote_cache_read_url` for both cases?

crates/mctx/src/config.rs
```diff
@@ -265,6 +302,12 @@ impl Config {
     pub fn remote_cache_bucket(&self) -> &str {
         &self.remote_cache_bucket
     }
+    /// Returns the remote-cache *reader* base URL override (plain HTTPS), if set
+    /// via the builder or `MINIMAL_REMOTE_CACHE_URL`. `None` reads via the GCS
+    /// client against [Self::remote_cache_bucket].
+    pub fn remote_cache_read_url(&self) -> Option<&str> {
```
twitchyliquid64: Could this return the typed `AnyUrl` ?

crates/mctx/src/lib.rs
```diff
@@ -386,29 +387,40 @@ impl Context {
...
+            Some(self.config.index_dir())
         };
 
-        let res = RemoteCache::new_with_gcs_bucket(
-            backend,
-            self.config.remote_cache_bucket(),
-            if force_fresh {
-                None
+        // When a read URL is configured (e.g. a Cloudflare R2 custom domain via
+        // MINIMAL_REMOTE_CACHE_URL), fetch artifacts over plain HTTPS to avoid
```
twitchyliquid64: > fetch artifacts over plain HTTPS to avoid GCS egress cost

Wait what? we were doing this wrong the whole time?

crates/mctx/src/lib.rs
```diff
@@ -386,29 +387,40 @@ impl Context {
         &self,
         auth: bool,
         force_fresh: bool,
-    ) -> Result<RemoteCache<GcsStorage>, RemoteError<GcsError>> {
+    ) -> Result<RemoteCache<AnyBackend>, RemoteError<AnyRespError>> {
         let start = SystemTime::now();
-        let backend = if auth {
```
twitchyliquid64, line 391: So we still need the `auth` conditional here, as this method returns the GcsStorage used by two paths:

1. _Upload_ by buildbot/`mip`
2. Anonymous _Download_ by anyone

bryan-minimal:
Pushed `b2363231` — addresses all four:

1. **Killed `remote_cache_bucket`.** One `remote_cache_url` knob now (default = the GCS bucket): `gs://…` / bare name → GCS, `https://…` → HTTPS mirror.
2. **Resolution moved into `Config`.** `parse_remote_cache_url` returns the typed `AnyUrl`, and `Config::remote_cache_url()` hands back the ready-to-go "where to get the artifacts" location (plus `remote_cache_write_bucket()` for the writer). rcache's `new_any` now takes the resolved `AnyUrl` directly instead of a bucket string.
3. **Reworded the egress comment.** It's reading from *R2* that avoids egress, not HTTPS per se — an `https://` URL still pointed at GCS egresses the same.
4. **Kept the `auth` conditional** — authed for the buildbot/mip upload path, anonymous for CI reads.

Writer now returns `anyhow::Result` and hard-errors when the cache is configured as an https mirror (no writable bucket). Validated in the sandbox: `clippy -D warnings` clean + mctx/rcache/common tests green.

crates/mctx/src/config.rs
bryan-minimal: Done — one `remote_cache_url` knob now (default = the shared bucket): `gs://…` / bare name → GCS, `https://…` → HTTPS mirror. `remote_cache_bucket` is gone. (b2363231)

crates/mctx/src/config.rs
bryan-minimal: Yep — `parse_remote_cache_url` now returns the typed `AnyUrl`, and `Config::remote_cache_url()` hands it straight to `new_any`, so no bucket string gets threaded through anymore.

crates/mctx/src/lib.rs
bryan-minimal: Nope — good news is we weren't overpaying. That comment was just wrong (why I reworded it). Plain HTTPS → GCS egresses identically to the client; egress is billed on bytes leaving Google to the internet regardless of API surface (the client is HTTPS underneath too). The only lever is *where the bytes live* — moving them to R2 ($0 egress) is the ~$460/mo win. HTTPS is just what lets the reader point at the R2 mirror; HTTPS-at-GCS saves nothing.

crates/mctx/src/lib.rs
bryan-minimal: Kept it — the GCS branch still does authed (buildbot/`mip` upload) vs. anonymous (download) exactly as before. Only the HTTPS-mirror case skips building a `Storage` client, since mirror reads are unauthenticated public GETs.

twitchyliquid64 (approved):

PR #650, opened 2026-07-07
feat(sessions): wire project + package composables end-to-end into the sandbox

evanspearman:
Resolves https://github.com/gominimal/inbox/issues/198

## Summary

Wires the daemon-side session composables end-to-end into the sandbox
launcher, so a session activated against a project actually gets that
project's packages, vars, and stack material — not the hardcoded 5-package
literal that previously ran.

Before this change: every session launched with the same static
`["base", "bash", "socat", "coreutils", "claude-code"]` package set,
the mfile was never consulted, and `Composition` was computed on the
daemon and then thrown away.

After: `CreateSession` builds a `Composition` from all three composables
(loadout, project, per-package), stashes it, and `SandboxLauncher::launch`
unions its packages and vars over a minimal baseline.

Every package included in the session goes through this pipeline:
1. The client's requested packages plus the project's declared packages
   (from `[session] packages`, `[stack] build_packages`/`runtime_packages`,
   and the graph-level Stack's own package lists) form the top-level set.
2. Their transitive runtime closure is walked.
3. Each package with `env_state_wiring` or `env_dir/file_mappings`
   contributes a `PackageComposable`; the resulting vars and patches are
   tagged `Source::Package` and gated through the same `PatchPolicy` /
   `VarsPolicy` any other contribution goes through.
4. Composition is materialized and merged into `EnvArgs` at launch.

The `Env::build` legacy path stays intact for `mip run <task>` — task
runs still use `SetupForPackages` as before. Session launches opt into
the composition path via `EnvArgs::without_package_attr_wiring()`.

## Notable pieces

- `mfile::ProjectComposable` and `mfile::PackageComposable`: two new
  `Composable` impls that read the `[session]` block and package
  `BuildSpec.attrs` respectively.
- `minimald::sessions::composables`: daemon-side pipeline that resolves
  the project mfile + graph, builds composables, and drives the
  `SessionComposer`. Split into its own module so each stage is unit-
  testable without spinning the manager mainloop.
- `Env::build` composition branch: bypasses `SetupForPackages` output for
  `env_vars`/`fs_mappings`; state dirs derive from resolved var values
  shaped `/state/<prefix>` instead.
- `mctx::Context` split into `Context` + `Arc<DaemonContext>` so per-
  session mctx state doesn't duplicate the daemon-scoped config, vcs,
  and cache setup.
- `ApproveProjectAndPackage` client hook: default `minimal activate`
  policy that auto-approves `Source::Project` / `Source::Package`
  contributions (project activation implicitly consents to what the
  project declares).
- Example project at `crates/sessions/example_project/` with a
  demo-punchline mfile — comment out `[session]`, re-activate, watch
  every demo command break.

## Deferred

- Composition patches → sandbox. Blocked on the file-upload path.
- Lifecycle hooks. Blocked on in-sandbox exec plumbing.
- Credential-class fs mappings. Dropped with a `tracing::warn!` pending
  the secrets strategy. `read_only = true` is ignored (with warn); minimal
  is copy-based so the flag has no meaning today.

## Summary by CodeRabbit

* **New Features**
  * Added optional project session configuration via `minimal.toml`’s `[session]` (packages, vars, patches, lifecycle hooks) and session-aware composition.
  * Enabled daemon-scoped shared state so runtime setup consistently reuses daemon config/cache.
  * Added composition-derived package/env wiring for real session launches (with session-provided overrides).
* **Bug Fixes**
  * Updated environment/cache handling to use daemon-backed locations for more reliable setup.
  * Improved pending activation gating: unapproved items are allowed only when they originate strictly from daemon-side project/package sources.
  * Enhanced handling of empty/absent session config and forward-compatible unknown fields.


crates/minimald/src/sessions.rs
```diff
@@ -177,6 +198,45 @@ pub struct Manager<L: Loader = DiskLoader> {
...
+    /// sandbox exec plumbing.
+    ///
+    /// Growth is bounded by the number of Active-but-unattached
+    /// sessions: [`GetSession`] drains the entry, [`DestroySession`]
+    /// and [`AbortSession`] drop it defensively. Without those two
+    /// drains a create-and-forget client (activate without attach,
+    /// then destroy) would leak indefinitely, which is why both
+    /// paths call `compositions.remove(&id)`.
+    ///
+    /// Held behind [`Arc`] so the drain-to-actor hop and any
```
twitchyliquid64: Can we do a pass through all the verbose commentary and trim them down to just the essentials?

i.e. this could just talk about how this is the pending composition for a in-construction session, discussion of the point of Arc and unbounded growth in the edge case is probably unnecessary

crates/minimald/src/session.rs
```diff
@@ -99,15 +99,86 @@ pub struct Session<S: SessionObject> {
...
+    ///
+    /// [`Manager::CreateSession`]: crate::sessions::Manager
+    /// [`SubmitVerdict`]: crate::sessions::Manager
+    #[cfg_attr(test, allow(dead_code))]
+    composition: Option<Arc<Composition>>,
+
+    /// Lazily-built [`mctx::Context`] rooted at this session's
+    /// workspace, cached across [`Self::context`] calls so repeated
+    /// attach / task-exec paths don't rebuild the daemon setup and
+    /// re-parse the workspace mfile.
```
twitchyliquid64, line 117: There are cases where we want a fresh `Context`, like its totally valid for a user to edit their `minimal.toml` to add some new tasks and then run them - thats why i had it originally as being reconstructed every time someone called for a context via the session handle.

Happy to leave it like this for now, but probably want to circle back to this later.

crates/minimald/src/session_host.rs
```diff
@@ -752,6 +767,57 @@ impl SessionLauncher for SandboxLauncher {
             armed: true,
         });
 
+        // Package + env-var union of the launcher baseline and every
```
twitchyliquid64, line 740: Would a cleaner way to do this just be to have some "default" loadout/composition that gets unioned in earlier, rather than having a baseline set of packages and stuff that gets combined at the last moment ?

evanspearman: I like that approach. I think I'll do that in a follow on as part of the loadout loading flow though.

crates/minimald/src/sessions.rs
evanspearman: Agreed. I'll get Claude to do a pass over all the doc comments in sessions and minimald.

twitchyliquid64 (approved):

PR #653, opened 2026-07-07
fix(graph): bound untrusted sizes in the wire decoder (record length + build_count)

bryan-minimal:
## What

`GraphReader::read_record` decoded a record's payload length from an untrusted LEB128 varint and passed it straight to `vec![0u8; len as usize]` (`wire.rs`, both the sync and async readers), with no upper bound. A malformed graph could therefore declare a huge length and drive a multi-gigabyte allocation that aborts the process.

This is NET-reachable: the build graph arrives over the remote-execution channel (`Graph::from_bytes`), and the allocation happens **before** the stream checksum is verified, so a valid checksum is not required to trigger it.

## How it was found

The new `graph_from_bytes` cargo-fuzz target (separate branch) hit it in ~672 executions with a **6-byte** input:

```
03 ad ad ad ad 0a
```

`0x03` is a record tag; `ad ad ad ad 0a` is the length varint, which decodes to ~2.8 GiB → `libFuzzer: out-of-memory (malloc(2779469485))`.

## Fix

- Add `MAX_RECORD_LEN` (512 MiB) — comfortably above any real build-spec or inlined local-file record.
- Add `check_record_len`, shared by the sync and async readers so they stay in lock-step, returning a new `WireError::RecordTooLarge { len, max }` before allocating.
- Regression test `oversized_record_length_is_rejected_not_allocated` seeded with the exact fuzzer input, asserting a clean `WireError` rather than an OOM.

## Validation

`cargo test -p graph --lib wire` — 18 passed (including the new regression test); `cargo clippy -p graph --lib -- -D warnings` clean. Relying on CI for the full sandbox run.

## Follow-ups (not in this PR)

- The same decoder still has an out-of-bounds slice on inner length fields (`materialize_local_file`) and an unsanitised entry `filename`; the fuzzer couldn't reach them until this OOM was fixed. Re-fuzzing on top of this should surface them next.

### DIFF
--- a/crates/graph/src/wire.rs
+++ b/crates/graph/src/wire.rs
@@ -37,6 +37,26 @@ const TAG_STACK: u8 = 0x06;
 const TAG_SUPPLY_CHAIN: u8 = 0x07;
 const TAG_FOOTER: u8 = 0xFF;
 
+/// Upper bound on a single framed record's payload length. A malformed stream
+/// can declare an arbitrary varint length; without this cap the decoder would
+/// `vec![0u8; len]` and abort the process on a multi-gigabyte allocation from a
+/// few bytes of input. 512 MiB comfortably exceeds any real build-spec or
+/// inlined local-file record.
+const MAX_RECORD_LEN: u64 = 512 * 1024 * 1024;
+
+/// Rejects a framed record whose declared length exceeds [`MAX_RECORD_LEN`], so
+/// a malformed varint cannot drive an unbounded `vec![0u8; len]` allocation.
+/// Shared by the sync and async readers to keep them in lock-step.
+fn check_record_len(len: u64) -> Result<(), WireError> {
+    if len > MAX_RECORD_LEN {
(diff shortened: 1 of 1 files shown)


twitchyliquid64 (approved):

bryan-minimal:
Folded in a second fix. Re-fuzzing on top of the record-length bound (which held: 1.3M execs, stable ~200 MB RSS, ~9x more coverage) surfaced a **second, distinct OOM** — the header's untrusted `build_count` was passed straight to `Arena::/HashMap::with_capacity`, so a 35-byte input declaring ~97M builds drove a `malloc(280089522504)` (~280 GiB).

Fix: cap the pre-size hint at `MAX_PREALLOC_BUILDS` (the collections still grow to fit the specs actually decoded). Both sync and async readers; regression test seeded with the fuzzer input.

`cargo test -p graph --lib wire` → 19 passed; `cargo clippy -p graph --lib -D warnings` clean.

PR #655, opened 2026-07-07
fix(installer): build + ship libkrun on MacOS

twitchyliquid64:
* Build libkrun from source- more mature than copying someone elses binary + we can strip out the other deps.
 * Ship it on MacOS, for now adjacent to the binary (which is weird, but i want to get it working at all).

I'm not 100% sure how the linking paths look, so we will iterate on that + ive added some prints in the build jobs to give us some more information.

## Summary by CodeRabbit

* **New Features**
  * macOS arm64 releases now include a packaged `libkrun` dynamic library.
  * Release artifacts and installer manifests now reference the new macOS arm64 component.

* **Bug Fixes**
  * Added checks to help ensure the shipped macOS library is self-contained before release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -228,6 +228,12 @@ jobs:
                   fi
             - name: Build release minvmd binary
               run: cargo build --release -p minvmd --bin minvmd
+            - name: Print libkrun load command (informational)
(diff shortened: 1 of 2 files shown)


norrietaylor (approved):
LGTM

PR #656, opened 2026-07-07
fix(graph): bounds-check local-file records and reject unsafe names

bryan-minimal:
Second batch of graph wire-decoder hardening, found by continuing the fuzz→fix→re-fuzz loop on top of #653 (which this stacks on and which fixed the two allocation OOMs).

## Findings (both in `materialize_local_file`)

**H1 — out-of-bounds slice panic (NET-reachable).** The payload was sliced with a wire-supplied `header_len` (varint) and `data_len` (JSON) without bounds checks:
```rust
serde_json::from_slice(&payload[pos..pos + header_len])   // pos+header_len can exceed payload.len()
let data_end = data_start + file_header.data_len as usize; // unchecked, can overflow
let file_data = &payload[data_start..data_end];
```
Once #653 removed the allocation OOMs that masked it, seeding the fuzz corpus with a valid graph-that-inlines-a-local-file let libFuzzer reach and corrupt these fields:
```
panicked at wire.rs: range end index 315 out of range for slice of length 207
==ERROR: libFuzzer: deadly signal
```

**H3 — filename path traversal.** The wire-controlled `filename` was joined onto the temp dir unsanitised (`subdir.join(&file_header.filename)`), so `..`/absolute names could escape the per-file directory.

## Fix

- Validate every wire-controlled offset stays within `payload` (`checked_add` + `<= len`), returning a clean `WireError` instead of slicing out of bounds — covers both the overflow and the out-of-range cases.
- Reject filenames that are absolute or contain `..`/root/prefix components before joining.
- Regression tests drive `materialize_local_file` directly for each; the existing local-file round-trips still pass (legit bare names are unaffected).

## Fuzzing status

After these fixes, re-fuzzing runs clean: **1.49M execs, coverage 3405 (up from 152 at the start of the campaign), RSS stable ~290 MB, no new crashes** — the fuzzer-reachable crash surface of the decoder is now hardened. The harness lives on `test/fuzz-graph-wire`.

## Validation

`cargo test -p graph --lib wire` → 21 passed; `cargo clippy -p graph --lib -- -D warnings` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of uploaded file data to prevent crashes from malformed inputs.
  * Rejected unsafe file paths, including absolute paths and directory traversal attempts, so extracted files stay within their intended temporary location.


crates/graph/src/wire.rs
twitchyliquid64, line 51: Nit but i think having tens of thousands of packages is in the realm of possibility, so maybe 32k?

twitchyliquid64 (approved):

crates/graph/src/wire.rs
twitchyliquid64: Nvm this is just the pre allocate, not a hard limit afaict

PR #663, opened 2026-07-07
feat(installer): implement uninstall, fix update components on MacOS

twitchyliquid64:
* Implement `--uninstall`
 * Fix bug where components with updates were not updated on OSX

## Summary by CodeRabbit

* **New Features**
  * Added an offline `--uninstall` mode to the installer.
  * Added support for `--force`, `--purge`, and `--dry-run` during uninstall.

* **Bug Fixes**
  * Improved reinstall behavior on macOS so signed binaries are re-downloaded when the installed bytes no longer match the expected release.
  * Uninstall now safely skips modified files unless forced, and handles repeated or empty uninstalls cleanly.


bryan-minimal (approved):
LGTM, only small issue would be someone who used the old version of this installer, but this isn't in use yet.

PR #667, opened 2026-07-08
ci(dogfood): canary the R2 cache read path via MINIMAL_REMOTE_CACHE_URL

bryan-minimal:
Points the one package-cache-reading CI job at the Cloudflare R2 mirror to validate the read-path migration end to end (build-servers#145 / minimal#643). This is the canary.

## What
The `dogfood` job (`minimal run build-smoke`) is the **only** job in this repo that reads the shared package cache on every push/PR — so its `minimal run` step now sets `MINIMAL_REMOTE_CACHE_URL=https://cache.minimal.farm/`. Everything else stays on GCS as the control.

- **Isolated & representative.** `setup-minimal` fetches the CLI from the separate **shim** bucket (a different bucket, unaffected by this var), and `minimal --no-fetch check` reads no remote cache. So this is a clean A/B: `dogfood` reads artifacts from R2, all other jobs read from GCS.
- **Index stays fresh.** The mutable `index.shisha` is bypassed to GCS at the edge (infra#277 — verified: `cache.minimal.farm/index.shisha` returns `no-store` and matches GCS byte-for-byte), so it never freezes through Sippy. Only the immutable, content-addressed artifacts come from R2.
- **This PR's own `dogfood` run is the first live test** — if reading artifacts over R2 breaks or misses, this PR's CI shows it before merge.

## Why (the point)
GCS cache egress is ~99% of the bill (~$460/mo, almost all CI reads). R2 egress is $0. This canary measures whether pointing reads at R2 collapses that egress **with hit-rate parity and zero sha256 mismatches**, before rolling it out more broadly.

## Rollback
Delete the `env` block → the reader falls back to the GCS client (default, unchanged). Trailing slash on the URL is required (`Url::join` footgun).

## Not in scope
Writes stay on GCS. The shim/CLI bucket stays on GCS. Broader rollout (more jobs, or baking the default) comes after this canary shows the egress collapse holds.

## Summary by CodeRabbit

* **Chores**
  * Improved Minimal build smoke-test configuration.
  * Added remote cache support to help make build validation more consistent and efficient.
  * Expanded workflow guidance for easier maintenance and troubleshooting.


norrietaylor (approved):

twitchyliquid64 (commented):
I dont think the `minimal` binary thats run here is a recent one, but lets get this in and soon i'll switch it to `mip`

PR #672, opened 2026-07-08
feat(minvmd,minimald): per-VM writable /dev/vdb ext4 volume

norrietaylor:
Implements first stage a per-VM writable ext4 volume that carries the package cache, rootfs-staging trees, and session state on one filesystem off RAM onto durable storage.

The host creates a sparse raw file (ensure_sparse_raw); the guest formats it on first boot via mkfs.ext4 keyed on ext4 superblock detection (byte offset 1080 magic 0x53EF) — idempotent, platform-portable (macOS has no mke2fs). Subsequent boots detect the superblock and skip mkfs.

Next steps include 
   * shutdown quiesce + crash safety
   * and session persistence, self-heal, upload-on-resume, provider index

Notes: minimum libkrun is now 1.19.0

## Summary

* **New Features**
  * Added automatic provisioning and mounting of a writable per-VM state volume (ext4 when needed) and an optional writable data volume disk at boot.
  * Added environment-variable controls for the data-volume disk’s direct I/O and sync mode.

* **Bug Fixes**
  * Improved ext4 state-volume handling to detect formatting/mount readiness and avoid unnecessary reformatting.

* **Documentation**
  * Clarified when reduced RAM “stop-gap” behavior becomes safe.

* **Tests / CI**
  * Added unit tests for ext4 detection and sparse raw volume creation/idempotency.
  * Strengthened Linux/macOS CI to verify required `libkrun` artifacts and the expected disk export.


crates/minimald/src/main.rs
```diff
@@ -353,6 +353,28 @@ async fn async_main() -> Result<(), MainError> {
         }
     }
 
+    // R1.5/R1.6: mount the per-VM writable data volume (/dev/vdb) and, when it
+    // mounts, relocate cache + state onto it so builds hardlinking from the cache
+    // stay on one filesystem (the EXDEV fix). When no volume is attached (the
+    // transitional case), state stays on the tmpfs default. Unit 2 (R2.4/R2.5)
+    // gates READY on this and removes the silent-fallback path.
+    if is_minimal_microvm() {
```
twitchyliquid64: Having lots of `is_minimal_microvm()` blocks isnt ideal for combinatorial complexity.

Instead, add a new+hidden+default-None field `mk_mount_state_volume: Option<String>` to `ListenArgs`, and instead set it to `Some("/dev/vdb".to_string())` in our existing `if is_minimal_microvm() {` block above at ~line 289. We can also safely update the new values for `minimal_state_dir` and `minimal_cache_dir` there.

The actual call to `mount_state_volume` can stay here, gated by `if let Some(...) = &listen_args.mk_mount_state_volume { ...`

crates/minimald/src/guest.rs
```diff
@@ -239,6 +239,154 @@ pub fn enter_rootfs(device: &str) -> std::io::Result<()> {
...
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            format!(
+                "data volume {device} is too small to format: {device_bytes} bytes \
+                 (minimum {MKFS_MIN_DEVICE_BYTES})"
+            ),
+        ));
+    }
+    let fs_blocks = device_bytes.saturating_sub(MKFS_MARGIN_BYTES) / EXT4_BLOCK_BYTES;
+    let status = std::process::Command::new("mkfs.ext4")
```
twitchyliquid64, line 342: Except for the documented+justified `lazy_itable_init=0,lazy_journal_init=0`, I'm not super keen on specifying options in a call to `mkfs.ext4` because the defaults are where they are due to a decade of tuning and blood. Do we really need all these tunables set?

crates/minvmd/src/volume.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use std::fs::OpenOptions;
+use std::io;
+use std::path::{Path, PathBuf};
+
+/// Default size of a freshly provisioned data volume: 32 GiB. The image is a
+/// sparse raw file, so the host allocates only the blocks the guest actually
+/// writes (allocate-on-write, verified by the sparsity proof gate); this value
+/// is the ceiling, not an up-front cost.
+pub const DEFAULT_VOLUME_BYTES: u64 = 32 * 1024 * 1024 * 1024;
```
twitchyliquid64: Nb: given how hard a resize is, and that this is sparse, is there a reason we cant make it mega-big? like 256Gb?

crates/minvmd/src/volume.rs
```diff
@@ -0,0 +1,236 @@
...
+/// so it survives clean restarts. `XDG_STATE_HOME` is honoured directly rather
+/// than via [`crate::state::StateDir::default_path`] because the `dirs` crate
+/// ignores `XDG_STATE_HOME` on macOS — honouring it keeps the volume isolable in
+/// tests (and beside the state dir) on every platform.
+#[must_use]
+pub fn resolve_data_volume_path() -> PathBuf {
+    if let Some(explicit) = std::env::var_os(DATA_VOLUME_PATH_ENV).filter(|v| !v.is_empty()) {
+        return PathBuf::from(explicit);
+    }
+    let state_dir = std::env::var_os("XDG_STATE_HOME")
```
twitchyliquid64: Compute this with `minimal_state_dir`: https://github.com/gominimal/minimal/blob/121599698bfb524f9f6b89252543f872096c0ed6/crates/paths/src/lib.rs#L143

norrietaylor: Good catch!

crates/minvmd/src/volume.rs
norrietaylor: I can try that, but we will pay for it on first boot.

crates/minimald/src/guest.rs
norrietaylor: - -i 65536 is to pay for lazy_itable_init=0: eager-zeroing the inode table is only cheap because we shrank it. Default ratio → ~2M inodes / ~512 MiB table on 32 GiB
- -b 4096  this is becuase libkrun reserves a small trailer on the raw backing file and shaves it off between boots. we hand mkfs a block count ~1 MiB below the device, and the count is in block-size units, so -b pins the unit.

crates/minimald/src/main.rs
norrietaylor: I can add a hidden `mk_mount_state_volume: Option<String>` field on `ListenArgs`, set to `Some("/dev/vdb")` in the existing microvm-config block, but `minimal_state_dir`/`minimal_cache_dir` must stay conditional on the mount succeeding, so they need to say down here.

crates/minvmd/src/volume.rs
norrietaylor: Good call — done, default is now **256 GiB** (`MINVMD_VOLUME_BYTES` still overrides). Since the image is sparse and offline resize is painful, sizing big up front is essentially free.

I measured first-boot cost across the four sizes to be sure the eager `mkfs` init doesn't scale with size (booted a fresh `/dev/vdb` at each, M-series host, `boot --foreground`):

| Volume | format → mount | mkfs metadata written | `/dev/vdb` mounted (kernel time) |
|-------:|:--------------:|:---------------------:|:--------------------------------:|
| 32 GiB  | 40 ms | 4.9 MiB | [0.074] |
| 64 GiB  | 36 ms | 5.4 MiB | [0.072] |
| 128 GiB | 37 ms | 6.5 MiB | [0.062] |
| 256 GiB | 46 ms | 8.5 MiB | [0.073] |

It's **flat** — the sparse backing plus the reduced inode ratio (`-i 65536`) means `mkfs.ext4` only writes ~5–8 MiB of metadata regardless of size, so format+mount stays ~40 ms at every size. So the size decision is fully decoupled from boot cost; 256 GiB it is.

(Numbers recorded in the `DEFAULT_VOLUME_BYTES` doc comment for posterity.)

crates/minvmd/src/krun/ctx.rs
```diff
@@ -250,6 +250,45 @@ impl Context {
         Ok(())
     }
 
+    /// Add a disk image as a virtio-blk device with cache/sync control
+    /// (`krun_add_disk3`). Unlike [`add_disk`][Self::add_disk], the caller
```
msample, line 254: Virtio-fs is to mount the host's file system.   Aren't we just having an ext4 formatted block device?

msample: oh, it is virtio-blk.  oops

msample (approved):

twitchyliquid64 (approved):

PR #682, opened 2026-07-09
feat(minimal): prompt to init when activate has no minimal.toml

0chroma:
## Summary

When `minimal activate` is run in a project without a `minimal.toml`, the client now prompts the user to create one instead of the daemon silently scaffolding a default (the temporary `mctx::scaffold_default_mfile` path, which does a blocking network git clone). This is the last remaining P0 item from gominimal/inbox#159.

### Changes

- **`cmd_activate`** now calls `ensure_mfile_or_prompt` after canonicalizing the project path. If no `minimal.toml` exists, it prompts `Would you like to create one? [Y/n]`:
  - **Yes (Linux)**: runs the same init flow as `minimal init` (detect stack, generate `minimal.toml`, confirm, write)
  - **Yes (macOS)**: bails with a clear message — `minimal init` is Linux-only due to `mctx`/`op` pulling in `hakoniwa`/`procfs`
  - **No**: bails with `No minimal.toml found. Run 'minimal init' to create one.`
- **`run_init_flow`** extracted from `cmd_init` so both `cmd_init` and the activate prompt share the same init logic (plan display, confirmation prompt, file write)
- **`activate_creates_session` test** updated to use a tempdir with a `minimal.toml` so the prompt doesn't fire

Closes the last P0 item on gominimal/inbox#159.

## Summary by CodeRabbit

* **New Features**
  * Activation now verifies the required project config file exists before starting a session.
  * If the file is missing on Linux, activation can guide you through creating it automatically.
* **Bug Fixes**
  * Activation no longer proceeds when the required project config file is absent.
  * Non-Linux systems now provide clearer instructions for creating the config file.
* **Tests**
  * Updated CLI integration test setup to include the required config file, preventing prompt-related flakiness.

### DIFF
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -819,6 +819,7 @@ fn ensure_mfile_or_prompt(
             .map_err(|e| anyhow::anyhow!("{e}"))?
     };
     run_init_flow(config, false)
+}
 
 /// Create a new session via the `CreateSession` RPC.
 pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(), anyhow::Error> {
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -772,6 +772,49 @@ async fn send_abort(client: &mut client::Client, session_id: sessions::SessionId
     }
 }
 
+/// Check for a `minimal.toml` at the project path. If missing, prompt
+/// the user to initialize one before proceeding with activation.
+fn ensure_mfile_or_prompt(
+    project_path: &camino::Utf8Path,
+    global: &GlobalArgs,
+) -> Result<(), anyhow::Error> {
+    if project_path.join(mfile::MFILE_NAME).exists() {
+        return Ok(());
+    }
+
+    eprintln!("\nNo {} found at {}.", mfile::MFILE_NAME, project_path);
+    eprint!("Would you like to create one? [Y/n] ");
+    std::io::stderr().flush().ok();
+
+    let mut input = String::new();
+    std::io::stdin()
+        .read_line(&mut input)
+        .context("reading stdin")?;
+    let trimmed = input.trim();
+    if !(trimmed.eq_ignore_ascii_case("y")
+        || trimmed.eq_ignore_ascii_case("yes")
+        || trimmed.is_empty())
+    {
+        bail!(
+            "No {} found. Run 'minimal init' to create one.",
+            mfile::MFILE_NAME
+        );
+    }
+
+    let config = if global.repo_dir.is_some() {
+        build_config(global).map_err(|e| anyhow::anyhow!("{e}"))?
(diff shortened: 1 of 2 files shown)


crates/minimal/src/lib.rs
```diff
@@ -725,6 +725,63 @@ async fn send_abort(client: &mut client::Client, session_id: sessions::SessionId
...
+        bail!(
+            "No {} found. Run 'minimal init' to create one.",
+            mfile::MFILE_NAME
+        );
+    }
+
+    #[cfg(target_os = "linux")]
+    {
+        let config = if global.repo_dir.is_some() {
+            build_config(global).map_err(|e| anyhow::anyhow!("{e}"))?
```
twitchyliquid64: 99% of `mctx`, `graph` etc we should be able to get working on non-linux, we just need to flag out the hakoniwa stuff and have stubs for the execution / run / process types.

Optional if you want to do that here (otherwise maybe leave a comment), but using that machinery here would be the ideal case.

crates/minimal/src/lib.rs
```diff
@@ -725,6 +725,63 @@ async fn send_abort(client: &mut client::Client, session_id: sessions::SessionId
...
+fn ensure_mfile_or_prompt(
+    project_path: &camino::Utf8Path,
+    global: &GlobalArgs,
+) -> Result<(), anyhow::Error> {
+    if project_path.join(mfile::MFILE_NAME).exists() {
+        return Ok(());
+    }
+
+    eprintln!("\nNo {} found at {}.", mfile::MFILE_NAME, project_path);
+    eprint!("Would you like to create one? [Y/n] ");
```
twitchyliquid64: This wont be the last time we need to prompt. Might be time to bring in a crate like dialoguer, or at least have a common codepath for confirmation prompts?

0chroma: Good call — extracted a shared `confirm` helper in the latest push. It deduplicates both prompt sites (activate missing-mfile and init confirmation). Skipped `dialoguer` for now since it's a new dep for a 2-call-site pattern, but easy to swap in later if prompts grow.

crates/minimal/src/lib.rs
0chroma: Agreed — the dep chain is accidental. `ProjectSetup::for_init` never touches the sandbox; it only needs VCS + graph + mfile parsing. I looked into the trait-extraction approach (split `sandbox2` into a lightweight `sandbox2-api` crate with the trait/err types, make `mctx` depend on the API unconditionally and on full `sandbox2` behind a feature). That avoids ~200 lines of `#[cfg]` and keeps the boundary clean.

Discussed with @norrietaylor / @msample — tracking as a follow-up rather than expanding this PR. Left a note on gominimal/inbox#159 with the direction.

twitchyliquid64 (approved):

twitchyliquid64 (approved):

PR #686, opened 2026-07-09
feat(sessions): loadouts, config, and debug commands

evanspearman:
Resolves https://github.com/gominimal/inbox/issues/202 and resolves https://github.com/gominimal/inbox/issues/203

## Summary

Ships end-to-end loadout support, a user client config file, and two debug
commands (`minimal dirs`, `minimal loadout list`). Adds a rolling
file-based daemon log for detached mode and per-session content logging.

**User-facing (`minimal`):**
- `--config-dir` global flag (routes through the new
  `paths::minimal_config_dir()`, XDG on Linux, `~/.config/minimal` on
  macOS for consistency with the state/cache dirs)
- `minimal loadout list` — enumerate loadouts from
  `<config>/minimal/loadouts/*.toml`, mark defaults from
  `[loadouts].default_loadouts` in `config.toml`
- `minimal dirs` — print all Config/State/Cache paths with existence
  markers (`*` = exists, `-` = missing, `?` = unresolved)
- `minimal activate --loadout NAME` (repeatable) / `--no-loadouts` to
  select which loadouts to apply; otherwise `[loadouts].default_loadouts`
  is used

**Library side (`sessions`):**
- New `sessions::client::config` — TOML with `deny_unknown_fields`,
  `[loadouts]` block for `default_loadouts` and `follow_symlinks`
- New `sessions::client::disk` — `list_loadouts` / `read_loadout_file`
  with per-entry `LoadoutEntry` results so listing tolerates malformed
  files without aborting
- Loadout `Inherit` vars pre-resolve from the process env;
  `NotPresent` → warn+drop rather than error
- Patch enumeration treats walkdir `NotFound` as warn+drop, so a
  loadout that opportunistically patches `~/dotfiles/helix/` doesn't
  fail activation on a host without that tree
- `ComposeOptions::with_follow_symlinks` owned builder

**Daemon (`minimald`):**
- `MINIMALD_DETACHED=1` routes tracing to a daily-rotated log file
  under `<state>/logs/minimald.log.YYYY-MM-DD`
- Per-session content logging: packages, vars, patches, hooks with
  their `Source` provenance
- `WindowAdjusted` SSH flow-control messages downgraded from warn to
  debug (they were spamming the log)

**Structure:**
- `crates/minimal/src/lib.rs` (2313 → 1613 lines) with the new logic
  split into `config`, `dirs`, `loadouts` modules
- `paths::minimal_config_dir()` alongside the existing `state`/`cache`
  helpers
- `just min *args` recipe for quick local invocations

## Summary

* **New Features**
  * Added `loadout` and `dirs` commands to list loadouts and inspect resolved config/state/cache directories.
  * Added `--config-dir` plus loadout activation controls (`--loadout`, `--no-loadouts`), and `minimald` rotation-backed logging in detached mode.
  * Added `min` convenience command to build and run with the daemon available automatically.

* **Bug Fixes**
  * Improved robustness when optional paths or referenced loadouts/patch sources are missing, with clearer warnings.
  * In loadouts, inherited variables now safely drop when the target value is unavailable.

### DIFF
--- a/crates/minimal/src/dirs.rs
+++ b/crates/minimal/src/dirs.rs
@@ -1,0 +1,423 @@
+//! `minimal dirs` — print the important directories and files a user
+//! might want to inspect when debugging.
+
+use std::path::PathBuf;
+
+use crate::GlobalArgs;
+use crate::config::resolve_minimal_config_dir;
+use crate::mesh_enrolment_path;
+
+/// Print the important directories and files a user might want to
+/// inspect when debugging. Grouped by role (config / state / cache)
+/// with an "exists?" marker so unused paths don't look identical to
+/// misconfigured ones.
+pub fn cmd_dirs(global: &GlobalArgs) -> Result<(), anyhow::Error> {
+    // Shown in the "Daemon logs" note to point users at today's
+    // rolling log file directly, e.g. `minimald.log.2026-07-08`.
+    let today = chrono::Local::now().format("%Y-%m-%d").to_string();
+    let dirs = DirsLookup {
+        config: resolve_minimal_config_dir(global),
+        mesh_enrolment: mesh_enrolment_path(global),
+        // `mesh_enrolment_path` follows `--minimal-dir` when set,
+        // so the row groups with State in that case rather than
+        // Config — otherwise a user reading `minimal dirs` would
+        // find the mesh file under the wrong header.
+        mesh_group: if global.minimal_dir.is_some() {
+            MeshGroup::State
+        } else {
+            MeshGroup::Config
+        },
+        state: match &global.minimal_dir {
+            Some(dir) => dir.clone(),
+            None => paths::minimal_state_dir()
+                .as_utf8_path()
+                .as_std_path()
+                .to_path_buf(),
+        },
+        cache: paths::minimal_cache_dir()
+            .as_utf8_path()
+            .as_std_path()
+            .to_path_buf(),
+        today,
+    };
+    print_dir_rows(&build_dir_rows(&dirs));
+    Ok(())
+}
+
+/// Resolved locations `cmd_dirs` prints, materialized once so
+/// [`build_dir_rows`] stays a pure function of its inputs (easy to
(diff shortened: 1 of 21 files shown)


crates/minimal/src/dirs.rs
```diff
@@ -0,0 +1,423 @@
...
+
+use crate::GlobalArgs;
+use crate::config::resolve_minimal_config_dir;
+use crate::mesh_enrolment_path;
+
+/// Print the important directories and files a user might want to
+/// inspect when debugging. Grouped by role (config / state / cache)
+/// with an "exists?" marker so unused paths don't look identical to
+/// misconfigured ones.
+pub fn cmd_dirs(global: &GlobalArgs) -> Result<(), anyhow::Error> {
```
twitchyliquid64, line 14: Do we want minimald/minvmd stuff in here?

twitchyliquid64: Eg $STATE/minimal/providers/local-0/{ssh.sock,minvmd stuff, data-vol.raw} etc

crates/sessions/src/client/mod.rs
```diff
@@ -1,2 +1,4 @@
 pub mod composer;
```
twitchyliquid64, line 1: Up to you but this might be a better candidate as a top-level crate, i.e. clientconf or clientstate or something

twitchyliquid64 (approved):
A few nits but up to you!

crates/minimal/src/dirs.rs
evanspearman: I think we do actually because the client is the main way that users will interact with minimal. That said, it only really makes sense if the daemon they are connected to is running locally. When we add support for connecting to remote daemons we should revisit this I think.

crates/sessions/src/client/mod.rs
evanspearman: I think that's reasonable, but I'll do it in a separate PR.

PR #702, opened 2026-07-10
ci: adopt nextest profiles; reshape the KVM lane to its end state

norrietaylor:
## What

Two halves, one dependency chain (the KVM reshape consumes the nextest config):

### 1. `.config/nextest.toml` + `--profile ci` in core-tests

- **`default`** (local runs): surface slow tests at 60s, never kill.
- **`ci`** (core-tests composite → `ci.yml test`, `ci-linux-native tests`, `release.yml test`): `fail-fast = false`, hung tests hard-killed at 5 min, leak visibility, `--no-tests=fail` so an empty selection fails loudly.
- **`vm`** (the KVM lane's archived harnesses): one VM at a time, boots get a generous 120s slow period but a hung boot dies at 6 minutes instead of burning the 30-minute job timeout, longer leak window for the vmm children.
- **No retries anywhere**: a same-invocation retry of a failed VM boot runs against the leaked `__krun-vmm`/gvproxy of the previous attempt ([#588](https://github.com/gominimal/minimal/issues/588)); reap + `terminate-after` remain the mitigation. And the workspace suite has no recorded flakes to paper over.

### 2. KVM lane end shape (`ci-linux-kvm.yml`)

- **nextest archive replaces `testbins.json` + `run-testbin.sh`** (deleted): `build-linux` ships `nextest-archive.tar.zst` in the testbed; `test-kvm` — still toolchain-free — selects harnesses with filtersets under `--profile vm`.
- **The unified session e2e joins the lane** (`scripts/session-e2e.sh`, the same proof the native and macOS lanes run), covering Deployment Model 3: `minimal --minvmd activate` auto-spawns minvmd, boots the microVM, round-trips a session exec over the vsock bridge, destroy/stop. The CLI is built in `build-linux` as a **separate cargo invocation** (a combined build feature-unifies libkrun into the CLI — the regression class the macOS lane hit) and ships in the testbed; a `sudo` reap precedes it (relay leftovers are root-owned); `cli-e2e-boot.log` joins the log artifact.
- **Daemon lifecycle blob → `scripts/lifecycle-e2e.sh`**: PATH-resolved minvmd (no cargo, macOS-reusable), temp workdir, EXIT-trap teardown so a failed assert can't strand a daemon.
- **`krun_add_disk3` export assert moves into `setup-libkrun-linux`**, mirroring the macOS composite and closing a gap: release.yml uses the composite and previously had no check. Early actionable failure; minvmd's compile-time link is the real backstop.

[...]

Issue #687, opened 2026-07-09
feature: CI refactor — clear build/test phases, three gating targets


norrietaylor:
## Summary

Comprehensive CI refactor for gominimal/minimal to eliminate redundant rebuilds, separate build from test phases across lanes, and gate merges on all three production platforms (macOS/HVF, Linux native minimald, Linux/KVM). The work is sequenced as 10 independently landable PRs (PR0–PR9) plus an admin ruleset flip.

## Problem

The current CI has structural problems: test jobs rebuild artifacts instead of consuming pre-built ones; monolithic jobs mix setup, build, and e2e; shared steps are duplicated across workflows and `release.yml` copies rather than reuses CI logic; only lint and unit tests gate merges — no VM boot blocks a bad merge; the `ci-docs-skip.yml` inverse-path hack creates wedge and silent-ungate failure modes; and the Mac mini's time is spent on unit/build work that could run on GH-hosted runners.

## Desired outcome

- One build per platform per run; test jobs consume pre-built artifacts
- Every lane reads as clear build → test phases; no monolithic jobs
- Shared steps (Rust setup, libkrun, guest artifacts, linkage rewrite) in one composite/script; `release.yml` reuses CI's setup
- Three required gating targets: `ci-macos-success`, `ci-linux-native-success`, `ci-linux-kvm-success`
- Robust required-check mechanics via changes-filter jobs + `if: always()` aggregators; `ci-docs-skip.yml` deleted
- Coverage gaps closed: native minimald daemon e2e, `crates/minimald/**` triggering the mac lane, release artifacts smoke-tested before staging
- Mac mini handles only hypervisor e2e; unit/build moves to GH-hosted runners
- Dev scaffolding removed: boop timeouts, split arm64 builds, dead conditionals, redundant verify workflows

## Scope notes

Work is sequenced: PR0 (done) → PR1 (in review) → PR2–PR9 + Admin ruleset flip; each PR soaks before the ruleset flip makes lanes required. Key files: `.github/workflows/ci.yml`, `ci-macos.yml`, `ci-linux-kvm.yml`, `ci-shell-installer.yml`, `ci-docs-skip.yml` (delete), `ci-netns.yml` (delete), `ci-gvproxy.yml` (delete), `release.yml`, and new `ci-linux-native.yml`. Networking test surface stays mothballed by decision; arm64-Linux binaries remain untested by decision.

---

This issue is the tracking issue for this work. A maintainer with write
[...]

PR #703, opened 2026-07-10
ci: cache discipline, slim CI debug info, and a --locked sweep

norrietaylor:
Stacked on [#702](https://github.com/gominimal/minimal/pull/702) (base = its branch; GitHub retargets to main when it merges — the diff here is N4 only).

## What

Three cross-lane changes sized to fork every cache exactly once:

1. **rust-cache discipline** — `setup-rust` gains `shared-key` + `save-if` inputs (defaults: `workspace`, main-only writes), forwarded from `core-tests` (`tests`). Per-class keys: `clippy`, `native-e2e` (`-p` builds unify features differently than the workspace suite), `mac-unit` (inline). **PR branches restore but never write** — PR churn stops LRU-evicting the main-branch caches from the shared 10 GB pool (the 13m42s cold compile on #702's first run was this problem live).
2. **Restore/save splits** for the two multi-GB raw caches (KVM `build-linux`, macOS `artifacts` stage2): `actions/cache/restore` + main-only `actions/cache/save` at job end, guarded on `cache-hit`.
3. **`CARGO_PROFILE_DEV_DEBUG=line-tables-only`** in CI (setup-rust env step; explicit `env:` on the mac/KVM jobs that bypass the composite) — usable backtraces, smaller target trees and caches, faster links. Deliberately not a named `[profile.ci]`: that would move output out of `target/debug` and break codesign/testbed/justfile paths. Plus **`--locked` on every workflow cargo invocation** — a stale Cargo.lock now fails loudly; previously only `cargo fetch` enforced it.

Part of the CI end-state migration ([#687](https://github.com/gominimal/minimal/issues/687)).

## One-time landing cost + prime plan

All keys fork cold at merge (new shared-key + debug-info change together, by design). After merge: `workflow_dispatch` each lane once on main to write the new keys, then a trivial PR must show restore-from-main hits and no `Saved cache` lines. The mini's persistent target dir rebuilds once (new fingerprints from the debug env).

Release.yml caches are deliberately untouched — they inherit these conventions in the release overhaul PR.

## Verification

- This PR's own runs prove the plumbing (inputs resolve, caches restore, `--locked` passes everywhere).
- Post-merge cache behavior: `gh api repos/gominimal/minimal/actions/caches` should trend down over the week; PR runs show no cache writes.

## Summary by CodeRabbit

- **CI Improvements**
  - Enforced `--locked` for Rust builds, tests, doctests, and nextest runs to ensure consistent dependency resolution.
  - Added a `shared-key` option for the core test action and introduced lane-specific cache keys for different CI jobs.
  - Updated caching to restore-only during most runs, saving cache entries only on the main branch when appropriate.
  - Reduced CI debug output (`line-tables-only`) to slim build artifacts.
  - Ensured archiving uses the locked dependency set as well.


norrietaylor:
@coderabbit review

norrietaylor:
Stacked on #702 per owner preference (base = `ci/nextest-profile`), so the diff shows only the cache-discipline delta. Note: lane workflows trigger only on PRs targeting `main`, so checks here stay empty until #702 merges and GitHub retargets this PR — I'll kick a run then; the plumbing gets its live proof at that point.

norrietaylor:
@CodeRabbit review

norrietaylor:
@CodeRabbit review

norrietaylor:
@CodeRabbit review

twitchyliquid64 (approved):

Issue #583, opened 2026-06-26
feature: pkg cache on the in-VM ext4 volume (volume-aware); VM path on Mac (HVF) and Linux (KVM)


norrietaylor:
> **July 29 sprint — committed item** (agreed Norrie/Tom, Jul 1 F2F). Owner: @norrietaylor.
> *Rewritten 2026-07-06: session workbenches (`/workbench`) move onto the writable volume; disk topology reworked; multi-minvmd forward-compat added; staging-location/OUTPUT_DIR claims corrected against code.*
> *Rewritten 2026-07-07: **committed scope narrowed to RAW two-disk (Phase 1); host-FS reflink CoW (Phase 2) and qcow2 overlays are both deferred.** FFI/library claims corrected against libkrun 1.19.0 and the code as it stands on `main` (no `Qcow2` `DiskFormat` variant exists; no host-side overlay-creation path exists; reflink verified on APFS).*

## Summary

The pkg cache must sit on the **same filesystem as the sandbox rootfs-staging trees**, because sandbox composition **hardlinks** out of the cache and hardlinks cannot cross devices (`EXDEV`). The same writable volume also hosts the **session workbenches** (the `/workbench` worktrees) and the rest of `minimal_state_dir` — for capacity and persistence, not EXDEV (see below).

This spec covers the **VM-backed model only** — the unified default on both platforms (libkrun): Mac via Apple HVF, Linux via KVM. The writable volume is **guest-only**: never mounted or served from the host. Project files cross the boundary by **workspace upload or git download**, not a host filesystem mount.

## Root cause

Building sandboxes hardlinks from the cache into the sandbox **rootfs staging** trees — `<state>/tasks/` for session/task envs (`crates/minimald/src/env.rs:282-284`) and `<state>/sandboxes/` for package builds (`crates/op/src/specs.rs:284`). The hardlink primitive has **no cross-device fallback**: `common::hardlink_dir_contents` tolerates only `AlreadyExists`, and returns every other error (including `EXDEV`) as `HardlinkFailed` (`crates/common/src/lib.rs:209-224`). So cache ⟷ rootfs-staging must be one filesystem.

Two corrections retained from the earlier text:

- Staging is **not** under `/tmp`, and hakoniwa does not stage the rootfs. Both staging trees live under `minimal_state_dir`. hakoniwa's `/tmp` TempDir is only a pivot_root bind-mount point; it merely has to stay writable.
[...]

PR #705, opened 2026-07-10
feat(minvmd,minimald): quiesce data volume on stop and gate READY on mount

norrietaylor:
Since [#672](https://github.com/gominimal/minimal/pull/672), session worktrees and the package cache persist on the per-VM `/dev/vdb` volume — user data with no host copy. This PR fixes the failure modes that made that persistence unsafe: stopping a VM corrupted-by-default, and a broken volume failed silently.

## Failure modes fixed

**1. Every `minvmd stop` was an unclean unmount — dirty journal, possible data loss.**
`stop` sent SIGTERM straight to the VMM with nothing syncing or unmounting the volume. Every boot after a stop replayed the ext4 journal ("recovering journal" on every startup), and with `sync_mode=relaxed` any writes from the final seconds before the kill — a `git commit` in a session, a build artifact — could vanish entirely.
*Now:* `stop` asks the in-VM minimald to shut down over the bridge UDS before signalling: sessions drain, the volume gets `syncfs` + unmount (falling back to remount-ro, which closes the journal even with open fds), and only then does SIGTERM fire. A clean stop leaves a cleanly-unmounted filesystem; SIGTERM/SIGKILL escalation is unchanged when the guest doesn't answer.

**2. A VM whose volume failed to mount reported READY anyway — a ghost VM that eats data.**
A mount failure was only logged; the guest silently kept its state on the boot tmpfs and signalled READY. Sessions appeared to work, wrote everything to RAM, and lost it all at the next stop — with zero indication anything was wrong. An undersized or corrupt volume image produced a "healthy" VM indistinguishable from a real one.
*Now:* the guest never substitutes tmpfs. On any mount failure it emits a distinct `MOUNT_FAILED\n<reason>\n` beacon instead of READY, and `minvmd boot`/`run` fail with a non-zero exit and the guest's actual failure reason.

**3. A failing volume offered no repair path and no safe disposition.**
There was no distinction between "this image holds your sessions and won't mount" and "this blank image was just provisioned wrong" — and no repair attempt at all.
*Now:* a formatted volume that fails to mount gets one `e2fsck -p` repair + mount retry before failing closed (never a reformat once a superblock exists). On failure, the host keys off whether the image pre-existed the boot: a pre-existing image may hold session data, so it is left untouched and the error points at repair or `MINVMD_DATA_VOLUME_PATH`; a blank image freshly created this boot is removed so the next boot isn't misdiagnosed as data-at-risk.

## Found along the way

Two lifetime-held write fds on the volume made every unmount/remount fail `EBUSY`, keeping the journal dirty even after a quiesce — the shutdown sequence only became verifiable once both were gone:

- minimald's lcache read tracker held a locked `alog/<n>.v1` open for the daemon's lifetime; the sessions manager now releases it during shutdown (`Cache::release_read_tracker`).
- [#690](https://github.com/gominimal/minimal/pull/690)'s `minimald.lock` lives in the provider dir, which in-guest sits on the volume; the vsock-mode instance lock now lives on the `/run` tmpfs (host-side placement unchanged — nothing outside the boot reads the guest's lock).

The quiesce EBUSY warning now lists the offending fd holders via a `/proc` scan, so the next accidental pin is diagnosable from the console log.

## Verification

New gated e2e suite (`MINVMD_E2E=1`, verified on macOS/HVF), asserting against the raw image bytes — no host ext4 tooling needed:

- clean stop → superblock marked cleanly unmounted (`EXT4_VALID_FS` set, `INCOMPAT_RECOVER` clear);
- unformattable fresh volume (below the mkfs floor) → no `vm-up`, non-zero exit, blank image removed;
- unmountable pre-existing volume (valid magic, corrupt fs) → no `vm-up`, non-zero exit, image preserved.

VM-booting tests (including the pre-existing `boot_e2e`) now also isolate `HOME`, so `dirs`-based path fallbacks can never escape onto a developer's real state dir.

Refs: [#583](https://github.com/gominimal/minimal/issues/583), builds on [#613](https://github.com/gominimal/minimal/pull/613)'s Shutdown RPC

## Summary by CodeRabbit

* **New Features**
  * Added more detailed VM boot failure reporting, including `MOUNT_FAILED` reasons, and better run diagnostics via `run.log`.
  * Added state-volume quiescing on shutdown (sync + unmount/detach) and fail-closed mounting with automatic repair for existing ext4.
  * Improved cache read-tracker release during daemon shutdown/cleanup.
* **Bug Fixes**
  * Prevented “ready” outcomes when storage mounting fails.
  * Ensured freshly provisioned failed volume images are discarded, while pre-existing images are preserved for investigation.
* **Tests**
  * Added/extended e2e coverage for volume quiescing and mount-failure handling.


norrietaylor:
Rebased onto latest main (over [#690](https://github.com/gominimal/minimal/pull/690)'s socket/path rework). Two things beyond mechanical conflict resolution:

- **[#690](https://github.com/gominimal/minimal/pull/690)'s lifetime `minimald.lock` reintroduced the dirty-journal bug this PR fixes**: in-guest, the provider dir lives on the data volume, so the lock's lifetime-held write fd made the quiesce unmount/remount-ro fail EBUSY on every clean stop (the new fd-holders diagnostic caught it: `pid 1 fd /var/lib/minimal/providers/local-0/minimald.lock`). Fixed by placing the *in-guest* (vsock-mode) instance lock on the `/run` tmpfs — nothing outside the boot reads the guest's lock, and boot-ephemeral coordination state doesn't belong on durable storage. Host-side (DM2) lock placement is unchanged.
- The e2e tests adopted [#690](https://github.com/gominimal/minimal/pull/690)'s conventions: short `/tmp` state dirs (the bridge UDS now lives inside the state dir and must fit `sun_path`), the `MINVMD_BIN` override, and the new `providers/local-0/` locations for the volume image.

All proofs re-verified on macOS/HVF after the rebase: clean stop → `EXT4_VALID_FS` set / `INCOMPAT_RECOVER` clear; both mount-failure modes fail loudly with correct image disposition.

twitchyliquid64 (approved):
"quiesce" lol

PR #721, opened 2026-07-11
fix: decouple mctx from hakoniwa for macOS build-system commands

0chroma:
## Summary

`mctx` had an unconditional dependency on `hakoniwa` → `libcgroups` → `procfs` (Linux-only), which forced `minimal init`/`add`/`update` to be `#[cfg(target_os = "linux")]`. This unblocks those commands on macOS.

### Background

Per discussion with @norrietaylor / @msample / @twitchyliquid64: the dep chain was accidental. `init` doesn't use the sandbox — it only needs VCS + graph + mfile parsing. `mctx` used `hakoniwa` directly in exactly one place: the return type of `Context::command` (an execution method `init` never calls). `sandbox2` already gated `hakoniwa` in its `Cargo.toml` to Linux, but its source referenced `hakoniwa` unconditionally.

### Options considered

**Option A: Gate `hakoniwa` directly in `mctx/Cargo.toml`**
Move `hakoniwa.workspace = true` into a `target.cfg(target_os = "linux")` section, `#[cfg]` the one `hakoniwa::Command` return in `env.rs`. ~5 lines.

**Option B: Re-export from `sandbox2` (chosen)**
Remove `mctx` → `hakoniwa` dependency entirely. `sandbox2` re-exports `hakoniwa::Command` as `sandbox2::Command`. Both `sandbox2` and `mctx` gate their hakoniwa-touching code to Linux.

Went with **B** because `sandbox2` is already the single source of truth for the sandbox. Having `mctx` depend on `hakoniwa` directly creates two dependency boundaries to maintain; re-exporting through `sandbox2` keeps one.

### Changes

| Crate | Change |
|-------|--------|
| `sandbox2` | `#[cfg(target_os = "linux")]` on `Container`, execution methods (`new_container`, `command`, `run`, `run_with_cancel`), `bind_mount`, `locked_mount_flags`, `SpawnFailed` error variant; re-export `hakoniwa::Command` |
| `mctx` | Remove direct `hakoniwa` dep; use `sandbox2::Command`; `#[cfg]` the `container`/`command` methods and `SpawnFailed` error conversion |
| `minimal` | Remove `#[cfg(target_os = "linux")]` from `init`/`add`/`update` commands and arg structs; make `mctx`/`op` unconditional deps |

### Verification

- `cargo test -p minimal -p sandbox2` — all tests pass
- `cargo clippy -p minimal -p sandbox2 -p mctx --all-targets -- -D warnings` — clean
- `cargo fmt -- --check` — clean
- `cargo tree -p minimal --target aarch64-apple-darwin` — no `hakoniwa`/`libcgroups`/`procfs` in macOS dep tree
- `cargo tree -p mctx --target aarch64-apple-darwin` — same, clean

Follow-up to the discussion on gominimal/inbox#159.

## Summary by CodeRabbit

* **New Features**
  * Build-system CLI subcommands (`init`, `add`, `update`) are now built and available on all supported platforms.

* **Bug Fixes**
  * Sandbox/container execution is now enforced as Linux-only; on non-Linux targets, attempts return a clear “sandbox unsupported” error instead of trying to run.
  * Improved cross-platform error handling and sandbox execution flow for spec builds and standalone tests.
  * Updated cache directory replacement to use Linux rename-exchange when available, with safer non-Linux fallbacks.


twitchyliquid64 (approved):
Great! CI shows theres a couple more to flag but then this should be g2g!

PR #731, opened 2026-07-13
feat(sessions): prompt user when shell process exits

twitchyliquid64:
- New crate `async-dialog` to implement ANSI terminal prompt, such as down an SSH connection.
 - `WeakManagerHandle` counterpart to `ManagerHandle`, used to avoid circular references
 - Prompt for when the shell process exits
 - Tests for teardown when the user selects deletion in the prompt

## Summary by CodeRabbit

* **New Features**
  * Added an async terminal dialog library with keyboard-driven selection menus (prompt support, default option, width truncation, and confirm/cancel controls).
  * Added a shell-exit prompt that lets users choose to detach or delete the session, with deletion performed through a manager-mediated control path.
* **Bug Fixes**
  * Improved session teardown behavior when shell connections end, including reliable cleanup after choosing “delete”.
  * Restored terminal rendering state cleanly after interactive prompts complete.
* **Tests**
  * Updated/extended session prompt interaction tests to validate delete/detach outcomes.

### DIFF
--- a/crates/minimald/src/session_host.rs
+++ b/crates/minimald/src/session_host.rs
@@ -5,6 +5,7 @@
 //!
 //! The [`Host`] struct holds the running state of an active session.
 
+use async_dialog::Selection;
 use either::Either;
 use russh::Channel;
 use russh::server::Msg;
@@ -25,6 +26,7 @@ use tokio::task::JoinHandle;
 
 use crate::RequestedPty;
 use crate::session::SessionPaths;
+use crate::sessions::SessionControl;
 #[cfg(not(test))]
 use sessions::NetworkMode;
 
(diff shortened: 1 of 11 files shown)


crates/minimald/src/session_host.rs
```diff
@@ -402,27 +420,69 @@ impl Binding {
...
+        if exit_reason == MainloopExitReason::StdoutErr {
+            // The shell process exited. For a bash shell, this usually meant someone pressed ctrl-d absent-mindedly.
+            // We presume they didnt want to completely destroy the session, perhaps just detach, but lets prompt
+            // to see where they wanted to go from here.
+            let _ = w.write_all(b"\r\n").await;
+            match async_dialog::Select::new()
+                .with_prompt(SHELL_EXIT_PROMPT)
+                .items([
+                    "Detach, leaving the session running",
+                    "Delete the session (all files permanently deleted)",
```
msample: it would be clearer if the Delete option's aside was "all in-session files permanently deleted"

twitchyliquid64: Done.

msample (approved):

PR #733, opened 2026-07-13
feat(sessions): load and apply user policy

evanspearman:
Load `user_policy.toml` from `<config>/minimal/` and thread it through
activation. Fills in the placeholder `UserPolicy::empty()` /
`UserPolicy::default()` sites that were shipping unenforced.

- `sessions::core::policy` — `read_user_policy_file` / `_or_default`
  loaders (empty policy on `NotFound`).
- `minimal::config::read_user_policy` — helper mirroring
  `read_client_config`. Loaded once in `cmd_activate`, passed into
  both `compose_user_contribution` and `drive_pending_to_active`.
- `drive_pending_to_active` now takes `UserPolicy` + `ComposeOptions`
  (was defaults) — daemon-response pending items get gated against
  the real policy.

## Policy semantics fixes

Two behavior corrections uncovered while wiring this up:

- **Deny beats ignore** on single-path check in both `VarsPolicy` and
  `ExpandedPatchPolicy`. Deny is the emergency stop — it can't be
  silently hidden behind an overlapping ignore glob. `PathDecision::combine`
  was already correct; single-path `decide` / `check` were inverted.
- **Policy patterns don't need to be absolute.** Split
  `expand_source` (patch-source, must-be-absolute walker input) from
  `expand_policy_pattern` (matcher against real paths). `deny =
  ["**/*.pem"]` in `user_policy.toml` now works. `~name/` tilde-user
  form still rejected loudly (was previously caught by the absolute
  check; now has its own `ExpandError::UnsupportedTildeUser` variant
  so both code paths surface the footgun).

## Summary by CodeRabbit

* **New Features**
  * Added support for loading user policy rules from `user_policy.toml`, with missing files defaulting to an empty policy.
  * User-defined variable and patch rules are now used during session composition/activation.
  * Policy pattern matching now supports non-absolute globs while still expanding home-directory paths.
* **Bug Fixes**
  * `deny` rules now consistently take precedence over `ignore` for both variables and patch matching, including overlapping rules.
  * Per-user tilde patterns (for example `~name/...`) are rejected earlier with a clearer error.

### DIFF
--- a/crates/minimal/src/config.rs
+++ b/crates/minimal/src/config.rs
@@ -30,3 +30,15 @@ pub fn read_client_config(
     let cfg_path = resolve_minimal_config_dir(global).join("config.toml");
     sessions::client::config::read_config_or_default(&cfg_path).map_err(|e| anyhow::anyhow!("{e}"))
 }
+
+/// Read the user policy from `<config>/minimal/user_policy.toml`.
+/// A missing file returns [`UserPolicy::empty`] — a fresh install
+/// activates fine without needing the file to exist.
+///
+/// [`UserPolicy::empty`]: sessions::core::policy::UserPolicy::empty
+pub fn read_user_policy(
+    global: &GlobalArgs,
+) -> Result<sessions::core::policy::UserPolicy, anyhow::Error> {
+    let path = resolve_minimal_config_dir(global).join("user_policy.toml");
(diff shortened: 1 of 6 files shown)


twitchyliquid64 (approved):
`cargo fmt && cargo clippy --all-targets --fix --allow-dirty -- -D warnings && cargo fmt`

PR #737, opened 2026-07-13
refactor(minimal): name the CLI binary target `min`

norrietaylor:
The user-facing command has been `min` for a while — clap says so (`#[command(name = "min")]`), and the installer lays the binary down as `bin/min`. But cargo still emitted `minimal`, and `min` was conjured by a **rename at release-staging time**. The dev build therefore disagreed with what ships: `just up` gave you `target/debug/minimal` while every doc, and every user, typed `min`.

This declares the binary target explicitly:

```toml
[[bin]]
name = "min"
path = "src/main.rs"
```

Declaring a `[[bin]]` disables autodiscovery, so `src/main.rs` now builds **only** as `min`. The `minimal` lib target and the package name are unchanged, so `cargo build -p minimal` still works.

## Callers updated

`--bin minimal` → `--bin min` and the built-binary paths across all five workflows (`ci-linux-kvm`, `ci-linux-native`, `ci-macos`, `nightly-tests`, `release`), the KVM lane's testbed staging that copies the CLI onto `PATH`, `session-e2e.sh`'s `mnl()` wrapper, and the justfile's binary variable (`minimal` → `min-bin`).

## What deliberately did NOT change

**The published release asset names** (`minimal-linux-amd64`, `minimal-macos-arm64`, …). Those are the contract that the install manifest and `scripts/install.sh` are built around; renaming them would churn the release channel for no user-visible gain, since the manifest already maps them to `bin/min`. Only the local build-output path feeding those steps moved, so `release.yml` now reads `mv target/release/min target/release/minimal-macos-arm64`.

**Two other `minimal` names that are not this binary**: the macOS shim (`~/.minimal/shim/bin/minimal`, a script that runs the CLI *inside a VM* — it's what `just artifacts` invokes to materialize the guest kernel), and the `minimal` crate/lib target.

## Verification

Everything that can be run locally, was:

- `cargo build -p minimal` emits `target/debug/min` and no longer emits `minimal`; `cargo metadata` confirms the targets are `[('minimal', lib), ('min', bin)]`. `min --version` / `--help` work.
- **Full session e2e on the macOS/HVF VM lane** — `E2E_VM=1 ./scripts/session-e2e.sh`, which resolves `min` from `PATH` and autospawns `minvmd` exactly as CI does. Cold activate → list → warm call → destroy → stop, all green. This is the same script the Linux native and KVM lanes run, so it exercises the `PATH`-based resolution those lanes depend on.
- **Release path simulated**: `cargo build --release -p minimal --bin min` succeeds, and the `otool` gate `release.yml` runs against it passes (links only `/usr/lib` + `/System`, no libkrun — the `default-features = false` opt-out still holds).
- All five workflows parse as YAML; `shellcheck scripts/session-e2e.sh` is clean; `just --list` / `just --evaluate min-bin` resolve; the installer harness is 148/148 green (its `bin/min` contract is untouched).
- No Rust code referenced the binary by name — `crates/minimal/tests/cli.rs` calls library functions, and only the minvmd tests use `CARGO_BIN_EXE_*` (for `minvmd`, unchanged).

## Follow-up, not done here

`docs/reference/cli.md` still writes `minimal run shell` etc., and `release.yml` names the generated completion files `bash/minimal`, `zsh/_minimal`, `fish/minimal.fish`. Both are pre-existing inconsistencies — the completion *contents* already register for `min`, since clap's command name was already `min` — and neither affects the build. Worth a separate docs pass.

## Summary by CodeRabbit

- **New Features**
  - The command-line interface is now provided as `min` instead of `minimal`.
  - Development, session, and end-to-end workflows now use the `min` command consistently.
  - Release builds and downloadable artifacts are aligned with the updated CLI naming.
- **Documentation**
  - Updated setup guidance, examples, comments, and troubleshooting messages to use commands such as `min activate` and `min ls`.
- **Bug Fixes**
  - Improved consistency between the CLI name used during development, testing, and release packaging.


twitchyliquid64 (approved):

twitchyliquid64:
Lets kick off an unstable after this also to see if we broke anything in the release

PR #748, opened 2026-07-14
feat(minimal): upload project files to session on activate

### LINKED ISSUE #767: feature: File sync INTO a Session
- Copying files on spin-up
- Dirty file notification on exit ([mock](https://app.notion.com/p/Proposal-Minimal-as-a-Session-Manager-360040938a8981cc9e01f675b40b071d?source=copy_link#5b663a049f3f4e9fa94abdc9b0bd6151))
  - Confirmation prompt: do you want to exit the session with uncommitted work? Disconnect (detach session) / Exit (kill the session, changes are lost) / Restart (restart the session, keep the changes)
- Mechanism: async-tar, rsync, etc.

# Spec
Currently when a user uses "minimal run" (original CLI), files in the context of the invocation CWD and the user are made available in the Minimal sandbox
- the CWD or its first ancestor directory that has a minimal.toml file or .minimal/minimal.toml file and all its contents (descendants to leaf directories and files)
- files that are explicitly mapped in from the minimal.toml task spec
- the "state dir" (~/.cache/minimal/state/<state-key>)
- files explicitly "patched" (mapped) by any of the packages referenced (note that doing a "min add" in the sandbox currently fails pull in these files but the minimal2 could solve for this with a backchannel to the client if the "min" client in the sandbox is the full min CLI talking to the sandbox parent's minimald)

The Minimal One CLI client "min" (replacing minimal CLI) can skip handling "state dir" as that will be handled by the minimald as cached state over many sandbox runs using that state_key (and the the user is not supposed to be able to initialize/update its contents except when in the sandbox).

All of the others must be uploaded to minmald such that they are available to unpack into the filesystem exposed to the sandbox.  User identity and access permissions must work as expected

The files uploaded to mimimald must not be lost if the system(s) hosting the minimald and its sandbox(es) are powered off or shut down (e.g. use durable Ext4 fs if minimald is running in a VM).

These files can only be deleted when the Session is explicitly destroyed or the non-interactive task completes.

This spec is for file upload only. For now synchronizing changed or new files out of the sandbox will be handled by the human using tools like git push.

0chroma:
## Summary

After `CreateSession` succeeds, the client now tar+zstd's the project directory and streams it to the daemon's `WorkspaceFilesTarZst` subsystem, which unpacks it into the session workspace. This is the file-sync-INTO direction from gominimal/minimal#767.

### Background

The daemon-side receiver already existed (`rpc.rs:404` — `WorkspaceFilesTarZst` subsystem), but the `minimal` CLI had no client-side sender. `cmd_activate` sent only `SessionConfig` + `WireContribution` over RPC — no file payload. The daemon read `project_path` directly on the host (only works when co-located).

### Changes

- **`client.rs`** — new `upload_workspace_files` method: opens a channel, sets `MINIMAL_SESSION_ID`, requests the `WorkspaceFilesTarZst` subsystem, streams a zstd-compressed tarball, and drains the response
- **`file_upload.rs`** — new module: `tar_directory` recursively walks a directory and builds an in-memory tar archive preserving relative paths, symlinks, and empty directories
- **`lib.rs`** — `cmd_activate` calls `upload_workspace_files` after session creation, before the attach prompt
- **`Cargo.toml`** — added `async-tar`, `async-compression`, `constcat` deps

### Open questions

- **Ignore/exclude files**: The upload currently sends the entire project directory including `.git/`, `node_modules/`, build artifacts, etc. Should we support `.gitignore` parsing, a hardcoded default exclude list (`.git`, `node_modules`, `target`, `__pycache__`, `.venv`), a `.minimalignore` file, or some combination? Tracked as an open question on gominimal/minimal#767.

Closes gominimal/minimal#767 (file sync INTO a session).

## Summary by CodeRabbit

* **New Features**
  * When activating, the full project workspace is now uploaded to the daemon as a zstd-compressed tar archive.
  * Upload preserves directories (including empty ones), regular files, and symbolic links, while excluding common heavy defaults like `.git`, `target`, and `node_modules`.
* **Bug Fixes**
  * Workspace upload failures now report clearer error details.
  * Unreadable or permission-restricted paths are skipped to prevent activation from failing.
* **Tests**
  * Added an integration test that activates a project and verifies uploaded contents (including nested files) via SFTP.

### DIFF
--- a/crates/minimal/src/file_upload.rs
+++ b/crates/minimal/src/file_upload.rs
@@ -29,6 +29,21 @@ async fn add_dir_entries(
     add_dir_entries_inner(tar, root, prefix).await
 }
 
+/// Directories always skipped during upload. A proper .gitignore
+/// implementation is tracked as a follow-up on #263.
+const DEFAULT_EXCLUDED_DIRS: &[&str] = &[".git", "target", "node_modules"];
+
+fn is_default_excluded(name: &str) -> bool {
+    DEFAULT_EXCLUDED_DIRS.contains(&name)
+}
+
+async fn file_type_will_be_dir(entry: &tokio::fs::DirEntry, entry_path: &std::path::Path) -> bool {
+    match entry.file_type().await {
+        Ok(t) => t.is_dir(),
+        Err(_) => entry_path.is_dir(),
+    }
+}
+
 fn add_dir_entries_inner<'a>(
     tar: &'a mut Builder<Vec<u8>>,
     root: &'a Path,
@@ -62,6 +77,13 @@ fn add_dir_entries_inner<'a>(
             let name = entry.file_name();
             let name_str = name.to_string_lossy();
             let entry_path = entry.path();
+
+            // Skip common heavy/irrelevant directories. A proper .gitignore
+            // implementation is tracked as a follow-up on #263.
+            if file_type_will_be_dir(&entry, &entry_path).await && is_default_excluded(&name_str) {
+                continue;
+            }
+
             let archive_path = if prefix.is_empty() {
                 name_str.to_string()
             } else {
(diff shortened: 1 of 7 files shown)


evanspearman (approved):

crates/minimal/src/file_upload.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use anyhow::Context as _;
+use async_tar::Builder;
+use tokio::fs;
+
+/// Builds an in-memory tar archive of `dir`, preserving relative paths.
+///
+/// Symlinks are stored as symlinks. Directory entries are included so
+/// empty directories survive the round-trip.
+pub async fn tar_directory(dir: &Path) -> Result<Vec<u8>, anyhow::Error> {
```
twitchyliquid64: This will OOM a lot of my repos, can we use [async_tar](https://docs.rs/async-tar/latest/async_tar/) to stream it instead?

Something like this:

```rust
        let encoder = async_compression::tokio::write::ZstdEncoder::new(some_async_writer);

        let mut tar = async_tar::Builder::new(encoder);
        for _path in files {
            tar.append_dir_all(/* path */).await?;
        }
        let encoder = tar.into_inner().await.unwrap();

        encoder.shutdown().await.unwrap();
        encoder.into_inner() // get the wrapped writer back
```

0chroma:
sorry, I should have probably marked this as a draft, I'm glad I got the early feedback in though! helpful seeing things that maybe I didn't originally consider

crates/minimal/src/file_upload.rs
```diff
@@ -0,0 +1,343 @@
...
+use std::path::Path;
+
+use anyhow::Context as _;
+use async_tar::Builder;
+use tokio::fs;
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+
+/// Directories always skipped during upload. A proper .gitignore
+/// implementation is tracked as a follow-up on #263.
+const DEFAULT_EXCLUDED_DIRS: &[&str] = &[".git", "target", "node_modules"];
```
twitchyliquid64: Nit / optional: I'm not super-keen on excluding `.git` as then my git commands won't work in the session. But happy to figure that out / how the ergonomics feel later.

twitchyliquid64 (approved):

PR #760, opened 2026-07-15
test: prove the in-sandbox min add path via interactive PTY attach on every lane

norrietaylor:
## What

Deepens the shared session e2e (`scripts/session-e2e.sh`) so that, on **every**
deployment lane (native DM2, KVM DM3, macOS DM1), it forks a real session
sandbox and proves the in-sandbox `min` helper end to end — not just the session
lifecycle. Also fixes the CI gap that prevented the VM lanes from running an
in-guest session at all, and improves git-command error reporting.

Three commits.

## `test: exercise the session sandbox via interactive attach`

Attaches an interactive shell to a live session (which forks a real hakoniwa
sandbox — on a VM lane, inside the guest over the vsock bridge) and proves
`min add`: a genuinely-absent tool (`jq` — the `shell` stack composes `base` +
`curl`, and `jq` is pulled in by neither) is recorded absent, `min add`ed, then
run; its version banner round-tripping proves `min add` reached the daemon over
the `/run/minenv_sock` relay and hardlinked the package into the live rootfs.

A session is **interactive by design**, so the proof drives it like a real user
through a **real PTY** (`scripts/e2e-attach-pty.py`), not a pipe: it pumps the
command stream, then answers the session-exit Detach/Delete prompt with
keystrokes (Down + Enter → Delete). A pipe is not a tty and cannot answer that
prompt. Selecting Delete tears the session down, so it must then be delisted —
doubling as the interactive-teardown proof. Output is matched with a bash glob,
not `printf | grep -q`: `min add`'s progress-bar flood can make `grep -q`
short-circuit and SIGPIPE the feeding `printf`, which under `pipefail` becomes a
false negative.

Host-side seed: since [#748](https://github.com/gominimal/minimal/pull/748)
`min activate` uploads the project into the session, so the e2e seeds a small,
**pinned** `minimal.toml` (the repo's `[upstream]` verbatim, incl.
`locked_commit`, plus a light `shell` stack) in a dir it owns and cleans
(`rm -rf` first — a persistent/self-hosted runner can carry a stale *unpinned*
leftover that the graph loader rejects). This also satisfies the
[#758](https://github.com/gominimal/minimal/pull/758) non-interactive `min
activate` pre-flight. Every seed is removed on teardown.

## `ci: provision gvproxy for the KVM/macOS session-e2e lanes`

The VM lanes boot a microVM whose guest daemon clones the upstream `pkgs` repo
during session mint — which needs the guest's egress (NAT + DNS), provided by
the gvproxy switch `minvmd` spawns. Neither VM lane provisioned the gvproxy
binary, and its resolution is best-effort (never errors), so `minvmd` booted the
VM **switchless**: the guest overlay had no other end and every egress died with
`Could not resolve host: github.com`. It looks like DNS but is dead L3 egress —
raw TCP to literal public IPs returns `EHOSTUNREACH`, and even the gateway is
silent.

Fetches the pinned gvproxy (`scripts/fetch-gvproxy.sh` +
`vendor/gvproxy/gvproxy.lock`, already used by the native
`minimald-root-integration` job) and points `minvmd` at it via
`MINVMD_GVPROXY_BIN` on each VM lane's session-e2e step. The macOS lane
auto-selects the `gvproxy-darwin` asset. Native lanes never needed it (`minimald`
runs on the host with real internet; no overlay).

## `fix(checkouts): report exit status on git command failure`

A failed git command surfaced only its stderr, which for a signal-killed process
can be a misleadingly benign line (e.g. a "templates not found" warning that
masks that the process was killed). Includes the rendered `ExitStatus` (exit code
or signal) in `GitCommandFailed` and its `mctx` wrapper.

## Validated

All three e2e lanes green — native, KVM, and macOS: the in-sandbox `min add jq` +
run + interactive Delete-teardown round-trips on each. Rebased onto `main`
(post-[#748](https://github.com/gominimal/minimal/pull/748)); `cargo`/`clippy`
clean via cross; `shellcheck`/`bash -n`/`py_compile`/`actionlint` clean.

## Summary by CodeRabbit

* **New Features**
  * Git command errors now display the process termination status alongside the failed command and error output.
  * Added interactive end-to-end coverage for installing and running tools inside an attached session.
* **Bug Fixes**
  * Improved session teardown verification when users choose to delete a session.
* **Tests**
  * Enhanced Linux and macOS session testing with pinned networking support.
  * Added validation for project setup, tool installation, session lifecycle, and restart behavior.

### DIFF
--- a/.github/workflows/ci-linux-kvm.yml
+++ b/.github/workflows/ci-linux-kvm.yml
@@ -335,17 +335,26 @@ jobs:
           MINVMD_BOOT_LOG: ${{ runner.temp }}/boot-daemon.log
         run: ./scripts/minvmd-lifecycle.sh
 
+      - name: Fetch pinned gvproxy switch binary
+        # minvmd spawns this as the per-host-VM switch that provides the guest's
+        # egress (NAT + DNS). Without it minvmd boots the VM switchless
+        # (gvproxy resolution is best-effort and never errors), so the guest's
+        # overlay has no other end and every egress — including the in-guest
+        # `pkgs` clone during session mint — dies with "Could not resolve host".
+        run: ./scripts/fetch-gvproxy.sh "$RUNNER_TEMP/gvproxy"
+
       - name: Session e2e via the CLI (activate/exec/destroy; autospawns minvmd)
         # The SAME unified proof every target lane runs (scripts/session-e2e.sh)
         # against Deployment Model 3 (native Linux + VM — the target this
         # lane owns; see docs/specs/03-spec-networking): from a clean state
         # `minimal --minvmd activate` must auto-spawn minvmd, boot the microVM,
         # and round-trip a session exec over the vsock bridge — the Linux/KVM
-        # counterpart of the macOS lane's CLI gate. E2E_PROJECT_DIR=/tmp: no
-        # project sync into the guest yet, so activate a path that exists in
-        # the guest image. minimal/minvmd resolve from the testbed on PATH.
+        # counterpart of the macOS lane's CLI gate. Since #748 activate uploads
+        # the project into the guest; E2E_PROJECT_DIR=/tmp still names a guest
+        # path. minimal/minvmd resolve from the testbed on PATH.
         env:
           MINVMD_BOOT_LOG: ${{ runner.temp }}/cli-e2e-boot.log
+          MINVMD_GVPROXY_BIN: ${{ runner.temp }}/gvproxy
         run: E2E_VM=1 E2E_MINIMAL_ARGS=--minvmd E2E_PROJECT_DIR=/tmp ./scripts/session-e2e.sh
 
       - name: Upload guest boot console logs
--- a/scripts/session-e2e.sh
+++ b/scripts/session-e2e.sh
@@ -1,36 +1,59 @@
 #!/usr/bin/env bash
 # The ONE session e2e, invoked by every target lane. Drives the real user
 # path through the `min` CLI — which abstracts where the daemon lives —
-# so the identical proof runs against all three deployment targets:
+# so the IDENTICAL proof runs against all three deployment targets:
 #
 #   Linux native (DM2)   minimald on the host          (no extra env)
 #   Linux KVM    (DM3)   minimald in a minvmd microVM  E2E_VM=1 E2E_MINIMAL_ARGS=--minvmd
 #   macOS HVF    (DM1)   minimald in a minvmd microVM  E2E_VM=1 (macOS is always VM-backed)
 # (DM numbers are the deployment models in docs/specs/03-spec-networking.)
 #
-# Flow: from a guaranteed-clean state, `min activate` must auto-spawn the
-# target's daemon and create a session; then list, warm-call, destroy
-# (verified delisted), and a clean `min stop`. Timing is reported but NOT
-# asserted.
+# Two proofs, in order, on EVERY lane:
 #
-# In-session exec over the daemon's SSH surface is NOT exercised here: only
-# `min run <task>` is accepted (arbitrary exec was removed), and a task needs
-# a `minimal.toml` in the otherwise-empty session workspace — which this shell
-# harness cannot seed on the VM lanes (the workspace lives inside the guest).
-# That path is proven in Rust instead: `exec::tests::end_to_end::exec_runs_echo_task`
-# (host daemon) and `minimald_exec_over_bridge` (over the libkrun bridge).
+#  1. Lifecycle: from a guaranteed-clean state, `min activate` must auto-spawn
+#     the target's daemon and create a session; then list, warm-call, destroy
+#     (verified delisted), and a clean `min stop` that the next command
+#     auto-respawns from.
+#
+#  2. Session sandbox: with the session live, attach an INTERACTIVE shell —
+#     which forks a real hakoniwa sandbox (on a VM lane, inside the guest, over
+#     the vsock bridge) — and prove the in-sandbox `min` helper. Inside the
+#     sandbox we `min add <tool>` a package that is NOT in the launcher
+#     baseline (base/coreutils/socat), then run it: the tool being absent
+#     before and runnable after proves `min add` reached the daemon over the
+#     in-sandbox `/run/minenv_sock` relay and hardlinked the package into the
+#     live rootfs. This is lane-agnostic: it operates on the session workspace,
+#     not the host project, so the identical attach+add+run runs everywhere.
+#     Timing is reported but NOT asserted.
+#
+#     A session is interactive by design, so we drive it like a real user
+#     through a REAL pty (scripts/e2e-attach-pty.py) rather than a pipe: pump
+#     the command stream, then, when the shell exits and the daemon shows its
+#     Detach/Delete prompt, answer it with keystrokes (Down + Enter => Delete).
+#     A pipe is not a tty and could not answer that prompt. Selecting Delete
+#     tears the session down, so it must be delisted afterwards. The same daemon
+#     prompt runs whether local or in-guest, so the pty driver covers every lane.
+#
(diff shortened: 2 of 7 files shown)


twitchyliquid64 (approved):

crates/minimald/src/session_host.rs
```diff
@@ -452,50 +460,71 @@ impl Binding {
...
+                {
+                    // User selected detach, keep going to disconnect
+                    Ok(Selection::At(0)) => {}
+                    // User cancelled selection, safest option is to detach
+                    Ok(Selection::Cancelled) => {}
+                    // User selected delete.
+                    Ok(Selection::At(1)) => self.delete_session(&mut w).await,
+                    Ok(Selection::At(_)) => unreachable!(),
+                    Err(e) => tracing::warn!(error = %e, "session-exit prompt failed"),
+                }
```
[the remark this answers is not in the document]
norrietaylor: Intentional under the reworked design (the PR objective changed since this review). Rather than time-bounding the prompt, the `min` client now reports whether its own context is a terminal (`MINIMAL_INTERACTIVE`), and the daemon uses that: a **non-interactive** client never reaches this prompt — it skips straight to delete-on-exit — so it can't wedge here.

For an **interactive** client, waiting is the correct behavior (there's a human at the tty who is expected to answer), and this path is unchanged from the pre-existing behavior on `main`. A dropped interactive connection isn't indefinite either: the channel read half hits EOF/error, `interact()` returns, and we fall through to detach (session preserved). So a bounded timeout would contradict the new "instead of timing out on the prompt" objective without closing a real gap.

crates/minimal/src/lib.rs
```diff
@@ -997,17 +997,26 @@ pub async fn cmd_attach(global: &GlobalArgs, args: AttachArgs) -> Result<(), any
...
-        "SendEnv=MINIMAL_SESSION_ID",
-        "-o",
-        &format!("ProxyCommand={proxy_cmd}"),
-        "-o",
-        "StrictHostKeyChecking=no",
-        "-o",
-        "UserKnownHostsFile=/dev/null",
-    ]);
+    ssh.env("MINIMAL_SESSION_ID", record.id.to_string())
+        .env("MINIMAL_INTERACTIVE", if interactive { "1" } else { "0" })
```
twitchyliquid64: The canonical way to signal interactivity is whether the ssh process requests a pty or not, if i recall correctly its `-t`vs `-T`.

crates/minimald/src/session.rs
```diff
@@ -703,6 +703,10 @@ impl Session {
             Some(pty) => pty,
             None => return Err(AttachError::NoPty),
         });
+        // Whether the attaching client's own terminal is interactive; drives the
+        // session-exit prompt (interactive) vs. delete-on-exit (non-interactive)
+        // decision in the binding.
+        let interactive = config.interactive();
```
twitchyliquid64: The more I think about this, I don't think we should do this. The intent of a session is _always_ interactive, its things like tasks and such that are intended for the more deterministic / exec-like execution. Lets use `expect` and/or pump keystrokes?

scripts/session-e2e.sh
```diff
@@ -1,36 +1,65 @@
...
+#     auto-respawns from.
+#
+#  2. Session sandbox: with the session live, attach an INTERACTIVE shell —
+#     which forks a real hakoniwa sandbox (on a VM lane, inside the guest, over
+#     the vsock bridge) — and prove the in-sandbox `min` helper. Inside the
+#     sandbox we `min add <tool>` a package that is NOT in the launcher
+#     baseline (base/coreutils/socat), then run it: the tool being absent
+#     before and runnable after proves `min add` reached the daemon over the
+#     in-sandbox `/run/minenv_sock` relay and hardlinked the package into the
+#     live rootfs. `min add` with no flag defaults to `--session` (this session
```
twitchyliquid64: Did we implement `--session` ?

scripts/session-e2e.sh
```diff
@@ -1,36 +1,65 @@
...
+#     only). This is lane-agnostic: it operates on the daemon-scaffolded
+#     session workspace, not the host project, so the identical attach+add+run
+#     runs everywhere. Timing is reported but NOT asserted.
+#
+#     Driving an interactive attach from a script: `min attach <sid>` (no
+#     `--command`) opens a PTY-backed shell; we pipe it a command stream ending
+#     in `exit`. When the shell exits, the daemon's session-exit prompt would
+#     historically block waiting for an arrow-key answer. The `min` client now
+#     reports whether its own context is a tty (MINIMAL_INTERACTIVE); a
```
twitchyliquid64: Lets just use `expect` or pump keystrokes instead of this `MINIMAL_INTERACTIVE` thing.

scripts/session-e2e.sh
norrietaylor: I don't think so. I have no idea where this came from. My clanker is getting testy with this one.

crates/minimald/src/session.rs
norrietaylor: Honestly, I could see this both ways. 

While I agree that our intent of a session as spec'ed was interactive and non-deterministic, generally I try to avoid adding unnecessary constraints. 

Additionally, there could be a use case here.  non-interactive and non-determinsitic are not mutually exclusive.  I could see passing a session to an agent running in prompt mode without a PTY.

I'm going to leave this one in draft mode while I figure out some deeper issues around why the DNS isn't resolving inside a github runner.

twitchyliquid64 (approved):

.github/workflows/ci-linux-kvm.yml
```diff
@@ -335,17 +335,26 @@ jobs:
...
         # and round-trip a session exec over the vsock bridge — the Linux/KVM
-        # counterpart of the macOS lane's CLI gate. E2E_PROJECT_DIR=/tmp: no
-        # project sync into the guest yet, so activate a path that exists in
-        # the guest image. minimal/minvmd resolve from the testbed on PATH.
+        # counterpart of the macOS lane's CLI gate. Since #748 activate uploads
+        # the project into the guest; E2E_PROJECT_DIR=/tmp still names a guest
+        # path. minimal/minvmd resolve from the testbed on PATH.
         env:
           MINVMD_BOOT_LOG: ${{ runner.temp }}/cli-e2e-boot.log
+          MINVMD_GVPROXY_BIN: ${{ runner.temp }}/gvproxy
```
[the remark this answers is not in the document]
norrietaylor: I am a CODEOWNER and we are building out the CI

scripts/session-e2e.sh
```diff
@@ -40,23 +63,67 @@
...
+elif [ -n "$E2E_VM" ] && [ "$E2E_PROJECT_DIR" = "/tmp" ]; then
+  # VM lanes pass /tmp; uploading all of /tmp is impractical, so seed a small
+  # subdir under it and upload that instead. On a persistent (self-hosted)
+  # runner this fixed path can carry a STALE project from a previous run — e.g.
+  # main's #748 seed, whose `[upstream]` has no `locked_commit`, which the graph
+  # loader then rejects ("does not pin it to a commit"). Start clean.
+  SEED_DIR="/tmp/mnl-e2e-project"
+  rm -rf "$SEED_DIR"
+  mkdir -p "$SEED_DIR"
+  PROJECT_DIR="$SEED_DIR"
```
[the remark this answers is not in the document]
norrietaylor: Fixed in 0b73aaad — the VM branch now uses `mktemp -d /tmp/mnl-e2e-project.XXXXXX` (a unique dir per run, matching the native branch), so concurrent VM runs on a persistent/self-hosted runner can no longer clobber each other's seed. The fresh dir also removes the need for the `rm -rf` I had added to guard against a stale/unpinned leftover. Removed on teardown as before.

Issue #747, opened 2026-07-14
feature: VM resource monitoring and configuration for minvmd


jessie-minimal:
## Summary

Add VM resource monitoring and runtime configuration to the `minvmd` host daemon. The existing `minvmd status` command exposes static boot-time values (`vcpus`, `ram_mib`); this work extends it to surface live resource consumption and adds a configuration surface for per-VM resource parameters.

## Problem

`minvmd status --json` currently reports `vcpus` and `ram_mib` as static values set at boot (R4.3). There is no mechanism to observe live resource utilisation during a session or to configure resource limits independently of stopping and restarting the daemon. This limits visibility for workload sizing and operational debugging.

## Desired outcome

- `minvmd status --json` returns live resource metrics (CPU utilisation, resident memory, disk I/O, uptime) alongside the existing static fields.
- A configuration surface (e.g., `minvmd config`) that validates and persists per-VM resource parameters (vcpu count, RAM MiB) to the state directory for consumption at next boot.
- Metrics are queryable without disturbing a running VM session.

## Scope notes

- The v0.1 spec deferred "RAM resizing and vcpu hot-add" as non-goals; this work covers monitoring current allocation and persisting configuration for the next boot cycle, not live hot-add.
- Likely affected files: `crates/minvmd/src/cmd/status.rs` (extend JSON output), `crates/minvmd/src/state.rs` (add config fields), a new `crates/minvmd/src/cmd/config.rs`, `crates/minvmd/src/vm.rs` (consume persisted config at boot time).
- Metric source (libkrun API vs. host-side `/proc/<vmm_pid>/status`) is a design decision for the spec phase.

---
Routed from gominimal/inbox#277 by inbox-route (https://github.com/gominimal/inbox/actions/runs/29353874624).
Routing-Key: inbox-route/I_kwDOSUhdos8AAAABIzLeMA

PR #775, opened 2026-07-15
feat(minvmd): resource monitoring, configuration, and warnings (#747)

norrietaylor:
## Summary

Implements [#747](https://github.com/gominimal/minimal/issues/747): live VM resource monitoring, a runtime configuration surface, and resource warnings for the `minvmd` host daemon.

- **Live metrics** — `minvmd status --json` gains a `metrics` object (`cpu_percent`, `resident_bytes`, `disk_read_bytes`, `disk_written_bytes`) sampled host-side from the VMM child PID (`State.vmm_pid`) via `sysinfo`; `null` when the VM is stopped. Every pre-existing field keeps its name and type.
- **Config surface** — `minvmd config set --vcpus/--ram-mib` validates and persists per-VM resource params to a new `config.toml`; `minvmd config show [--json]` reports the effective values and each value's source. Boot resolution is `env ?? persisted config ?? default`.
- **Warnings** — *proactive* over-allocation guardrails at `config set` (host cores/memory, x86_64 MMIO hole) and *reactive* memory/disk-pressure thresholds surfaced by `status`, plus a supervisor post-exit resource hint. Ports the `minimal-vm-mac` prior art (which samples the host VMM by PID and points users at the knob) to a typed, tested, host-side implementation.

## Design decisions

- **Config lives in a separate `config.toml`, not `State`.** `State::stopped()` / `StartingGuard` reset runtime state on every stop or crash, which would silently wipe config kept there.
- **Metrics are host-visible VMM-process usage**, not guest-internal per-process metrics (minvmd runs no in-guest agent). The reactive warnings are host-observable thresholds (sampled RSS vs the booted RAM cap; sparse-image allocation vs the image's own size), not guest cgroup/`df` introspection.
- **The `Running` state records the booted `vcpus`/`ram_mib`** (`State.booted_*`, `serde`-default for back-compat). `status` reports and warns against the value the live VM **actually booted with**, not a later `config set`'s next-boot resolution — otherwise `config set --ram-mib 16384` on a VM booted at 2048 would silence a genuine `memory_pressure` warning. (This came out of an adversarial self-review of the diff.)

## Out of scope / noted

- Live vcpu hot-add / RAM resize of a running VM (v0.1 non-goal; reaffirmed).
- A **pre-existing** `stop.rs` quiesce concern surfaced during review — `minvmd stop` unconditionally signals the VMM child after the guest Shutdown RPC. Untouched here (belongs to spec #08); worth a separate issue.

## Testing

- `cargo test -p minvmd` — 123 lib + 4 CLI-integration tests pass; gated VM harnesses (`MINVMD_E2E=1`) self-skip.
- `cargo clippy -p minvmd --all-targets -- -D warnings` — clean; `cargo fmt` applied.
- New `tests/config_cli_integration.rs` drives the real binary (config set/show/merge, invalid-value rejection, `status --json` schema) with no VM required.
- The R2.5 boot-across-restart proof artifact requires a Mac with libkrun + kernel/rootfs (`MINVMD_E2E=1`) and is documented in the spec's verification table.

## Spec

SDD spec added at `docs/specs/09-spec-minvmd-resource-monitoring/` (spec + architecture, frozen `R{1,2,3}.x` IDs, proof artifacts, assumption ledger).

## Summary by CodeRabbit

* **New Features**
  * Added `minvmd config show` and `config set` for viewing and configuring VM vCPUs and memory.
  * Added persistent resource settings with environment and default value precedence.
  * Enhanced `status` output with booted resources, uptime, process ID, and live CPU, memory, and disk metrics.
  * Added resource validation and host-capacity warnings.
* **Bug Fixes**
  * Improved handling of abnormal VM exits and preserved compatibility with older state files.
* **Documentation**
  * Added resource monitoring and configuration specifications.


crates/minvmd/src/cmd/mod.rs
```diff
@@ -70,6 +71,22 @@ pub fn ready_timeout() -> std::time::Duration {
...
+/// resource config (see [`effective_vcpus`]).
+pub const DEFAULT_VM_VCPUS: u8 = 2;
+
+/// Conservative upper bound on guest vcpus. The generic guest kernel's
+/// `CONFIG_NR_CPUS` is not pinned in this repo, so a resolved count is clamped
+/// here (and `config set` rejects above it): the boot path used to hardcode `2`,
+/// and an unbounded count from config/env could exceed the guest kernel's
+/// ceiling and panic the boot. Raise once the guest kernel's real limit is
+/// confirmed higher.
+pub const MAX_VM_VCPUS: u8 = 8;
```
msample: It would be more flexible if it would get the number of CPUs on the system and back off a couple as the max. E.g. I'm running this on big iron and want a lot of sandboxes to run concurrently in a VM.

crates/minvmd/src/metrics.rs
```diff
@@ -0,0 +1,82 @@
...
+//! long-running VM), and the data volume's host allocation is a monotonic
+//! high-water mark, not the guest ext4's free space. The reliable reactive
+//! resource signal is instead the supervisor's abnormal-exit hint
+//! (`cmd::run`), which fires on a guest workload's non-zero exit.
+
+use serde::Serialize;
+
+/// Live host-visible resource usage of the VMM process.
+#[derive(Debug, Clone, Serialize)]
+pub struct VmMetrics {
```
twitchyliquid64, line 20: I think we should do this stuff as an RPC in minimald, so we can get it for the daemon case as well.

norrietaylor: I was thinking I want to stream them to a collector somewhere 😛 

RPC sounds good! Do you think I should add here?

crates/minvmd/src/cmd/mod.rs
norrietaylor: Done in 0d9d963d — the fixed cap is gone. The ceiling is now host-derived: `max_vm_vcpus(logical_cores)` = host logical cores minus a 2-core reserve for the host side, floored at the default (2) so tiny hosts keep the baseline, saturated to the `u8` vcpu type. Both the boot-time clamp and `config set` validation use it, so big iron gets `cores − 2` instead of 8.

One caveat carried into the spec's assumption ledger: the guest kernel's `CONFIG_NR_CPUS` still isn't pinned in this repo, so on a very-many-core host the derived ceiling could in theory exceed it — generic configs ship ≥ 64, so it's flagged needs-confirm rather than re-capped.

twitchyliquid64 (approved):

PR #698, opened 2026-07-10
ci(linux-kvm): make the lane gate-ready — changes filter, aggregator, apt retries

norrietaylor:
## What

PR5 of the CI refactor ([#687](https://github.com/gominimal/minimal/issues/687)). Off `main`, independent of the mac stack.

Converts the KVM lane to the always-triggered pattern so it can become a required check after a soak:

- **Trigger-level path filter → `changes` job** (dorny/paths-filter, SHA-pinned `v3.0.2`, `pull-requests: read`). Same path list, now evaluated inside the workflow; `workflow_dispatch` runs everything. The header documents why trigger-level `paths:` must never return: a required context skipped by one stays "Expected" forever and wedges the PR.
- **New `if: always()` aggregator `ci-linux-kvm-success`** — green when the e2e passed *or* was path-skipped, red only on failure/cancellation. It joins the ruleset after the lane soaks green (admin step in #687); until then the lane stays advisory exactly as today.
- **Evidence-backed apt retries** (`scripts/ci/retry.sh`, 3×10s): mining ~200 runs per VM lane found **zero** failures in the artifact/libkrun/gvproxy fetches but a consistent transient class in apt — the only main-branch KVM failure in the window ([run 28897380911](https://github.com/gominimal/minimal/actions/runs/28897380911)) and both of its manual re-runs died in `Install build dependencies`, and the mac `artifacts` job failed twice in `Install protoc`. Those three call sites (KVM deps, `setup-rust` protoc, mac-artifacts protoc) are now retried; fetches stay un-retried.

## Verification

`actionlint` — no new findings (one pre-existing SC2012 carried over verbatim). This PR triggers the lane itself (always-triggered now), so its own run exercises the changes-gate, the retried apt step, and the aggregator.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **CI Improvements**
  * Improved Linux/KVM test workflow efficiency by running only when relevant changes are detected.
  * Added reliable status reporting for the Linux/KVM test lane.
  * Enabled prebuilt test binaries to run consistently across different environments.
* **Bug Fixes**
  * Improved executable discovery for end-to-end tests when build and test paths differ.
  * Improved installation reliability for required protocol-buffer tooling in CI.

PR #779, opened 2026-07-16
ci(macos): darwin testbed — the mini stops compiling

norrietaylor:
Implements **N6** — the last ordered subtask of the CI refactor epic [#687](https://github.com/gominimal/minimal/issues/687): all compilation moves off the Mac mini.

## What changed

**`unit` (GitHub-hosted arm64) is now the mac lane's build phase**, mirroring the KVM lane's `build-linux` → `test-kvm` split ([#698](https://github.com/gominimal/minimal/pull/698)):

- Unit tier runs via `cargo nextest run -p minvmd -p sessions` (integration harnesses excluded — they're the hypervisor tier), with doctests split into their own step, core-tests style.
- Builds `minvmd` and `min` (separate invocations — the feature-unification guard and its otool assert move here, next to the build they check).
- Stages a **`mac-testbed` artifact**: `bin/minvmd` + `bin/min` + `lib/libkrun.1.dylib` in the shipped layout, with `scripts/rewrite-macos-linkage.sh` applied to minvmd (the script's own verification reds the PR on a broken ship layout), plus the minvmd+sessions nextest archive.

**`e2e` (the mini) compiles nothing**: toolchain, libkrun provisioning, and every cargo invocation are gone. It downloads the testbed, restores exec bits, codesigns minvmd with the hypervisor entitlement, then:

- replays the kernel-less FFI smoke and the `_integration` harnesses from the archive (`cargo-nextest ... --archive-file`, auto-discovery filtersets unchanged);
- runs the unified session e2e with the testbed `bin/` on PATH.

Archive-extracted test binaries resolve `@rpath/libkrun.1.dylib` via `DYLD_FALLBACK_LIBRARY_PATH` pointing at the testbed `lib/` — exported **in-script**, because SIP strips `DYLD_*` crossing protected shells, so a step-level `env:` entry may never reach the test processes. The testbed minvmd needs no such help: its rewritten `@loader_path/../lib` rpath finds the bundled dylib — meaning the e2e now **boots the production linkage layout for real**, replacing the old static rewrite-on-a-throwaway-copy check.

## Why

- **One compile serves both tiers** — the unit/e2e build divergence (hosted `cargo test` build + a second full build on the mini) collapses into the single hosted build.
- **The mini does hypervisor work only** — a flake rerun replays just this test job, and the shared runner stops burning time on cargo.
- **Test what ships** — the exact dylib and the exact `bin/../lib` layout users receive is what CI boots.

Also updates the `docs/ci-strategy.md` mapping note (the archive-replay pattern now lives in the KVM *and* macOS lanes, per §7's "build nothing on this machine"). The CONTRIBUTING.md contract table needs no change — harness auto-discovery by `_integration` suffix is untouched.

## Validation

- `actionlint` clean.
- This PR touches `ci-macos.yml` + `scripts/**`-adjacent paths, so the `changes` filter runs the full lane against itself: the real proof is this PR's own `ci-macos-success`.
- Note: the mini's first run simply stops using its persistent `target/` dir; stale contents there are inert.

Refs: [#687](https://github.com/gominimal/minimal/issues/687) (N6)

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -315,6 +315,20 @@ jobs:
                   pkill -f '__krun-vmm' 2>/dev/null || true
                   pkill -f gvproxy 2>/dev/null || true
                   sleep 1
+            - name: Daemon lifecycle (run --detach → status → stop)
+              # Exercises the supervised daemon path (detach, status polling,
+              # documented stop semantics) that the integration harnesses and
+              # the CLI autospawn below do not cover directly — the same proof
+              # the KVM lane runs, and the script's header was written
+              # target-agnostic for exactly this testbed. minvmd resolves from
+              # PATH (the testbed bin/, added at unpack). Requires jq on the
+              # runner; the script fails fast with a clear error if absent.
+              env:
+                  MINVMD_KERNEL_PATH: ${{ runner.temp }}/kernel/vmlinuz
+                  MINVMD_ROOTFS_PATH: ${{ runner.temp }}/rootfs/rootfs.img
+                  MINVMD_INITRAMFS: ${{ runner.temp }}/initramfs/initramfs.cpio
+                  MINVMD_BOOT_LOG: ${{ runner.temp }}/boot-daemon.log
+              run: ./scripts/minvmd-lifecycle.sh
             - name: Fetch pinned gvproxy switch binary
               # minvmd spawns this (gvproxy-darwin) as the host-side switch that
               # provides the guest's egress (NAT + DNS). Without it minvmd boots
@@ -354,6 +368,7 @@ jobs:
                   name: minvmd-boot-logs
                   path: |
                       ${{ runner.temp }}/boot.log
+                      ${{ runner.temp }}/boot-daemon.log
                       ${{ runner.temp }}/cli-e2e-boot.log
                   if-no-files-found: ignore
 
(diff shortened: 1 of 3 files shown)


norrietaylor:
Two updates after the first run's red `e2e`:

1. **Failure cause** — `krun_smoke_bring_up` spawns its helper via the compile-time `CARGO_BIN_EXE_krun_smoke_child` path, which only exists on the build runner; the archive replay on the mini hit `NotFound`. Fixed by preferring nextest's runtime `NEXTEST_BIN_EXE_krun_smoke_child` (the helper ships in the archive — the "1 non-test binary" in the extraction log), falling back to the compile-time path for plain `cargo test`. Same relocation pattern as the harnesses' `MINVMD_BIN` override from [#698](https://github.com/gominimal/minimal/pull/698).

2. **Scope addition** — the mac lane now runs `scripts/minvmd-lifecycle.sh` (run --detach → status → stop), which its header already earmarked for "a macOS testbed later". Placed reap → lifecycle → session e2e, mirroring the KVM lane.

⚠️ One prerequisite: the lifecycle script requires **host `jq` on the mini** (it fails fast with a clear `::error` if absent). If the mini doesn't have it yet, a one-time `brew install jq` there is needed before this lane can go green.

twitchyliquid64 (approved):

PR #790, opened 2026-07-16
feat(minimal): warn and confirm before uploading from non-VCS directories

### LINKED ISSUE #770: Warn and Ask to Upload Files When A Session is Started when CWD is not a VCS Root
Currently we recursively upload all files (with some exceptions) in the current working directory into the session when a session is started. This is a good default if the CWD is the root of a version control repository, but not otherwise. There is a high risk of a user accidentally uploading their entire home directory which they almost certainly don't want.

Instead, we should check to see if the CWD is a VCS repository root. If it is not, we should ask the user to confirm that they want the files from the CWD uploaded before we do it.

We should support the following VCS systems:
- Git
- Mercurial
- SVN
- CVS
- Jujutsu (might get this one for free by supporting Git because it uses Git as a backend)

This should cover the vast majority of cases, but there are some other VCS that are either _very_ legacy or industry specific (the gaming industry in particular has some specialized ones).

0chroma:
## Summary

When activating a session from a directory that is not a version-control repository root, the CLI now warns the user and asks for confirmation before the recursive file upload. This prevents accidentally uploading large directory trees (e.g. a home directory) when running `minimal activate` from the wrong place.

- Adds `is_vcs_root(&Path) -> bool` to detect Git, Mercurial, SVN, CVS, and Jujutsu repository roots (including Git worktree `.git` files)
- In `cmd_activate`, if the project path is not a VCS root and stdin is a TTY, prompts the user before uploading; declining skips the upload and starts the session with an empty workspace
- Non-interactive contexts (CI, pipes, agents) skip the prompt and proceed with the upload as before
- `--sync none` remains available for explicit opt-out

Closes #770.

## Summary by CodeRabbit

* **New Features**
  * Added detection of common version-control repository roots (including Git worktrees/submodules).
  * For tarball sync, the upload source now uses the nearest project configuration directory (when available).
* **Bug Fixes**
  * If a configuration is present but malformed, activation now fails instead of falling back.
  * Confirmation for non-repository sources is now shown only in interactive terminal runs.
  * If confirmation is declined, the session workspace starts empty.
* **Tests**
  * Added coverage for repository-root detection, upload-root resolution, and malformed configuration handling.

> [!NOTE]
> ### Warn and prompt before uploading from non-VCS directories in `cmd_activate`
> - Adds `is_vcs_root` in [file_upload.rs](https://github.com/gominimal/minimal/pull/790/files#diff-b417966fca8aff7d83bfe7def43f0915d3cc2fcc1c3d02953c288fc85274bbf2) to detect VCS roots by checking for `.git`, `.hg`, `.svn`, `CVS`, and `.jj` markers (supports `.git` as file or directory).
> - Adds `resolve_upload_root` in [lib.rs](https://github.com/gominimal/minimal/pull/790/files#diff-1b62c63af5a5bd2dbc4b550bf4bc6d0fa87e649b54dc3a587da5a8fa947b3ce1) to walk up the directory tree and find the nearest `minimal.toml`, using its repo root as the upload directory instead of the invocation path.
> - In `--sync` tarball mode, `cmd_activate` now warns the user and prompts for confirmation when the upload root is not a VCS root; on decline, the upload is skipped and the workspace starts empty.
> - Non-interactive stdin bypasses the prompt and proceeds without confirmation.
> - Behavioral Change: uploads now originate from the resolved project root rather than the current working directory when a `minimal.toml` is found above it.
>
> 
> #### Changes since #790 opened
>
> - Changed `resolve_upload_root` function to return `Result<Utf8PathBuf, anyhow::Error>` and propagate non-NotFound mfile errors [b2bbefd]
> - Updated `cmd_activate` async function to handle `resolve_upload_root` Result and modified confirmation prompt default [b2bbefd]
> - Updated tests for `resolve_upload_root` to accommodate `Result` return type and error expectations [b2bbefd]
> - Moved upload root resolution in `cmd_activate` to execute before daemon connection establishment [f64bbd9]
> - Changed the conditional check in `cmd_activate` async function from `!std::io::stdin().is_terminal()` to `!can_prompt_interactively()` when determining whether to skip user confirmation prompts before uploading files from non-VCS root directories [b739b94]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized c07be15. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>
>


crates/minimal/src/lib.rs
```diff
@@ -971,11 +971,30 @@ pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(),
...
+            // Guard against accidentally uploading a non-VCS directory
+            // (e.g. `~`): if the project path is not a recognized VCS
+            // root, warn and ask for confirmation before the recursive
+            // upload. On non-interactive stdin (CI, pipes, agents) we
+            // proceed without prompting — `--sync none` remains available
+            // for explicit opt-out (#770).
+            let should_upload = file_upload::is_vcs_root(utf8_path.as_std_path())
+                || !std::io::stdin().is_terminal()
+                || confirm(&format!(
+                    "{utf8_path} is not a version control repository root. \
```
msample: is the utf8_path one of:
* the CWD with a minimal.toml  / .minimal/minimal.toml
* if above doesn't hold, an ancestor dir that has one of those minimal.toml files
* if none of the above hold the CWD

Ie it reflects the previous CLIs search alg?

twitchyliquid64: This is a good point, we probably want to walk the tree and always do the upload/thing from the root of the project.

For doing that for the mfile you can run `mfile::File::from_dir_recursive` then you can call `repo_path()` to get the base. https://github.com/gominimal/minimal/blob/main/crates/mfile/src/lib.rs#L652

0chroma: ahh good catch yes

twitchyliquid64 (approved):

Issue #788, opened 2026-07-16
established attach stream dies silently mid-session; client hangs forever with no EOF or timeout


norrietaylor:
## Summary

An established interactive attach went permanently silent mid-session while the VM, minimald, and the session itself all stayed healthy. The client hung forever with no EOF, no error, and no timeout. The evidence localizes the failure to the per-connection stream on the host↔guest vsock transport (`ssh.sock` → vsock → minimald): delivery stopped in both directions on that one stream, with the peer socket left open, while concurrent streams on the same transport kept working.

## Environment

- macOS host (Darwin 25.4.0, Apple Silicon), minvmd/libkrun backend, provider `local-0`
- Client: `min 0.0.1 (5ca67a5)` from the unstable release channel
- Server: `minimald 0.0.1 (dirty 5ca67a5)` (see release-hygiene note below)
- Session: `test` (`019f6947-e537-7fe1-a3db-fe3f27df7624`), host-net

## Timeline (PDT, 2026-07-15)

| Time | Event |
|---|---|
| 21:41 | VM boot (`minvmd run`, VMM pid 96763) |
| 21:56 | Interactive attach: `ssh -tt local-0` via `min proxy --socket .../ssh.sock` (pids 20371/20372) |
| ~22:05–22:09 | In-session activity: `min add rust`, `cargo`, `min add just` (package download), `just` |
| ~22:09 | Terminal froze immediately after the `just` error output; a final typed character was echoed, then nothing. `min ls` last-activity: 22:09:53. Last write to `data-vol.raw`: 22:08 |
| 22:39–22:50 | Investigation: everything except this one connection verified healthy; a concurrent re-attach to the same session got the same live shell |

## Findings

**The client did not crash.** `ssh` (20371) and its `min proxy` ProxyCommand child (20372) remained alive, 0% CPU, parked in `kevent`/read awaiting bytes that never arrived.

**The VM was healthy.** `sample` of the VMM showed both vCPUs idle in `wait_for_interrupt`, block workers and the vsock muxer normally parked in `kevent`. Guest: 3.6 GB of 4 GB RAM available, uptime consistent with the 21:41 boot. Host disk: 772 GB free.

**minimald was healthy, including on the same `ssh.sock`.** RPC (`min ls`, `min version`) responded normally. Fresh shell-request attaches to the *same* session succeeded and reached the *same still-running shell* (scrollback intact) while the wedged connection stayed dark.

[...]

PR #798, opened 2026-07-17
fix(minimald,minimal): keepalives instead of the inactivity reaper

norrietaylor:
## Summary

Fixes the reproduced mechanism behind [#788](https://github.com/gominimal/minimal/issues/788): russh's inherited `inactivity_timeout: Some(600s)` reaps healthy-but-quiet attaches; a lost close across the vsock ([#588](https://github.com/gominimal/minimal/issues/588)) then leaves the client — no keepalives, no timeout — hung forever. Full evidence chain and on-demand reproduction in the issue thread.

- minimald: `inactivity_timeout: None` (explicit), `keepalive_interval: 20s`, `keepalive_max: 3`, with named constants documenting the policy
- min client: mirror keepalive config — any silent transport death becomes a visible error within ~60s

Out of scope (tracked in #788): binding teardown on connection death (owned separately), bounded `attach()` for the establishment race, #588 upstream.

## Test plan

- [x] `cargo check -p minimal` (macOS) — client config compiles
- [ ] CI full matrix
- [ ] Field validation: idle an attached session >10 min on the new build — expect no reap, no wedge; kill the daemon under a live attach — expect client error within ~60s

Refs: #788

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSH connection resilience during extended inactivity, including laptop sleep and wake cycles, by refining keepalive behavior.
  * Enhanced server-side keepalive handling to better distinguish real disconnects from idle periods without unintended session drops.
* **Chores**
  * Increased CI Linux KVM build timeout to allow more time for the build phase.

### DIFF
--- a/crates/minimal/src/client.rs
+++ b/crates/minimal/src/client.rs
@@ -81,7 +81,15 @@ impl Client {
             })?
         };
 
-        let config = Arc::new(russh::client::Config::default());
+        let config = Arc::new(russh::client::Config {
+            // Mirror of the daemon's liveness policy: a silently dead
+            // daemon or transport becomes an error within ~60s instead of
+            // an indefinite hang. An idle connection is not a dead one, so
+            // inactivity_timeout stays at its default None.
+            keepalive_interval: Some(std::time::Duration::from_secs(20)),
+            keepalive_max: 3,
+            ..Default::default()
+        });
         let handshake = async {
             let mut handle = russh::client::connect_stream(config, stream, MinimalClientHandler)
                 .await
--- a/crates/minimal/src/client.rs
+++ b/crates/minimal/src/client.rs
@@ -81,15 +81,12 @@ impl Client {
             })?
         };
 
-        let config = Arc::new(russh::client::Config {
-            // Mirror of the daemon's liveness policy: a silently dead
-            // daemon or transport becomes an error within ~60s instead of
-            // an indefinite hang. An idle connection is not a dead one, so
-            // inactivity_timeout stays at its default None.
-            keepalive_interval: Some(std::time::Duration::from_secs(20)),
-            keepalive_max: 3,
-            ..Default::default()
-        });
+        // The client deliberately runs without keepalives: a laptop closed
+        // for an hour should reconnect transparently on wake rather than have
+        // the link torn down mid-sleep. The server's long-interval keepalive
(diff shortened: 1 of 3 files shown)


crates/minimal/src/client.rs
```diff
@@ -81,7 +81,15 @@ impl Client {
             })?
         };
 
-        let config = Arc::new(russh::client::Config::default());
+        let config = Arc::new(russh::client::Config {
+            // Mirror of the daemon's liveness policy: a silently dead
+            // daemon or transport becomes an error within ~60s instead of
+            // an indefinite hang. An idle connection is not a dead one, so
+            // inactivity_timeout stays at its default None.
+            keepalive_interval: Some(std::time::Duration::from_secs(20)),
```
twitchyliquid64: Nit: i think we will want to tune this when we go remote, as its not unusual for someone to close a laptop for an hour and open it and expect it to work (which can work when you dont have keepalives)

crates/minimald/src/server.rs
```diff
@@ -455,6 +455,18 @@ impl Server {
     }
 }
 
+/// How often an otherwise-quiet connection is probed with an SSH keepalive.
+/// The peer's transport replies automatically and replies count as inbound
+/// activity, so an idle-but-alive attach is never mistaken for a dead one.
+/// The client applies the mirror policy (see `minimal`'s `Client`).
+const KEEPALIVE_INTERVAL: std::time::Duration = std::time::Duration::from_secs(20);
```
twitchyliquid64: This is too much, maybe like 10 mins?

twitchyliquid64: i.e. battery life, this will be waking the radio every 20s (for remote)

twitchyliquid64 (approved):
I suggest:

 - keepalive interval in the 10 mins to 1 hr range
 - No client keepalives, or have them be in the 10mins or more range

crates/minimald/src/server.rs
norrietaylor: yikes! will fix good call

crates/minimal/src/client.rs
norrietaylor: agreed - in fact isn't that a major use case. being able to close your laptop and still have this working in the background... 

I'll make it better now.

Issue #801, opened 2026-07-17
Diagnostics subsystem: `min bug` support bundles


norrietaylor:
Tracking issue for the diagnostics subsystem: the `min bug` support-bundle command, the daemon-side bundle serving, persistent daemon logs, and the observability conventions they ride on.

## Motivation

[#788](https://github.com/gominimal/minimal/issues/788) was diagnosed post-mortem from files on disk — and the two biggest gaps were files that did not exist (no persisted daemon logs, no in-VM state capture). [PR gominimal/minimal#784](https://github.com/gominimal/minimal/pull/784) built the whole subsystem in one branch (+4107/−68 over 38 files) and got out of hand; it stays open as the **draft reference implementation** while the work re-lands as a spec plus a serial series of functionally complete PRs, each under 1000 lines.

## Spec

`docs/specs/10-spec-diagnostics/` — spec + architecture, merged in [#802](https://github.com/gominimal/minimal/pull/802).

## Status

**Units 1–7 are merged to `main`. Only Unit 8 (the bundle explorer) remains.**

The full host→daemon→guest path now works: `min bug` collects host state, reaches each provider's daemon over `DiagBundleTarZst`, nests the daemon's own bundle after a bounded bomb-check, and falls back to a read-only volume harvest when the daemon is unreachable. Every bundle — host and guest — converges on one `manifest.json` schema.

Units 6 and 7 went through an automated-review pass (CodeRabbit + Macroscope) before merge; 10 Macroscope findings were validated against the code and fixed, including two High-severity ones — a log-directory symlink escape and a nested-bundle bomb-cap bypass through tar PAX/longname extension members. Two of the review fixes corrected latent defects the audit surfaced rather than the units themselves (an unconditional `async_tar` Drop panic on any `finish` error; three unscoped archive paths). See [#878](https://github.com/gominimal/minimal/pull/878) and [#880](https://github.com/gominimal/minimal/pull/880) threads.

## Units

- [x] Unit 1 — `feat(diagnostics)`: crate foundation (`BundleWriter` generic over `AsyncWrite`, manifest, fail-closed redaction engine, listing, command capture) — [#803](https://github.com/gominimal/minimal/pull/803) (merged)
- [x] Unit 2 — `feat(minimal)`: `min bug` host bundle (command + system/env/config/state/log collection) — [#816](https://github.com/gominimal/minimal/pull/816) (merged)
[...]

PR #803, opened 2026-07-17
feat(diagnostics): add the diagnostics crate foundation

norrietaylor:
## Summary

Unit 1 of the diagnostics spec ([#802](https://github.com/gominimal/minimal/pull/802), tracking [#801](https://github.com/gominimal/minimal/issues/801)): the app-agnostic machinery every later unit consumes.

- **`bundle`** — `BundleWriter<W: BundleSink = tokio::fs::File>`, one writer, two modes (R1.2/R1.3): `create()` writes an owner-only (0600) file with entries under `{root}/` and flush+fsyncs on finish; `stream()` writes rootless through any blessed sink (`DuplexStream` — the daemon shape) and flush+half-closes. Both append the same `manifest.json` as the final entry at the mode-specific path. `BundleSink` is sealed; its `Sync` supertrait is load-bearing (`async_tar::Builder` requires it).
- **No-follow file collection** (R1.4): `add_file_tail` opens with `O_NOFOLLOW` plus a descriptor `fstat` regular-file check — no stat-then-open TOCTOU window — and bounds the read with `take(cap)` against live-log growth.
- **`redact`** — the fail-closed key-based engine, unchanged from the reference implementation (compound public-key regressions included) (R1.5).
- **`listing`** — now returns a structured `Listing { text, truncated }`; an unreadable root is an `Err` the caller records in its manifest instead of an empty-looking listing; the cap appends an explicit truncation marker (R1.6).
- **`capture`** (new) — `command_capture(cmd, args, timeout)`: stdin closed, `kill_on_drop`, and a timed-out child is killed with its partial stdout/stderr returned inside a typed `CaptureError::Timeout` — partial evidence from a hung tool is still evidence (R1.7). Replaces three private near-duplicates in the reference tree.
- **Curated API** (R1.8): crate-root `pub use`, `#[non_exhaustive]` on the manifest types. Manifest field is `version` (producer-agnostic — the daemon emits manifests too). The reference tree's unused `walkdir` dependency is dropped.

## Test plan

- [x] 20 unit tests: redaction compound keys, file-vs-stream round-trip identical modulo root with manifest last at the exact mode-specific path, 0600 archive mode, symlink refused at open, tail-cap bounding, listing truncation marker + unreadable-root error, capture timeout kills the child and retains partial output
- [x] `cargo clippy -p diagnostics --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt` clean

Spec: `docs/specs/10-spec-diagnostics/` (Unit 1, R1.1–R1.8).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1505,6 +1505,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "diagnostics"
+version = "0.0.1"
+dependencies = [
+ "anyhow",
+ "async-compression",
+ "async-tar",
+ "chrono",
+ "libc",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
+]
+
+[[package]]
 name = "diff"
 version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1519,6 +1519,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-stream",
+ "walkdir",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/checkouts",
     "crates/common",
     "crates/decode",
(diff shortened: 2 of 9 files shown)


twitchyliquid64 (approved):
YESSSSS

PR #811, opened 2026-07-17
feat(minimal): add bulk session destruction

jtnkminimal:
## Summary by CodeRabbit

* **New Features**
  * Enhanced the destroy CLI to terminate either one specified session or all active sessions in a single command.
  * Added interactive yes/no confirmation for bulk destruction (configurable via force), while supporting non-interactive/forced execution.
* **Bug Fixes**
  * Bulk destruction now stops and reports when any session removal fails.
  * “Destroy all” now succeeds cleanly when there are no active sessions.
* **Tests**
  * Expanded CLI integration coverage for both single-session and all-sessions destroy flows.

### DIFF
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -343,7 +343,14 @@ pub struct LsArgs {
 #[derive(Debug, Args)]
 pub struct DestroyArgs {
     /// Session identifier (UUID or session name)
-    pub session: String,
+    #[arg(required_unless_present = "all", conflicts_with = "all")]
+    pub session: Option<String>,
+    /// Destroy all sessions
+    #[arg(long)]
+    pub all: bool,
+    /// Skip confirmation when destroying all sessions
+    #[arg(long, short, requires = "all")]
+    pub force: bool,
 }
 
 #[derive(Debug, Args)]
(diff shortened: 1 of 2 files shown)


crates/minimal/src/lib.rs
```diff
@@ -343,7 +343,14 @@ pub struct LsArgs {
...
 pub struct DestroyArgs {
     /// Session identifier (UUID or session name)
-    pub session: String,
+    #[arg(required_unless_present = "all", conflicts_with = "all")]
+    pub session: Option<String>,
+    /// Destroy all sessions
+    #[arg(long)]
+    pub all: bool,
+    /// Skip confirmation when destroying all sessions
+    #[arg(long, short, requires = "all")]
```
twitchyliquid64, line 352: If a user did `--all`, do we really need a confirmation?

crates/minimal/src/lib.rs
```diff
@@ -601,6 +608,19 @@ fn confirm(question: &str) -> Result<bool, anyhow::Error> {
         || trimmed.eq_ignore_ascii_case("yes"))
 }
 
+/// Prompt for confirmation, defaulting to no.
+fn confirm_default_no(question: &str) -> Result<bool, anyhow::Error> {
```
twitchyliquid64: Assuming we are keeping the confirmation: use the existing `confirm()` function, add parameters for the default no etc as you need to

crates/minimal/src/lib.rs
jtnkminimal: I feel a confirm is a nice to have for now, we can remove if it gets annoying but I'd say we're the ones that will be killing everything, most folks will have pets running not cattle.

jtnkminimal:
Addressed in [`38103415`](https://github.com/gominimal/minimal/commit/38103415): the existing `confirm` helper now accepts an explicit default, renders `[Y/n]` or `[y/N]`, and all existing call sites pass their intended default. The duplicate `confirm_default_no` helper has been removed.

twitchyliquid64 (approved):

PR #847, opened 2026-07-20
feat(client): user policy callbacks

evanspearman:
The client's activate flow now prompts the operator (or emits a paste-ready user_policy.toml snippet in --no-prompt/non-TTY mode) whenever Phase 3 hits a project- or package-declared var/patch the user's UserPolicy can't auto-decide, instead of aborting the composition. This closes the loop the callbacks feature has been building toward: daemon-side project/package composables can now contribute vars and patches without the operator having to hand-craft their user_policy.toml up front.

What's in

- crates/minimal/src/prompt.rs — two PolicyHooks impls: InteractivePrompt (dialoguer-driven, six choices per item, permanent choices mutate a RefCell-held policy that save_user_policy writes atomically via write-tmp+rename with a RemoveOnDrop guard and .bak) and NoPromptHook (fake-approves so both var and patch hooks fire and accumulate every pending item into a summary; the caller checks the summary before submitting).
- carries_user_data short-circuit on both gate_vars (loadout side) and classify_var (client-side gate). Vars whose value is a Specified literal or an InheritWithDefault fallback skip the policy check — policy exists to gate user-env crossings into the sandbox, not to referee what packages hardcode. The bit crosses the wire via WireResolvedVar.carries_user_data.
- ItemDecision::IgnoreOnce joins AllowOnce / UseRule so IgnorePermanent and IgnoreOnce have a clean expression.
- WirePatchVerdict::Approved grows a per-file destination so dir mappings (~/.claude/** etc.) fan out to unique per-file destinations instead of collapsing every file onto the pending's shared base. Consumed on the daemon side after SandboxRelPath::try_new rejects both absolute paths and .. components, so a malicious client can't traverse.
- mfile::PackageFsMapping::{File, Dir} replaces the stringly-typed "path ends in /** = dir" convention. PackageComposable::contribute owns the walk-pattern shaping, so no caller can accidentally construct a bare-dir mapping the client walker would drop.
- compute_verdict + submit_verdict_and_wait split out of drive_pending_to_active so cmd_activate's non-interactive branch can inspect the NoPromptHook summary before actually submitting (fake approvals never reach the daemon).
- Ordering: min activate is now CreateSession → WorkspaceFilesTarZst → ConfigureLoadout → SubmitVerdict. The workspace upload sits between create and configure so ConfigureLoadout's graph_from_all_packages sees the project's minimal.toml (upstream layers still come through the daemon's own SourceProvider/VCS side channel).
- Per-name graph resolution in build_composables: one unresolvable package no longer wipes out every other package's contribution. Failures land as warn! with the missing name(s).

Other correctness / UX fixes

- paths::RelPath::try_new rejects .. components (new Error::ContainsParentDir).
- can_prompt_interactively() requires both stdin and stderr to be terminals — the old stdin_is_tty was misnamed (checked only stderr).
- save_user_policy uses a RemoveOnDrop guard so a failed rename doesn't leave <policy>.tmp behind. The --no-prompt snippet renders through toml_edit so paths containing " or \ escape correctly.
- apply_var_verdicts dedupes by PendingId up front (mirrors apply_patch_verdicts); duplicate entries surface as duplicate, not unknown.
- Client-side missing-walk-root emits an actionable InvalidWireItem instead of shipping an incomplete verdict.
- DenyPermanent glob-fail warning wording differs from the allow/ignore downgrade so the operator isn't told a rule "applies to this activation only" when the activation is actually aborting.
- Compose error messages carry the offending project_path.
- is_writable walks every missing ancestor (not just the immediate parent) and treats an empty relative-path parent as ..
- extract_fs_mappings trims trailing slashes and drops empty/root paths with a warn.

Docs

crates/sessions/docs/COMPOSITION.md rewritten to match the current shape: three-RPC flow, carries_user_data short-circuit as a first-class invariant, PackageFsMapping file/dir distinction, NoPromptHook's fake-approve+summary-check pattern, and the daemon's use of SourceProvider for upstream layers (workspace tar only carries the mfile itself in the common case).

## Summary by CodeRabbit

* **New Features**
  * Added interactive policy prompts during activation with per-item allow/ignore/deny decisions, including saved policy updates when possible.
  * Added `--no-prompt` for non-interactive activation; if approvals are still required, it aborts with a paste-ready `user_policy.toml` snippet.
  * Improved package filesystem mapping with typed file/dir behavior and correct directory globbing/trailing-slash handling.
* **Bug Fixes**
  * Rejected unsafe relative paths containing parent-directory traversal.
  * Prevented `Draft` pending contributions from being overwritten by subsequent loadout configuration.
  * Improved provenance-aware gating and stricter verdict validation (including duplicate/invalid approval safeguards).
* **Documentation**
  * Updated session composition docs and the example `minimal.toml` to reflect the new activation/prompt flow.


twitchyliquid64 (approved):
I just realized I never got you an RPC for uploading to the homedir: the `WorkspaceFilesTarZst` unpacks to what gets mapped to `/workbench` and not `/home`. Do we need an RPC for uploading to the home dir?

evanspearman:
> I just realized I never got you an RPC for uploading to the homedir: the `WorkspaceFilesTarZst` unpacks to what gets mapped to `/workbench` and not `/home`. Do we need an RPC for uploading to the home dir?

I'm just planning out that work now. I can just add it as part of that change.

norrietaylor (approved):
LGTM

norrietaylor:
I did have claude run an ultraview as it is pretty big. It came back very very clean but claude did raise one thing:


> 🟠 cmd_activate leaks a Draft session on transient failure — crates/minimal/src/lib.rs:1017–1034
> 
> After CreateSession succeeds, three failure paths propagate via ?/bail! without send_abort, leaking the daemon-side Draft slot until manual min destroy or restart (no reaper exists):
> - tar upload failure — lib.rs:1021–1024
> - ConfigureLoadout transport failure — lib.rs:1032–1038
> - the Errorable::Err arm — lib.rs:1039–1044
> 
> These are ordinary failure modes (network hiccup, permission error, compose fault). Fix is mechanical: mirror the send_abort-on-error pattern already used by submit_verdict_and_wait (lib.rs:800–828) and drive_pending_to_active — the invariant is even documented at lib.rs:797–799.

I leave it at your discretion if this is something to fix right now!

evanspearman:
> I did have claude run an ultraview as it is pretty big. It came back very very clean but claude did raise one thing:
> 
> > 🟠 cmd_activate leaks a Draft session on transient failure — crates/minimal/src/lib.rs:1017–1034
> > After CreateSession succeeds, three failure paths propagate via ?/bail! without send_abort, leaking the daemon-side Draft slot until manual min destroy or restart (no reaper exists):
> > 
> > * tar upload failure — lib.rs:1021–1024
> > * ConfigureLoadout transport failure — lib.rs:1032–1038
> > * the Errorable::Err arm — lib.rs:1039–1044
> > 
> > These are ordinary failure modes (network hiccup, permission error, compose fault). Fix is mechanical: mirror the send_abort-on-error pattern already used by submit_verdict_and_wait (lib.rs:800–828) and drive_pending_to_active — the invariant is even documented at lib.rs:797–799.
> 
> I leave it at your discretion if this is something to fix right now!

I think I'll defer these as I think there are going to be some related changes in the next PR.

PR #860, opened 2026-07-20
feat(stdlib): redistributable attr class (superseded)

bryan-minimal:
Superseded by #861 (same change, reworded).

PR #861, opened 2026-07-20
feat(stdlib): redistributable attr class

bryan-minimal:
Adds the `redistributable | Bool` attr class (default-absent = true): declares whether a package's built artifacts may enter a shared binary cache. Set false when the upstream license or terms forbid redistribution — build infrastructure then neither uploads, indexes, nor seals the artifacts, and consumers build the package locally from the upstream source via the planner's existing index-miss fallback.

Schema + docs only — nothing in this repo reads it. Attrs are unhashed, so setting it causes zero rebuilds.

Supersedes #860.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6220,7 +6220,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "stdlib"
-version = "0.0.17"
+version = "0.0.18"
 dependencies = [
  "fd-lock",
  "tempfile",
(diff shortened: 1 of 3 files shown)


twitchyliquid64:
Also bump the version of the `stdlib` crate - then in `pkgs` we can set that to the minimum version in `minimal.toml`

bryan-minimal:
Done — stdlib bumped to 0.0.18 in the follow-up commit, and gominimal/pkgs#477 (draft) now sets `minimum_version = "0.0.18"` in minimal.toml alongside the attrs.

twitchyliquid64 (approved):

PR #872, opened 2026-07-21
feat(minvmd): forward RUST_LOG to the guest via the kernel cmdline

norrietaylor:
## Problem

The microVM's `/init` **is** minimald (pid-1), so the kernel starts the guest daemon with an **empty environment** — nothing from the minvmd host process crosses the VM boundary. The guest daemon's log level is therefore unraisable from the host, which blocked a live debugging session: the host VMM child had `RUST_LOG=info,russh=debug,minimald=debug` in its environment and zero DEBUG lines reached the guest console log.

libkrun's `krun_set_env` / `krun_set_exec(envp)` do not help — they inject into "the context of the executable" for libkrun's own guest-init path (an `exec_path` relative to `krun_set_root`), which this initramfs boot model does not use. minvmd does not even bind `krun_set_env`.

> [!NOTE]
> ### Forward `RUST_LOG` from host to guest via kernel cmdline in `minvmd`
> - Reads `RUST_LOG` from the host environment when building the kernel command line and appends `RUST_LOG=<value>` to the base `console=hvc0` cmdline if the value passes validation.
> - Validation in the new `kernel_cmdline` function rejects empty values, values containing whitespace, and values that would cause the total cmdline to exceed the kernel's `COMMAND_LINE_SIZE` (2048 bytes); rejected values log a warning and fall back to the base cmdline.
> - Returns a `Cow<'static, str>` to avoid allocation when no forwarding is needed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 44d72ae.</sup>
> 
>

### DIFF
--- a/crates/minvmd/src/vm.rs
+++ b/crates/minvmd/src/vm.rs
@@ -5,6 +5,7 @@
 //! selects how the VM attaches to the per-host gvproxy switch supervised by
 //! [`crate::net`] (R1.4, R1.5).
 
+use std::borrow::Cow;
 use std::path::PathBuf;
 
 use minimald_rpc::{EgressPolicy, NetworkMode};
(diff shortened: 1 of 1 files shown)


norrietaylor:
@macroscope review

twitchyliquid64 (approved):

PR #874, opened 2026-07-21
fix(minimald): retry ENOBUFS on the guest vsock write path

### LINKED ISSUE #869: minimald: vsock upload killed by guest skb-overhead reset (ENOBUFS) — libkrun emits sub-KB packets under backpressure
## Summary

Uploading project files over the guest vsock connection fails with a Linux `ENOBUFS` ("No buffer space available", os error 105) raised inside the guest. The client sees:

```
error: Failed to upload project files: copying tar stream to channel: channel closed
```

Root cause is now measured rather than inferred: **libkrun emits sub-KB vsock packets under receiver backpressure, and Linux ≥ 6.12.92 resets the connection when queued `sk_buff` overhead exceeds the receive window.** The credit accounting is *not* violated. Details and evidence below.

## Reproduction

```
Client: minimal 0.5.0-rc2.dev.10.g4bfdcb09
Server: minimald 0.5.0-rc2.dev.10.g4bfdcb09
Stdlib: 0.0.18
```

```
min activate --loadout dev --name triage --attach .
```

Minimal deterministic fixture: a project containing a single file that compresses to **12,784,144 bytes** (a 45 MiB slice of an ext4 image taken at offset 95 MiB). Failure rate on a stock guest is **~89%** (8 of 9 in one sample; 6 of 6 in another).

The trigger is **write rate**, not volume, duration, or file count. Measured, three runs each:

| Payload | On the wire | Result |
|---|---|---|
| 512 MB incompressible random | 512 MB (zstd self-throttles the client to ~180–256 MB/s) | passes |
| 20,000 files / 78 MB | 78 MB | passes |
| 45 MB ext4 slice | 12.78 MB | **fails 3/3** |
| 45 MB ext4 slice from a different offset | 0.11 MB | passes |

Anything that slows the host→guest path makes it disappear: a 1 µs sleep per 64 KiB in the client (a 1.28 GB upload then completes in 22 s), `RUST_LOG=debug` on the client, or `MINVMD_KRUN_LOG=debug` in libkrun (4/4 passes). `tokio::task::yield_now()` in place of the sleep does **not** help — it needs wall-clock drain time, not scheduler fairness.

## Guest-side log

```
INFO conn{conn=5 transport="vsock"}: minimald::server: accepted connection peer=cid: 2 port: 4144818557
WARN conn{conn=5 transport="vsock"}: minimald::server: session ended with error error=Protocol error: No buffer space available (os error 105)
```

## Root cause

`ENOBUFS` here is a **receive-side** error, not a send failure. Guest kernel 6.12.94, `net/vmw_vsock/virtio_transport_common.c`:

```c
static bool virtio_transport_inc_rx_pkt(struct virtio_vsock_sock *vvs, u32 len)
{
    u64 skb_overhead = ((u64)skb_queue_len(&vvs->rx_queue) + 1) * SKB_TRUESIZE(0);
    if ((u64)vvs->buf_used + len > vvs->buf_alloc ||     /* (A) peer exceeded credit */
        skb_overhead > vvs->buf_alloc)                   /* (B) queued skb overhead  */
        return false;
```

On rejection, `virtio_transport_recv_connected()` sends a RST, calls `virtio_transport_do_close()`, sets `sk->sk_err = ENOBUFS`, and `vsock_remove_sock()`. The socket is destroyed; the sticky error surfaces on russh's next operation, which is why the whole SSH session dies rather than one channel.

The chain, measured end to end:

1. Guest userspace applies backpressure; the receive queue fills — `rx_bytes ≈ buf_used ≈ 250 KB` of a 256 KiB `buf_alloc`, i.e. essentially nothing drained.
2. libkrun's `recv_to_pkt()` (`src/devices/src/virtio/vsock/unix.rs`) clamps every read to remaining credit — `max_len = min(buf.len(), peer_avail_credit())` — so as credit shrinks, its packets shrink with it.
3. The stream degenerates into **~540-byte packets** (measured `len` 332–744 B; queue average `buf_used/rx_qlen` = 541–564 B).
4. Per-skb overhead is `SKB_TRUESIZE(0)` = 576 bytes on arm64 — **more than the payload each packet carries**. With `buf_alloc` = 262144 the ceiling is 455 queued skbs, and `(455+1) × 576 = 262,656 > 262,144`.
5. Condition (B) fires while credit is still available. The connection dies at **98.2–98.9%** of the upload, consistently.

### Evidence

A guest kernel instrumented with a `pr_err` on the rejection path (patch is 22 lines, only in the already-fatal branch, no fast-path cost) attributed **8 of 8** rejections to (B), with (A) false every time:

```
VSOCKDBG instrumented kernel active (issue 869) SKB_TRUESIZE_0=576
VSOCKDBG reject A_credit=0 B_skbdepth=1 buf_used=246388 len=640
  buf_used_plus_len=247028 buf_alloc=262144 rx_qlen=455
  skb_overhead=262656 SKB_TRUESIZE_0=576 rx_bytes=246080
```

Credit headroom still unused at the moment of each reset: **5,156 / 6,268 / 9,272 / 10,396 / 11,204 / 12,112 / 15,020 / 15,116 bytes**. `rx_qlen` was exactly 455 in all eight events — the `262144/576 = 455.1` threshold. The bug did not vanish under this instrumentation (8/8 failures), unlike every other probe attempted.

### Why it started recently

The reset behaviour is new. Bracketing the stable tree: absent in v6.12.91, present from **v6.12.92**, introduced by `a4f0b001782b` ("vsock/virtio: reset connection on receiving queue overflow"), backported as `06fa755325df`. Before it, the packet was silently dropped. The commit message names this exact case — *"or if the skb overhead … is too high"*.

`gominimal/pkgs` bumped the guest kernel **6.12.43 → 6.12.94 on 2026-07-01** ([gominimal/pkgs#311](https://github.com/gominimal/pkgs/pull/311)), stepping over 6.12.92. Same upload code, same libkrun, newly fatal.

## Ruled out

- **Backpressure/retry on the write path.** This issue originally proposed it. It cannot work: by the time `ENOBUFS` is observable the socket has already been closed and removed. Implemented and tested in-guest on [#874](https://github.com/gominimal/minimal/pull/874) — the adapter engaged (`retries=1`, never reaching its deadline) and the upload still failed 3/3. That PR should be closed.
- **libkrun violating credit.** Disproven by the instrumented kernel: 8/8 rejections had credit to spare.
- **[libkrun#772](https://github.com/libkrun/libkrun/issues/772)** (guest→host credit starvation). Fixed by `1bebfaac`, which is an ancestor of v1.19.0 — present in every version we run.
- **Upgrading libkrun 1.19.0 → 1.19.4.** Zero files matching `vsock` changed between those tags.
- **Host resource contention.** Fails 4/4 on an idle host (18 cores, 128 GiB, no swap; guest is 10 vCPU / 16 GiB).
- **tokio / tokio-vsock.** `stream.rs` faithfully relays the kernel's sticky `sk_err`; only `WouldBlock` is retryable and this is not that.

## Fix

The defect belongs to libkrun: it should never emit a packet whose skb overhead exceeds its payload. **The first attempt at that fix has been tested and does not work** — see below.

### Attempted and refuted: a minimum-credit floor

The proposed fix extended libkrun's existing `WaitForCredit` path from "credit is zero" to "credit is below a floor" (~4 KiB). Runtime A/B on macOS, same dev stack and minvmd binary, libkrun built from the same pinned commit with the same script, only the dylib differing:

|---|---|---|---|
| patched (4 KiB credit floor) | 6 | 0 | **6** |

The floor changes nothing. The most likely reason the lever was wrong: libkrun's packets are probably small because `recv()` returns only what is buffered in the host unix socket at that instant, **not** because `peer_avail_credit()` is clamping `max_len`. If so, gating on credit cannot help — a fix has to coalesce on *available bytes*. This also fits the timing sensitivity: slowing the path down lets more bytes accumulate per read, producing larger packets and fewer skbs.

Consistent with that, `MINVMD_KRUN_LOG=debug` makes the reproducer pass 3/3 on **both** the stock and patched builds — a static queue-accounting ceiling should not dissolve when the vsock path is merely slowed.

### Found while testing the refuted patch

A separate, real libkrun defect: with the credit floor in place, a fast failure became a multi-minute hang. `push_packet()` only writes into the used ring; the IRQ is raised solely by `process_proxy_update()` when `signal_queue` is set, which in `process_event` comes from `recv_pkt()`'s `have_used`. So when the first packet of a batch must wait, the credit request is queued **with no interrupt** while polling is disarmed in the same breath, and nothing wakes the connection. Reachable on unpatched upstream — just rare, since the request is normally batched behind data packets. `tsi_stream.rs` has the identical gap. Worth upstreaming on its own merits.

In flight:

- The patch-carrying mechanism for the macOS source build (`vendor/libkrun/patches/` + an apply step in `scripts/build-libkrun-macos.sh`) is verified both ways — applies cleanly, and a deliberately broken patch hard-fails the build. Worth keeping independently of any particular patch.
- The same mechanism for Linux via the `gominimal/pkgs` `libkrun` package, CI-green. Both are held in draft until a validated patch exists to carry.
- Guest-side hardening in minimald: raise the vsock receive window. **Validated: 0/6 failures at an 8 MiB window against 6/6 on the same machine at 256 KiB.** Note `SO_VM_SOCKETS_BUFFER_MAX_SIZE` must be raised *before* `SO_VM_SOCKETS_BUFFER_SIZE`, since `vsock_update_buffer_size()` clamps the latter to the former and both default to 256 KiB — a request that appears to succeed otherwise changes nothing.
- An upstream libkrun issue, pending.

**As of now the only measure demonstrated to fix the reproducer is the enlarged guest receive window.**

The window change is headroom, not immunity: it does not alter the ratio between the two ceilings, it keeps a normal upload out of the credit-starved tail where packets degenerate. Once libkrun is patched everywhere it is defence in depth rather than a fix.

## Adjacent defects found while investigating

- `upload_workspace_files` returns `Ok(())` when the connection drops mid-unpack: `channel.wait()` yields `None` both on a clean close and on a dead connection, so a reset during unpack is reported as success and activation proceeds with an unpopulated worktree ([gominimal/minimal#901](https://github.com/gominimal/minimal/issues/901)).
- Daemon-side errors on every oneshot RPC never reach the client — `ServeOneshot::handle_channel` writes nothing on the error path, so the reason is logged only in the daemon and the user sees `EOF while parsing a value` (same issue).
- Some failures **hang the client indefinitely** instead of returning an error (observed at 13 min and ~9 min before being killed).
- `min activate` ignores `-C/--repo-dir` and uploads the cwd instead ([#873](https://github.com/gominimal/minimal/issues/873)).
- The upload excludes only `target` and `node_modules`, so it ships `.git` and everything else — 1.28 GB in this repo.

norrietaylor:
Fixes [#869](https://github.com/gominimal/minimal/issues/869).

## What actually breaks

A `write(2)` on the guest AF_VSOCK socket fails with `ENOBUFS` (os error 105)
when the virtio TX virtqueue is full — the guest has queued more than the
host-side VMM has drained. `tokio-vsock` reaches that error *through* an already
satisfied write-readiness guard and surfaces the raw errno as an `io::Error`
that is **not** `ErrorKind::WouldBlock`, so nothing above it retries. In
`minimald` the error then escapes through russh and kills the **entire SSH
connection** rather than the one channel that was writing — which is why the
client sees `Failed to upload project files: … channel closed`, and why
`boot.log` shows the failure at connection scope:

```
WARN conn{conn=5 transport="vsock"}: minimald::server: session ended with error
     error=Protocol error: No buffer space available (os error 105)
```

## Measured evidence (controlled experiments, each repeated 3×)

The failure is **not** intermittent, not memory exhaustion, and not a function
of payload size, file count or wall-clock duration. The discriminator is the
client's *instantaneous write rate* into the vsock:

| Payload | Compresses to | Result |
|---|---|---|
| 512 MB incompressible random data | ~512 MB | **passes** — zstd self-throttles the client to ~180–256 MB/s |
| 20,000 small files / 78 MB | — | **passes** |
| 45 MiB slice of an ext4 image | 12.8 MB | **fails every time**, `ENOBUFS` 0.60 s after the connection is accepted |
| same-size slice, low entropy | 0.11 MB | **passes every time** |

Two further results pin down the mechanism:

- Inserting `tokio::time::sleep(1µs)` after each 64 KB client write makes it
  pass, including a 1.28 GB upload that ran 22 s.
- Replacing that sleep with `tokio::task::yield_now()` still **fails** at 0.60 s.
  Chunking to 64 KB with no sleep also still fails.

So the transport needs *wall-clock time to drain*, not scheduler fairness. Host
resources are not implicated (18 logical cores / 128 GiB, guest 10 vCPU /
16 GiB, zero swap in use, 8 host cores idle).

## The fix

A new `EnobufsBackpressure<S>` adapter (`crates/minimald/src/backpressure.rs`)
wraps the accepted vsock stream and restores the backpressure semantics every
caller above it already assumes:

- A write attempt failing with `ENOBUFS` — matched on
  `raw_os_error() == Some(libc::ENOBUFS)`, never on the rendered message — is
  retried after a short sleep instead of propagating.
- Backoff starts at 100 µs, doubles to a 4 ms cap, and **resets** as soon as any
  write lands. (Tokio's timer granularity is 1 ms, so the first step is in
  practice "one timer tick".)
- A contiguous run of `ENOBUFS` lasting longer than **5 s** gives up and returns
  the underlying `io::Error` unchanged, so a genuinely wedged transport still
  surfaces exactly as it does today rather than hanging the session forever.
- Reads, flush, shutdown and every non-`ENOBUFS` error pass through untouched. A
  healthy stream behaves exactly like the stream it wraps.
- `tracing::warn!` when the retry loop first engages and when it gives up (with
  retry count and elapsed time), so this is visible in `boot.log` next time.
  Subsequent stalls on the same stream drop to `DEBUG` so a sustained upload
  cannot bury the log under its own backpressure.

### Why a timer rather than a readiness wait

Registering the waker and returning `Poll::Pending` is the idiomatic move, and
it is wrong here. `tokio-vsock`'s `poll_write_priv` obtains a write-readiness
guard and then calls `write(2)`; a non-`WouldBlock` error does **not** clear that
readiness, so there is no guarantee of a fresh writability edge to wake on — the
wait can hang forever, while a bare retry-on-wake loop would busy-spin instead.
What clears the condition is the host draining the virtqueue, which costs
wall-clock time. The sleep-fixes / `yield_now`-does-not result above is the
direct evidence for that.

### Why this is not blunt throttling

Nothing is inserted into the healthy path. No rate limit, no chunking, no
unconditional sleep, and no allocation at all while writes are succeeding — the
fast path costs one `Option::is_none()` check. The backoff exists only while
`ENOBUFS` is actually being returned and unwinds the moment a write lands.

### Relationship to the libkrun >= 1.19.0 floor

`server.rs` already requires libkrun >= 1.19.0 because 1.18.1 had a
multi-descriptor TX-chain bug in its vsock device. This is a different problem
and a different kind of fix: the device is behaving *correctly* here — a full
queue is a legitimate state — but the socket reports it as a fatal error rather
than as backpressure. This adapter is a mitigation for a transport that
under-delivers backpressure. It does not work around a device defect and it does
not substitute for the version floor.

### Blast radius

Applied to the vsock listener's accepted stream only. The UDS path stays
byte-identical: `ENOBUFS` is not one of its failure modes, and narrowing the
blast radius matters more than uniformity here.

## Tests

Six unit tests against a mock writer with a scripted error sequence, in
`crates/minimald/src/backpressure.rs`. The adapter is generic over the inner
`AsyncRead + AsyncWrite` stream precisely so this is testable without a VM:

- N transient `ENOBUFS` then success — the write completes, and every attempt is
  accounted for.
- The backoff resets after a successful write — observed through the *cadence*
  of write attempts under `tokio::test(start_paused)`, not by reaching into
  internal state.
- A stall past the 5 s deadline propagates the original error unchanged.
- A non-`ENOBUFS` error (`EPIPE`) propagates immediately and is never retried.
- Reads pass through untouched.
- A read error is propagated even when it is `ENOBUFS` (reads are never
  retried).

## Verification: what was and was not done

**Done.** macOS cannot build `minimald` at all (it depends on `procfs`), so
everything below ran under `cross` + Docker against
`aarch64-unknown-linux-musl`, matching `just test-cross`:

```
$ CROSS_CONTAINER_OPTS="--env HOME=/tmp" cross test -p minimald --lib \
      --target aarch64-unknown-linux-musl --locked
running 151 tests
...
test backpressure::tests::a_non_enobufs_error_is_never_retried ... ok
test backpressure::tests::a_read_error_is_propagated_even_when_it_is_enobufs ... ok
test backpressure::tests::a_stall_past_the_deadline_propagates_the_original_error ... ok
test backpressure::tests::backoff_resets_after_a_successful_write ... ok
test backpressure::tests::reads_pass_through_untouched ... ok
test backpressure::tests::transient_enobufs_is_retried_until_the_write_lands ... ok

test result: ok. 151 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

The full `cross test -p minimald` (all targets, including the integration
harnesses) is green as well; the `netns_root_integration` cases stay `ignored`
as they do on `main` — they are gated on `MINIMALD_NETNS_TEST` and run in the
`ci-linux-native` netns job.

`cross clippy -p minimald --all-targets --target aarch64-unknown-linux-musl -- -D warnings`
is clean, and `cargo fmt --all --check` passes.

**Not done:** this has **not** been verified in-guest. Confirming the fix for
real requires rebuilding the guest initramfs and restarting the VM, and a live
VM was in use on the development host. No `minvmd`, `min` binary, initramfs or
daemon state was rebuilt, re-signed or restarted for this change. **CI and an
in-guest run are the gates.**

### Repro recipe for whoever verifies in-guest

The trigger is a payload that compresses well enough that zstd does *not*
self-throttle the client, yet not so well that the compressed stream is trivial
— about 3.5:1. The measured case was a 45 MiB slice of an ext4 image, which
`zstd -3` takes to 12,784,144 bytes:

```sh
mkdir -p /tmp/enobufs-repro && cd /tmp/enobufs-repro
# 45 MiB at ~3.5:1. An ext4 image slice is what was measured; any payload with
# a comparable ratio should do.
dd if=/path/to/some.ext4 of=slice.bin bs=1M count=45
zstd -c -3 slice.bin | wc -c   # expect ~12.8 MB; 0.11 MB or ~45 MB will NOT repro

min activate --loadout dev --name enobufs-repro --attach .
```

Before this change that fails ~0.60 s after the vsock connection is accepted,
with `Failed to upload project files: copying tar stream to channel: channel
closed` on the client and the `session ended with error … (os error 105)` line
above in `boot.log`.

After this change the upload should complete. If the queue does fill, `boot.log`
carries `vsock tx queue full; engaging ENOBUFS backoff` at `WARN` once per
connection and nothing further. A `vsock tx queue still full past the ENOBUFS
stall deadline` `WARN` would mean the 5 s bound was reached and the error was
propagated exactly as before — that is the signal to widen the bound rather than
assume the fix regressed.

Controls that should keep passing either way: a 45 MiB all-zero file (compresses
to ~0.11 MB) and 512 MB from `/dev/urandom` (incompressible, so zstd
self-throttles the client).

## Adjacent things noticed, deliberately not touched

- `Server::run`'s benign-hangup classification (`server.rs`) matches on the
  rendered error string (`"early eof"`, `"Broken pipe"`, `"Disconnected"`). It is
  log framing only and the existing comment says so, but it is the same class of
  fragility this PR is careful to avoid on the `ENOBUFS` match.
- The `ENOBUFS` tore down the *whole* SSH connection rather than just the upload
  channel. This PR stops the error being raised in the common case, but the
  blast-radius question — should one channel's write error kill the session? —
  is untouched and deserves its own issue.

> [!NOTE]
> ### Retry ENOBUFS errors on the guest vsock write path with exponential backoff
> - Adds `EnobufsBackpressure<S>` in [backpressure.rs](https://github.com/gominimal/minimal/pull/874/files#diff-0a70211243435ecd6e988b758927d8bae895f0fb15e45beac4b51f36baae6ccd), an `AsyncRead + AsyncWrite` adapter that retries writes failing with `ENOBUFS` using exponential backoff (100µs initial, 4ms max, 5s deadline before giving up).
> - Wraps accepted `VsockStream` connections with `EnobufsBackpressure` in [server.rs](https://github.com/gominimal/minimal/pull/874/files#diff-6194757015134b93b4dd82adff3e05ff12029271e611911d4c2ebfcdec5095b4) so all vsock writes automatically get this retry behavior.
> - Non-`ENOBUFS` errors and all reads pass through unchanged.
> - Tests use `#[tokio::test(start_paused = true)]` with the `test-util` Tokio feature to verify backoff timing without real delays.
> - Behavioral Change: vsock write calls that previously surfaced `ENOBUFS` immediately will now stall for up to 5s before propagating the error.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a48fe23.</sup>
> 
>

### DIFF
--- a/crates/minimald/Cargo.toml
+++ b/crates/minimald/Cargo.toml
@@ -83,7 +83,9 @@ tokio-rustls = { workspace = true, optional = true }
 # The two-namespace UC7 mesh integration test (`tests/mesh_uc7.rs`) drives a
 # tokio runtime and `setns(2)` directly.
 libc.workspace = true
-tokio.workspace = true
+# `test-util` (not in tokio's `full`) for `#[tokio::test(start_paused = true)]`:
+# lets the ENOBUFS backoff cadence and stall deadline elapse in virtual time.
+tokio = { workspace = true, features = ["test-util"] }
 # Parity test in `sessions/composables.rs` compares `SetupForPackages` output
 # to the composition-only path from a nickel-declared BuildSpec.
 decode.workspace = true
--- a/crates/minimald/src/backpressure.rs
+++ b/crates/minimald/src/backpressure.rs
@@ -1,0 +1,473 @@
+//! Backpressure for a transport that under-delivers it.
+//!
+//! A `write(2)` on a guest AF_VSOCK socket fails with `ENOBUFS` when the
+//! virtio TX virtqueue is full — the guest has queued more than the host-side
+//! VMM has drained. Unlike `EWOULDBLOCK` this is a hard error: `tokio-vsock`
+//! surfaces the raw errno as an [`io::Error`] that is not
+//! [`ErrorKind::WouldBlock`](std::io::ErrorKind::WouldBlock), so no layer above
+//! retries it. In minimald that error escapes through russh and kills the whole
+//! SSH connection, not just the channel that was writing, which is how a file
+//! upload turns into `channel closed` on the client.
+//!
+//! [`EnobufsBackpressure`] is the missing backpressure: it absorbs `ENOBUFS`
+//! on the write path, waits for the host to drain, and re-issues the write.
+//!
+//! # Why a timer and not a readiness wait
+//!
+//! The obvious fix — register the waker and return [`Poll::Pending`] until the
+//! socket is writable again — does not work here. `tokio-vsock` reaches this
+//! error *through* a satisfied write-readiness guard, and a non-`WouldBlock`
+//! error does not clear that readiness, so there is no guarantee of a fresh
+//! writability edge to wake on: the wait can hang forever, and a bare
+//! retry-on-wake loop would busy-spin instead. What actually clears the
+//! condition is the host draining the virtqueue, which takes wall-clock time.
+//! Measurement agrees: inserting a 1 µs sleep between client writes makes a
+//! 1.28 GB upload succeed, while [`tokio::task::yield_now`] in the same place
+//! does not. So the retry is driven by a timer.
+//!
+//! # Relationship to the libkrun version floor
+//!
+//! [`Listener`](crate::server::Listener)'s vsock impl requires libkrun >=
+//! 1.19.0 because 1.18.1 had a multi-descriptor TX-chain bug in its vsock
+//! device. This is a different problem and a different kind of fix: the device
+//! is behaving correctly here — a full queue is a legitimate state — but the
+//! socket reports it as a fatal error rather than as backpressure. This adapter
+//! restores the backpressure semantics the caller already assumes; it does not
+//! work around a device defect, and it does not substitute for the version
+//! floor.
+
+use std::io;
+use std::pin::Pin;
+use std::task::{Context, Poll, ready};
+use std::time::Duration;
+
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+use tokio::time::{Instant, Sleep};
+
+/// Delay before the first retry of an `ENOBUFS` write. Deliberately far below
+/// the drain time of a full queue: the common case is a queue that clears
+/// almost immediately, and a long first wait would cost throughput on every
+/// stall. Note that tokio's timer has millisecond granularity, so this is a
+/// floor of "one timer tick", not a literal 100 µs.
+const INITIAL_BACKOFF: Duration = Duration::from_micros(100);
+
+/// Ceiling for the exponential backoff. A queue that has not drained in a few
+/// milliseconds is not going to drain faster for being polled harder, and this
+/// bounds the wasted syscalls during a long stall.
+const MAX_BACKOFF: Duration = Duration::from_millis(4);
+
+/// How long a single contiguous run of `ENOBUFS` may last before the error is
+/// propagated to the caller. Past this the transport is not congested, it is
+/// broken, and reporting that is better than hanging the session forever.
+const MAX_STALL: Duration = Duration::from_secs(5);
+
+/// State of an in-progress `ENOBUFS` stall: one contiguous run of failed write
+/// attempts. Absent (`None` on the adapter) whenever the stream is healthy, so
+/// the fast path costs one `Option` check and allocates nothing.
+#[derive(Debug)]
+struct Stall {
+    /// The armed retry timer. Held here — not recreated per poll — so that
+    /// re-polling resumes the same wait instead of restarting it. `None` once
+    /// it has fired and the write has not yet been re-attempted.
+    timer: Option<Pin<Box<Sleep>>>,
+    /// Delay for the next backoff step, doubling up to [`MAX_BACKOFF`].
+    delay: Duration,
+    /// When this run of `ENOBUFS` began, for the [`MAX_STALL`] deadline.
+    started: Instant,
+    /// Retries issued in this run, for the logs.
+    retries: u32,
+}
+
+/// An [`AsyncRead`] + [`AsyncWrite`] adapter that treats `ENOBUFS` on the write
+/// path as backpressure rather than as a fatal error.
+///
+/// A write attempt that fails with `ENOBUFS` is retried after a short sleep,
+/// backing off exponentially from `INITIAL_BACKOFF` to `MAX_BACKOFF` and
+/// resetting once a write lands. A contiguous run of failures lasting longer
+/// than `MAX_STALL` gives up and returns the underlying error unchanged, so a
+/// genuinely wedged transport still surfaces instead of hanging.
+///
+/// Only `ENOBUFS` is retried. Every other error, and every read, passes through
+/// untouched, so a healthy stream behaves exactly like the stream it wraps.
+///
+/// Generic over the inner stream rather than tied to
+/// [`tokio_vsock::VsockStream`](https://docs.rs/tokio-vsock) so the retry state
+/// machine can be exercised against a mock writer in unit tests.
+///
+/// # Buffer safety
+///
+/// The adapter never consumes bytes it does not report and never retains the
+/// caller's buffer: `ENOBUFS` means the write moved zero bytes, so a retry is
+/// free to write whatever buffer the caller offers at that time. A caller is
+/// therefore allowed — as [`AsyncWrite`] permits — to poll again with a
+/// different buffer after a [`Poll::Pending`], and no partial write can be
+/// re-issued against mismatched data.
+#[derive(Debug)]
+pub struct EnobufsBackpressure<S> {
+    inner: S,
+    /// The current stall, if a write is presently backing off.
+    stall: Option<Stall>,
+    /// Stall runs seen on this stream, cumulative. Used to keep the log to one
+    /// `WARN` per stream (see [`Self::arm_backoff`]).
+    stalls: u64,
+}
+
+impl<S> EnobufsBackpressure<S> {
+    /// Wraps `inner`, retrying `ENOBUFS` on its write path.
+    pub fn new(inner: S) -> Self {
+        Self {
+            inner,
+            stall: None,
+            stalls: 0,
+        }
+    }
+
+    /// Records an `ENOBUFS` write attempt and arms the next retry timer.
+    ///
+    /// Returns `false` when this run of failures has outlived `MAX_STALL`,
+    /// meaning the caller must propagate the error instead of retrying.
+    fn arm_backoff(&mut self) -> bool {
+        let now = Instant::now();
+        // Taken and put back so the whole update runs on an owned value; a
+        // give-up simply declines to put it back, clearing the stall.
+        let (mut stall, first) = match self.stall.take() {
+            Some(stall) => (stall, false),
+            None => {
+                self.stalls += 1;
+                let stall = Stall {
+                    timer: None,
+                    delay: INITIAL_BACKOFF,
+                    started: now,
+                    retries: 0,
+                };
+                (stall, true)
+            }
+        };
+
+        let elapsed = now.saturating_duration_since(stall.started);
+        if elapsed >= MAX_STALL {
+            tracing::warn!(
+                retries = stall.retries,
+                elapsed_ms = elapsed.as_millis(),
+                stalls = self.stalls,
+                "vsock tx queue still full past the ENOBUFS stall deadline; propagating the error"
+            );
+            return false;
+        }
+
+        let delay = stall.delay;
+        stall.delay = (delay * 2).min(MAX_BACKOFF);
+        stall.retries += 1;
+        stall.timer = Some(Box::pin(tokio::time::sleep(delay)));
+        let retries = stall.retries;
+        let stalls = self.stalls;
+        self.stall = Some(stall);
+
+        if first {
+            // One `WARN` per stream, then `DEBUG`: a sustained upload can stall
+            // hundreds of times, and burying boot.log under it would defeat the
+            // point of logging this at all. The give-up warning above is not
+            // rate-limited — it is the one that reports a real failure.
+            if stalls == 1 {
+                tracing::warn!(
+                    retries,
+                    elapsed_ms = elapsed.as_millis(),
+                    delay_us = delay.as_micros(),
+                    "vsock tx queue full; engaging ENOBUFS backoff (further stalls log at DEBUG)"
+                );
+            } else {
+                tracing::debug!(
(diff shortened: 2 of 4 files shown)


twitchyliquid64 (approved):

norrietaylor:
closed as invalid

Issue #870, opened 2026-07-21
feat(rcache): fetch the cache index by pinned commit (immutable per-commit snapshots), with root-index fallback


bryan-minimal:
## Problem

`RemoteCache` fetches the cache index as a single **mutable** root object, `index.shisha` (`crates/rcache/src/remote.rs:210`). Consequences:

- **Unbounded**: the object is the whole catalog and grows with every publish (~2.4 MB today).
- **Mutable**: it cannot be safely served through an aggressive CDN/mirror tier — a cache that doesn't revalidate freezes the catalog and new artifacts go invisible. This is the one object blocking immutable-cacheable distribution of the index.
- **Wrong granularity**: a project pins an exact upstream commit (`locked_commit`), but reads whatever catalog is current — the index can drift ahead of the pinned commit.
- **Unverifiable**: the mutable object carries no signature; there is nothing stable to verify.

## What already exists server-side

The build infrastructure publishes, per commit on watched branches, three **immutable, create-only** objects under the repo's key prefix (`<host>/<owner>/<repo>/`):

| Object | Format | Purpose |
| --- | --- | --- |
| `<commit>.shisha` | **byte-identical to the root `index.shisha`** | the zero-format-change stepping stone |
| `<commit>.closure.shisha` | 68-byte records, `SpecHash(32) | flags(4) | sha256(32)`, canonical order | the bounded catalog-at-C; a Sigstore bundle (`.sigstore.json`) is published beside it |
| `<commit>.map.json` | JSON, name → `{spec_hash, sha256}` per arch | name-based lookup without a graph load |

## Proposal (phased; no writer lockstep at any point)

**Phase 1 — by-commit fetch, zero format change.**
The CLI knows the upstream repo URL and `locked_commit`. Construct `<repo-slug>/<locked_commit>.shisha`, fetch it **first**, parse with the existing `index_file` parser (same bytes), and **fall back to root `index.shisha` on 404** (commits predating the snapshots, or non-watched branches). Local index caching becomes naturally per-commit (immutable ⇒ cache forever).

**Phase 2 — switch to the bounded closure.**
Read `<commit>.closure.shisha` instead (~40× smaller, bounded): entries-only parse of the same record shape. Optionally verify the Sigstore bundle beside it against a configured trust policy before trusting the mapping — giving "the catalog at commit C, verified" without trusting the bucket, CDN, or mirror.

**Phase 3 — what this unlocks.**
[...]

Issue #879, opened 2026-07-21
rcache: materialize() hashes error bodies — transient non-200 from the mirror surfaces as HashMismatch, fails the build


bryan-minimal:
## Symptom

Dogfood run on #872 failed with:

```
ERROR materialize{span_name=socat}: rcache::remote: error=HashMismatch { want: "b4e1e876ac9f4816eb2f9107cc01ef6f30978c678aaba7d6a6dbdb47ae27b128", got: "bb2fb7b51f18478e3ccde6376dc39285b20472f0394e7034fd2e0bdf6a5716e2" }
```

https://github.com/gominimal/minimal/actions/runs/29867965238/job/88760839894

## Diagnosis

The artifact itself is fine:

- `b4e1e876….zst` on the mirror is stable (`last-modified` 4 days before the failure) and its bytes hash correctly to `b4e1e876…` when fetched now.
- The `got` value `bb2fb7b5…` does not correspond to any artifact (404 as a content key). It is the hash of whatever error/garbage body the mirror returned in place of the object during the failing request.

So this was a transient serving error at the HTTP mirror, reported as a scary "cache corruption" hash mismatch, and it hard-failed the whole build on a single blip. A re-run passes.

## Root cause

`RemoteCache::materialize` (`crates/rcache/src/remote.rs`) streams the response body straight into the hasher without checking the status code, and never retries:

```rust
let mut resp = self.backend.execute(req).await?;
// ...
while let Some(chunk) = resp.chunk().await? {
    w.write_all(&chunk)...
}
// hash compare -> HashMismatch
```

A 404/5xx error page from the mirror is hashed like artifact bytes. Contrast: the index fetch in `RemoteCache::new` does branch on `status_code()` (404 → empty index).

## Proposed fix

1. Check `resp.status_code()` before hashing. Non-200 → a distinct `Error::RemoteFetch { status }` so transient mirror errors are legible as such, instead of masquerading as corruption.
2. Retry once (small backoff) on non-200 and on a first-attempt hash mismatch — a truncated 200 body also presents as a mismatch. A mismatch that survives the retry stays a hard `HashMismatch` error, so real corruption still fails loudly.

The content check did its job here (refused bad bytes); this issue is about error legibility and resilience, not integrity.

bryan-minimal:
Implementation up in #931 (configurable retries, default 3, exponential backoff + status-check before hashing). Stacked on #912; the "Fixes" link and auto-close activate when it retargets to main after #912 lands.

PR #912, opened 2026-07-22
feat(rcache): read the index from per-commit snapshots (auto/pinned/root)

bryan-minimal:
Phase 1 of #870, shaped per the [config-enum suggestion](https://github.com/gominimal/minimal/issues/870#issuecomment-5039089733): the *policy* (which index object to read) is resolved centrally and handed to the fetch layer as concrete instructions.

## What

- **mfile** — `CacheConfig` enum (`GlobalIndex` / `CommitIndex` / `CommitIndexOnly`), resolved by `File::cache_config()` from `[cache] index_source` (`auto` | `pinned` | `root`), the upstream pin (`repo` + `locked_commit` -> `<host>/<owner>/<repo>/<commit>.shisha`, matching the publisher's keying), and an optional override. `auto` is the default and degrades to the root index when there's no pin.
- **rcache** — `IndexSource { Root, Snapshot }` threaded through the reader constructors; no policy in this layer.
  - Snapshot 404 => `Error::SnapshotMissing` (the caller decides on fallback) — never an empty index, and corrupt-but-present snapshots still error rather than fall back.
  - Local index cache: snapshots are immutable => cached copies never expire (warm runs skip the index fetch entirely); the root copy keeps the 5-minute window.
  - A snapshot's GCS generation never seeds `into_writer` — writers always refetch and compare-and-swap against the root object.
  - `new_over_https` / `new_with_gcs_bucket` are unchanged and always read root (writer-adjacent paths, where root is authoritative).
- **mctx** — the one construction site translates `CacheConfig` -> `IndexSource`, honours the `MINIMAL_INDEX_SOURCE` env override, implements auto-mode fallback, and logs provenance (`debug` for the chosen snapshot, `info` when falling back).

## Modes

| mode | reads | missing snapshot |
| --- | --- | --- |
| `root` | root `index.shisha` | n/a (kill-switch / control arm) |
| `auto` (default) | per-commit snapshot | falls back to root (logged) |
| `pinned` | per-commit snapshot only | hard error naming the object (testing mode) |

## Testing

- `mfile`: slug normalization across URL forms (https, `.git`, credentials, `ssh://`, scp-like), mode x pin resolution matrix, `[cache]` parsing + override precedence.
- `rcache`: snapshot fetch over real HTTP transport, snapshot-404 => `SnapshotMissing` (root present, proving no silent root read), snapshot-cache-never-expires vs root-expiry behavior.
- Rollout is a no-op until snapshots exist for a pin: `auto` falls back identically to today's read path.

> [!NOTE]
> ### Read remote cache index from per-commit snapshots in `RemoteCache`
> - Adds `IndexSource` enum (`Root` or `Snapshot { object }`) to `rcache` so `RemoteCache::new` can load either the global root index or a per-commit snapshot index object.
> - Adds `CacheSettings` and `IndexSourceMode` (`auto`, `pinned`, `root`) to `mfile`, plus a `cache_config` resolver that maps the project's `minimal.toml` `[cache]` config and upstream pin to the appropriate `IndexSource`.
> - `mctx` reads an optional `MINIMAL_INDEX_SOURCE` env-var override and calls `cache_config` to select the index source, passing it to `RemoteCache::new_any`; returns `SnapshotMissing` errors and optionally falls back to the root index.
> - Snapshot local cache copies never expire; root index copies use a short TTL. Index data is validated against whole-record boundaries and written atomically via a temp file.
> - Behavioral Change: `RemoteCache` now returns `SnapshotMissing` instead of an empty index when a per-commit snapshot object is absent, and non-2xx responses are now hard errors rather than parsed.
>
> 
> #### Changes since #912 opened
>
> - Added configuration-based index source selection with automatic fallback to `rcache` [394dc17]
> - Replaced manual index source selection in `mctx` with configuration-based approach [394dc17]
> - Added `mfile` dependency to `rcache` crate [394dc17]
> - Modified `MINIMAL_INDEX_SOURCE` environment variable parsing in `mctx::Context` cache configuration to return configuration error for invalid values [beb0969]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fa086ec.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added configurable remote cache index selection (global, pinned commit, root) with an environment override for index source resolution.
  * Added support for immutable per-commit cache snapshots.
* **Bug Fixes**
  * Missing snapshot objects now surface a dedicated error and do not fall back to the root index, preventing unintended cache behavior.
* **Documentation**
  * Expanded remote cache documentation to clarify how the resolved minimal index source determines which remote index object is read.

### DIFF
--- a/crates/mctx/src/lib.rs
+++ b/crates/mctx/src/lib.rs
@@ -15,9 +15,7 @@ use common::{SpecOrigin, Target};
 use google_cloud_storage::client::Storage as GcsStorage;
 use lcache::CacheBinProvider;
 use ot::OpTracker;
-use rcache::{
-    Error as RemoteError, IndexSource, RemoteBinProvider, RemoteCache, RemoteCacheWriter,
-};
+use rcache::{Error as RemoteError, RemoteBinProvider, RemoteCache, RemoteCacheWriter};
 
 mod error;
 pub use error::Error;
@@ -527,6 +525,9 @@ impl Context {
             None
         };
 
+        // Rollout lever: overrides the file setting while per-commit reads
+        // bed in. Retire (env var first, then likely the file setting) once
+        // snapshot reads are the settled default.
         let override_mode = match std::env::var("MINIMAL_INDEX_SOURCE") {
             Ok(v) => Some(
                 v.parse::<mfile::IndexSourceMode>()
@@ -534,37 +535,18 @@ impl Context {
             ),
             Err(_) => None,
         };
-        let (source, fallback) = match self
+        let config = self
             .mfile
             .cache_config(override_mode)
-            .map_err(RemoteError::Config)?
-        {
-            mfile::CacheConfig::GlobalIndex => (IndexSource::Root, false),
-            mfile::CacheConfig::CommitIndex { object } => (IndexSource::Snapshot { object }, true),
-            mfile::CacheConfig::CommitIndexOnly { object } => {
-                (IndexSource::Snapshot { object }, false)
-            }
-        };
-        if let IndexSource::Snapshot { object } = &source {
-            tracing::debug!("cache index: per-commit snapshot {object}");
-        }
-
-        let ot = self.daemon.config.ot.clone();
-        let res = RemoteCache::new_any(
-            url.clone(),
-            gcs_storage.clone(),
-            index_dir.clone(),
-            ot.clone(),
-            source,
+            .map_err(RemoteError::Config)?;
+        let res = RemoteCache::new_any_configured(
+            url,
+            gcs_storage,
+            index_dir,
+            self.daemon.config.ot.clone(),
+            &config,
         )
         .await;
-        let res = match res {
-            Err(RemoteError::SnapshotMissing { object }) if fallback => {
-                tracing::info!("per-commit index {object} not found; using root index");
-                RemoteCache::new_any(url, gcs_storage, index_dir, ot, IndexSource::Root).await
-            }
-            other => other,
-        };
         tracing::trace!("remote cache init took {:?}", start.elapsed());
         res
     }
--- a/crates/mctx/src/lib.rs
+++ b/crates/mctx/src/lib.rs
@@ -15,7 +15,9 @@ use common::{SpecOrigin, Target};
 use google_cloud_storage::client::Storage as GcsStorage;
 use lcache::CacheBinProvider;
 use ot::OpTracker;
-use rcache::{Error as RemoteError, RemoteBinProvider, RemoteCache, RemoteCacheWriter};
+use rcache::{
+    Error as RemoteError, IndexSource, RemoteBinProvider, RemoteCache, RemoteCacheWriter,
+};
 
 mod error;
 pub use error::Error;
@@ -481,7 +483,10 @@ impl Context {
     /// Builds and returns a remote cache reader for the configured cache
     /// location. [Config] has already resolved where artifacts come from — a GCS
     /// bucket (the default) or an HTTPS mirror (honouring
-    /// `MINIMAL_REMOTE_CACHE_URL`) — so this just wires up the matching backend.
+    /// `MINIMAL_REMOTE_CACHE_URL`) — and the minimal file resolves *which index
+    /// object* to read ([`mfile::File::cache_config`], overridable via
+    /// `MINIMAL_INDEX_SOURCE`), so this just wires up the matching backend and
+    /// follows those instructions.
     ///
     /// `auth` selects authenticated vs. anonymous GCS access (the buildbot / mip
     /// upload path reads authed; anonymous CI reads use the public path). It's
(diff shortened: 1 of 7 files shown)


bryan-minimal:
Both review findings addressed in 32c8a14b:

- **Corrupt local copy**: real, and actually subtler than described — the wire parser treats `UnexpectedEof` as clean end-of-stream, so a truncated copy wouldn't wedge with a parse error; it would *silently parse as a shorter index*, permanently, since snapshot copies never expire. Fixed with (a) atomic temp-file+rename writes (best-effort, logged), and (b) a load-time check that a local copy is a whole number of 68-byte records, falling through to a refetch on any bad copy. Truncation at an exact record boundary is indistinguishable in-format; the atomic write is the guard against truncation landing at all.
- **Filename collision**: snapshot local filenames now flatten the full object key (`github.com_gominimal_pkgs_<commit>.shisha`), so distinct snapshots never share a file.

New regression test covers the corrupt-copy refetch; the existing expiry test covers the flattened naming.

bryan-minimal:
CodeRabbit's outside-diff Major is valid and now fixed in 78e333ab: non-404 error statuses on the index fetch no longer fall through to body parsing. The empty-5xx-body case was the sharp edge: it parses as an empty index, and with this PR's local caching would also have been written to disk. Now surfaced as a backend error before parsing, with a regression test (500 + empty body => Error::Backend). Same bug class as the artifact-fetch path tracked in #879.

bryan-minimal:
Post-review sweep of the couplings around the changed code — three checks, all clean, documenting for reviewers:

1. **`minimal update`'s index invalidation** (`op/project/update.rs` → `invalidate_remote_index`, which deletes the local root-index copy): snapshots don't need it — a pin bump changes the snapshot's local cache filename, so the stale-copy problem structurally can't occur; the new pin's copy doesn't exist yet and is fetched fresh. The root-file deletion still covers `root`/fallback reads. No change needed.
2. **Read-path parity**: both production endpoints serve per-commit snapshots today — verified the snapshot for this repo's own pkgs pin (`c854d6b1`) returns 200 with identical sizes via anonymous GCS (`minimal-staging-cache`) and the HTTPS mirror. Also means the dogfood job starts exercising auto-mode's snapshot path (not just the fallback) as soon as this merges.
3. **Publisher/reader keying contract**: the object key is now derived independently on the publish side and here; drift would be silent (404 → auto falls back to root). Tracked in the publisher's repo to converge on `mfile::commit_index_object` after its next dependency bump.

bryan-minimal:
Macroscope's follow-up finding (remote.rs:347) is valid and fixed in e16695de: the whole-record check now applies to freshly fetched index bodies, not just local copies. It was right about the laundering: a truncated download parses as a shorter index, and the local write would re-serialize it as well-formed whole records — permanently passing the local check. Now rejected before parsing, with a regression test (complete response, mid-record body => InvalidData). Transport-level truncation usually fails the Content-Length check, but a mirror returning complete-but-wrong bodies is a failure mode we have observed, so the format-level check earns its keep.

crates/mctx/src/lib.rs
```diff
@@ -521,8 +526,45 @@ impl Context {
         } else {
             None
         };
-        let res =
-            RemoteCache::new_any(url, gcs_storage, index_dir, self.daemon.config.ot.clone()).await;
+
+        let override_mode = match std::env::var("MINIMAL_INDEX_SOURCE") {
```
twitchyliquid64, line 531: Do we plan to turn this env-var down once things are settled?

crates/mctx/src/lib.rs
```diff
@@ -521,8 +526,45 @@ impl Context {
...
+            Err(_) => None,
+        };
+        let (source, fallback) = match self
+            .mfile
+            .cache_config(override_mode)
+            .map_err(RemoteError::Config)?
+        {
+            mfile::CacheConfig::GlobalIndex => (IndexSource::Root, false),
+            mfile::CacheConfig::CommitIndex { object } => (IndexSource::Snapshot { object }, true),
+            mfile::CacheConfig::CommitIndexOnly { object } => {
```
twitchyliquid64: Can `RemoteCache::new` or `RemoteCache::new_any` consume `mfile::CacheConfig` and do this logic in there? Feels like a cleaner separation than in mctx as not everything goes through mctx.

crates/mctx/src/lib.rs
bryan-minimal: yeah, that was my idea

bryan-minimal: Yes — it's the rollout lever: `pinned` proves the snapshot path end-to-end and `root` is the kill-switch while this beds in. Marked it as such in 394dc177 (comment says retire the env var first, then likely the `[cache] index_source` file setting too, once snapshot reads are the settled default). Happy to file the retirement as a tracked follow-up once it's merged.

crates/mctx/src/lib.rs
bryan-minimal: Done in 394dc177: `RemoteCache::new_any_configured` consumes `mfile::CacheConfig` directly and owns the translation, the auto-mode fallback, and the provenance log lines — so any caller gets identical behavior. mctx shrinks to resolving the env override and handing over the config. Dependency-wise it's clean: rcache -> mfile adds nothing transitively (graph already depends on mfile). The primitive `new_any(..., IndexSource)` stays for callers that need explicit control (and the writer-adjacent constructors remain root-only). Two new tests cover the configured path: auto falls back when the snapshot is missing; pinned errors rather than falling back even when root could serve.

crates/mctx/src/lib.rs
twitchyliquid64: Sweet, just checking. Defs want to keep this logic pretty tight long term

twitchyliquid64 (approved):

crates/mctx/src/lib.rs
```diff
@@ -521,8 +524,29 @@ impl Context {
...
+        // Rollout lever: overrides the file setting while per-commit reads
+        // bed in. Retire (env var first, then likely the file setting) once
+        // snapshot reads are the settled default.
+        let override_mode = match std::env::var("MINIMAL_INDEX_SOURCE") {
+            Ok(v) => Some(
+                v.parse::<mfile::IndexSourceMode>()
+                    .map_err(|e| RemoteError::Config(format!("MINIMAL_INDEX_SOURCE: {e}")))?,
+            ),
+            Err(_) => None,
+        };
```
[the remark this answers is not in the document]
bryan-minimal: Fixed in beb0969e: only NotPresent means no override; a set-but-non-unicode value now surfaces as a Config error like any other bad value.

PR #920, opened 2026-07-22
fix(minimal): add idle-progress timeout to upload to prevent indefinite hang

### LINKED ISSUE #886: minimal: upload failure sometimes hangs the client indefinitely instead of erroring
## Summary

When a workspace upload fails, the client sometimes **hangs indefinitely** instead of returning an error. The daemon has already logged the failure and torn the connection down; the client never notices and never exits.

Observed four times during the [gominimal/minimal#869](https://github.com/gominimal/minimal/issues/869) investigation, at **~2 min, ~9 min, 13 min, and 75 s** — each killed by an external timeout, never by the client itself. The same command on the same fixture usually returns `error: Failed to upload project files: copying tar stream to channel: channel closed` within a few seconds, so this is an alternate outcome of the same failure, not a different one.

## Reproduction

Same fixture as [#869](https://github.com/gominimal/minimal/issues/869): a project whose single file compresses to ~12.78 MB, uploaded to a guest whose vsock connection is reset mid-transfer.

```
min activate --no-loadouts -n <name> <fixture>
```

Most runs error out; a minority hang. No timeout bounds the wait, so an interactive user sees `Uploading project files...` and nothing else, indefinitely.

## Guest side

The daemon logs the connection dying, at the same point it does in the erroring case:

```
WARN conn{conn=N transport="vsock"}: minimald::server: session ended with error
     error=Protocol error: No buffer space available (os error 105)
```

So the information needed to fail fast exists on the daemon side and reaches the transport; the client simply does not act on it.

## Likely area

`crates/minimal/src/file_upload.rs` / `crates/minimal/src/client.rs`. The upload streams a tar into an SSH channel via `tokio::io::copy` with no deadline anywhere in the path. Candidate mechanisms, unconfirmed:

- `stream_tar_zstd` spawns the tar builder and awaits `build.await` after the copy; if the copy's error path and the builder task interleave unluckily, one side can wait on a duplex whose peer is gone.
- russh's `ChannelTx` blocks on window availability. A channel whose peer died without a window adjustment leaves the writer parked with no waker.

Worth noting the client has **no timeout on the upload at all** — the transport-level `HANDSHAKE_TIMEOUT` (`client.rs`) bounds only connection setup, and there is no equivalent for the transfer.

## Impact

Worse than the failure it accompanies. An error is diagnosable; a silent hang looks like a slow upload, and in CI it burns the job's whole time budget. Independent of the [#869](https://github.com/gominimal/minimal/issues/869) root cause: once that is fixed the hang becomes rarer but the missing deadline remains.

## Suggested direction

A deadline (or an idle-progress timeout) on the upload, so a dead peer surfaces as an error rather than an indefinite wait — plus a check of whether the error path can deadlock against the tar builder task.

0chroma:
<summary>

When the SSH connection drops mid-upload, russh's `ChannelTx` can block on window availability indefinitely — the peer is gone and no window adjustment ever arrives. The `tokio::io::copy` call in `stream_tar_zstd` had no deadline, so the client hung forever showing "Uploading project files..." with no way to detect the dead peer.

## Root cause

`stream_tar_zstd` used `tokio::io::copy` to pump data from the tar builder pipe to the SSH channel writer. `tokio::io::copy` has no deadline. russh's `ChannelTx` blocks on window availability when the SSH receive window is full. If the peer dies without sending a window adjustment (e.g. a vsock reset mid-transfer), the writer parks with no waker and the copy never returns.

The post-upload drain loop in `upload_workspace_files` (`while let Some(msg) = channel.wait().await`) also had no timeout — a wedged transport that never delivers a `Close` would hang there too.

## Fix

- **`stream_tar_zstd`**: Replaced `tokio::io::copy` with a manual copy loop. Each direction (pipe read + channel write) gets a 30s idle-progress timeout. On timeout, the tar builder task is **aborted** rather than left detached.
- **`upload_workspace_files`**: Added a 120s idle-progress backstop on the `channel.wait()` drain loop — generous enough for a legitimate large unpack, but catches a wedged transport.

Closes #886.

</summary>

> [!NOTE]
> ### Add idle-progress timeouts to upload to prevent indefinite hangs
> - Adds a 30s `IDLE_TIMEOUT` to the pipe-through loop in [`file_upload.rs`](https://github.com/gominimal/minimal/pull/920/files#diff-b417966fca8aff7d83bfe7def43f0915d3cc2fcc1c3d02953c288fc85274bbf2), replacing `tokio::io::copy` with a manual read/write loop where each `rx.read` and `w.write_all` is individually wrapped in `tokio::time::timeout`.
> - Adds a 120s `DRAIN_IDLE_TIMEOUT` to the post-upload channel drain loop in [`client.rs`](https://github.com/gominimal/minimal/pull/920/files#diff-9e29b53f60b5cd0ae317db17efc685735a41fb9e23b33e834d06f331db09b952), replacing an unbounded `channel.wait()` loop with one that times out if no message arrives.
> - On timeout or write error, the pipe is drained into a sink so the tar builder can finish cleanly before returning an error, avoiding panics from dropping an unfinished builder.
> - Adds tests for the stalling writer case and a `CountingReader` adapter to verify read byte counts.
> - Behavioral Change: uploads that previously hung indefinitely when the peer stopped advancing the window now fail with a stall error after 30s (pipe) or 120s (drain).
>
> 
> #### Changes since #920 opened
>
> - Added idle-progress timeout mechanism to `stream_via_pipe` function to prevent indefinite hangs during upload [65c9764]
> - Added test case `stream_via_pipe_bails_when_builder_stalls` to validate timeout behavior [65c9764]
> - Reduced read buffer size in `stream_via_pipe` manual copy loop [65c9764]
> - Updated error messages in `stream_via_pipe` function [65c9764]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c87fb24.</sup>
> 
>
> 

## Summary by CodeRabbit

* **Bug Fixes**
  * Uploads now detect idle/stalled transfers and fail fast with a clear “upload stalled” error instead of hanging indefinitely.
  * Improved robustness of post-upload draining, including better interpretation of clean vs unexpected connection closures.
* **Tests**
  * Added regression coverage to ensure streaming errors occur promptly when the builder/writer stalls.
  * Added assertions for reliable byte-pull accounting during small, repeated reads.

### DIFF
--- a/crates/minimal/src/file_upload.rs
+++ b/crates/minimal/src/file_upload.rs
@@ -103,17 +103,13 @@ where
 
     const IDLE_TIMEOUT: Duration = Duration::from_secs(30);
 
-    // Drive build and copy concurrently. `build` runs on the
-    // caller task; the pipe read side pumps into `writer` on the
-    // same task via `tokio::join!` — no spawn, so no `'static`
-    // bound on W or on any of `build`'s captures.
-    //
-    // We can't use `try_join!` here: cancelling `build` mid-
-    // `append_data` would drop an unfinalized `async_tar::Builder`,
-    // which panics from its `Drop` impl. Instead, if `writer`
-    // errors or the idle-progress timeout fires, keep pulling
-    // bytes out of the pipe (into the void) so `build` can
-    // complete its `finish()` path and unwind normally (#886).
+    // Drive build and copy concurrently with `try_join!`. On error
+    // either side cancels the other: if the writer stalls, the build
+    // future is dropped — `TarZstArchive`'s `Drop` mem::forgets the
+    // inner `async_tar::Builder`, so no panic; if the build errors, the
+    // copy is dropped — the pipe and writer are cleaned up. Whichever
+    // side errors first surfaces, giving the root cause, not a
+    // secondary "broken pipe" from the other (#886).
     let build_fut = build(tx);
     let copy_fut = async {
         // Buffer the wire writes (#923): the encoder can emit many
@@ -126,7 +122,7 @@ where
         // stall on either the pipe read (tar builder stuck) or the
         // channel write (peer gone, window frozen) surfaces as an
         // error instead of an indefinite hang (#886).
-        let mut buf = [0u8; 64 * 1024];
+        let mut buf = [0u8; 8 * 1024];
         loop {
             match tokio::time::timeout(IDLE_TIMEOUT, rx.read(&mut buf)).await {
                 Ok(Ok(0)) => break,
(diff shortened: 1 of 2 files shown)


crates/minimal/src/file_upload.rs
```diff
@@ -103,27 +111,53 @@ where
...
+
+        // Manual copy with an idle-progress deadline instead of
+        // tokio::io::copy. Each direction gets its own timeout: a
+        // stall on either the pipe read (tar builder stuck) or the
+        // channel write (peer gone, window frozen) surfaces as an
+        // error instead of an indefinite hang (#886).
+        let mut buf = [0u8; 64 * 1024];
+        loop {
+            match tokio::time::timeout(IDLE_TIMEOUT, rx.read(&mut buf)).await {
+                Ok(Ok(0)) => break,
```
twitchyliquid64, line 128: is this how EOF comes through? I always thought it was an error of type EOF, but i googled it and yeah its a 0 byte read. Hmmm.

crates/minimal/src/file_upload.rs
```diff
@@ -103,27 +111,53 @@ where
...
-            // root cause we want to surface.
-            let _ = tokio::io::copy(&mut rx, &mut tokio::io::sink()).await;
-            return Err(e);
+
+        // Manual copy with an idle-progress deadline instead of
+        // tokio::io::copy. Each direction gets its own timeout: a
+        // stall on either the pipe read (tar builder stuck) or the
+        // channel write (peer gone, window frozen) surfaces as an
+        // error instead of an indefinite hang (#886).
+        let mut buf = [0u8; 64 * 1024];
```
twitchyliquid64: We can turn this down to 4k or 8k, as the BufWriter will only emit 4k blocks anyway.

twitchyliquid64 (approved):

crates/minimal/src/file_upload.rs
0chroma: yeahhh, it's expected behavior so it comes through via an `Ok(...)` Unexpected EOF would be `io::ErrorKind::UnexpectedEof`. As an aside I also kinda hate how much I'm doing these nested result types, it feels very cursed.

PR #925, opened 2026-07-22
revert(minimald): remove the 8 MiB guest vsock receive window

norrietaylor:
Reverts [#922](https://github.com/gominimal/minimal/pull/922) (`ff79c8c2`), which restored the 8 MiB guest vsock receive window that [#916](https://github.com/gominimal/minimal/pull/916) had removed.

The resulting `crates/minimald/src/main.rs` is **byte-identical to its post-#916 state** — a clean revert to a tree that was already green on `main`.

## Why remove it

The window was always headroom, never the fix. It works by keeping a normal upload out of the fragmented tail where the guest's skb-overhead ceiling trips, but it does not change the ratio between the byte ceiling and the skb ceiling — so a large enough upload or a slow enough reader still reaches it. Against the real repo (141 MB compressed) the failure reproduces with the window in place.

The fragmentation itself is fixed one layer down, in libkrun ([#921](https://github.com/gominimal/minimal/pull/921), approved): coalesce reads into full descriptors and wait briefly on EAGAIN rather than emit a packet too small to carry its own receive-queue accounting. Instrumentation showed 86% of the sub-skb packets come from the fill loop bailing on EAGAIN with descriptor space and credit both free — exactly what that patch addresses, and it measures 16/0 against 0/10.

With the real fix landing, the window is redundant complexity in minimald: an FFI `setsockopt` dance (`SO_VM_SOCKETS_BUFFER_MAX_SIZE` before `SO_VM_SOCKETS_BUFFER_SIZE`, read-back, best-effort logging) whose only purpose was to paper over the libkrun defect. Remove it and let #921 carry the fix.

Refs: #869

> [!NOTE]
> ### Remove 8 MiB guest vsock receive window from `minimald`
> Reverts the AF_VSOCK receive window size increase in [main.rs](https://github.com/gominimal/minimal/pull/925/files#diff-94e641147dfc527a957e34ac84f21d47f330c992b120da13441a31765c72f981). The `set_vsock_rx_window` helper, the `VSOCK_RX_WINDOW_BYTES` constant, and the call site in `async_main` are all removed. The vsock listener now uses kernel-default buffer sizes, and the associated debug/warn logs about window sizing are no longer emitted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2a10804.</sup>
> 
>

### DIFF
--- a/crates/minimald/src/main.rs
+++ b/crates/minimald/src/main.rs
@@ -21,27 +21,6 @@ use tokio_vsock::{VMADDR_CID_ANY, VsockAddr, VsockListener};
 #[cfg(target_os = "linux")]
 const DEFAULT_VSOCK_PORT_BASE: u32 = 2222;
 
-/// Receive window installed on the guest's vsock listener, in bytes.
-///
-/// `virtio_transport_inc_rx_pkt` refuses an incoming packet — resetting the
-/// connection and setting `sk_err` to `ENOBUFS` — when either the peer overruns
-/// its credit (`buf_used + len > buf_alloc`) or the queued socket-buffer
-/// overhead does (`(queue_len + 1) * SKB_TRUESIZE(0) > buf_alloc`). At the
-/// 256 KiB default window the second ceiling lands at 455 queued skbs on arm64,
-/// where `SKB_TRUESIZE(0)` is 576 bytes. libkrun clamps each read to the credit
-/// still outstanding, so under receiver backpressure its packets shrink to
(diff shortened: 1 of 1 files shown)


norrietaylor:
Superseded by #924, which reverted the 8 MiB window on main. Closing in favour of that. The real fix is the libkrun-side change (#921 merged; #926 proposes replacing it with packet-count backpressure).

PR #931, opened 2026-07-23
feat(rcache): retry transient artifact fetches with exponential backoff

### LINKED ISSUE #879: rcache: materialize() hashes error bodies — transient non-200 from the mirror surfaces as HashMismatch, fails the build

bryan-minimal:
Fixes #879. **Stacked on #912** (same `Error` enum and `[cache]` config section) — shows only its own commit once #912 merges and this retargets.

## What

The incident class: a single transient mirror error failed a whole build, presenting as `HashMismatch` because the error body was fed to the hasher — indistinguishable from real corruption.

- **Status check before hashing** — `materialize()` now rejects error statuses up front (`Error::Backend` via `error_for_status`, `Error::Fetch { status }` fallback). An error page never reaches the hasher, so `HashMismatch` again means what it says. The index fetch's status fallback switches to the same `Fetch` variant for consistency.
- **Retry with exponential backoff** — transient failures retry with a 500ms base delay, doubling per attempt (500ms/1s/2s at the default). Transient = transport errors, error statuses, and hash mismatches (a complete-but-wrong body, the observed mirror failure mode). `NotFound` (spec absent from the index) and local I/O errors are not retried. Extraction runs once, after a verified fetch.
- **Configurable, default 3** — `[cache] fetch_retries = N` in minimal.toml, `MINIMAL_FETCH_RETRIES` env override (resolved centrally in the same place as `index_source`), `RemoteCache::with_fetch_retries()` on the API. `0` disables retries. Each retry logs a warn line with attempt count and delay.

## Testing

Mock backend gained scripted per-URL response sequences (status + body) and a call counter; retry tests run under tokio's paused clock so backoff costs no wall time:

- 503 then wrong-body then success → succeeds, exactly 3 fetch attempts
- persistent wrong-body with `fetch_retries = 1` → `HashMismatch` after exactly 2 attempts (real corruption still fails loudly)
- error status with `fetch_retries = 0` → fails fast as a status error, not `HashMismatch`, proving the error page never reached the hasher
- absent spec → `NotFound` with zero fetch attempts

> [!NOTE]
> ### Retry transient artifact fetches in `RemoteCache` with exponential backoff
> - `RemoteCache::materialize` now retries failed artifact downloads on transient errors (transport failures, HTTP error status, hash mismatches) using exponential backoff with a 500ms base, up to a configurable retry count (default: 3).
> - Adds `RemoteCache::fetch_verified` to encapsulate per-attempt fetching, status checking, streaming to a temp file, and SHA-256 verification.
> - Adds per-commit snapshot index support via a new `IndexSource` enum; `new_any_configured` resolves the index source from `mfile::CacheConfig`, with automatic fallback to the root index when a snapshot is missing.
> - Retry count is configurable via `MINIMAL_FETCH_RETRIES` env var, the `[cache] fetch_retries` field in `minimal.toml`, or `RemoteCache::with_fetch_retries`.
> - Behavioral Change: HTTP error responses are no longer passed to the hasher and now surface as `Error::Fetch` instead of silently producing an invalid or empty index.
>
> 
> #### Changes since #931 opened
>
> - Refactored test mock infrastructure by introducing `ScriptedResponses` type alias and applying it to `MockBackend::scripted` field [483b242]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ed52664.</sup>
> 
>
>


norrietaylor (approved):

crates/rcache/src/remote.rs
```diff
@@ -400,48 +420,95 @@ impl<B: FetchBackend> RemoteCache<B> {
...
+        Ok(tar_file)
+    }
+
+    /// Failures worth retrying: transport errors, error statuses, and hash
+    /// mismatches (a complete-but-wrong body, e.g. from a misbehaving
+    /// mirror). Local I/O errors and a spec absent from the index are not.
+    fn is_transient(e: &Error<<B::Response as FetchResponse>::Error>) -> bool {
+        matches!(
+            e,
+            Error::Backend(_) | Error::Fetch { .. } | Error::HashMismatch { .. }
```
twitchyliquid64, line 524: If the status code was 200, do we want to consider HashMismatch a transient error?

twitchyliquid64 (approved):

crates/rcache/src/remote.rs
bryan-minimal: Yes, deliberately — a 200 carrying wrong bytes is exactly the observed incident that motivated this issue: the mirror served a complete response whose body hashed to garbage, and the same key served correct bytes minutes later. Transport-level truncation usually fails the Content-Length check, so complete-but-wrong is precisely the shape the mismatch retry is for.

The asymmetry that makes it safe: if the object is *persistently* corrupt, retrying costs ~3.5s of backoff and then surfaces the same loud HashMismatch (covered by the exhausts-retries test); if it's the transient mirror class, the retry saves the whole build. Sharpened the is_transient doc comment to record this reasoning in-code.

PR #934, opened 2026-07-23
feat(minimal): add `--provider local-minvmd`, remove `--minvmd`

### LINKED ISSUE #739: Use `--provider local-minvmd` instead of `--minvmd` flag for Linux
## Summary

For Linux, we should switch from the `--minvmd` flag to `--provider local-minvmd`. The `--provider local-minvmd` form is clearer and preferred.

It should apply for MacOS too, implicitly for now tho, since it's the only provider but specifying explicitly should not be an error.

## Proposed change

- Replace the `--minvmd` flag with `--provider local-minvmd` for Linux and Mac.

## Context

Requested via Slack discussion: "I like `--provider local-minvmd` better than `--minvmd`. Lets make the change."

norrietaylor:
## What

Replace the `min` CLI's opaque boolean `--minvmd` flag with a `--provider <PROVIDER>` global option:

- `--provider local-minimald` — host-native `minimald`, the default.
- `--provider local-minvmd` — `minimald` inside the minvmd microVM.

`--minvmd` is **removed outright** (it existed only transiently within this PR as a hidden alias). All callers now use `--provider local-minvmd`, including the three CI workflows, updated with codeowner approval. Backend selection is centralized behind `GlobalArgs::use_minvmd()`.

## Distinct per-backend provider identities

Previously both backends shared one provider-instance identity, `local-<N>` (e.g. `local-0`) — used for the on-disk `providers/<name>` dir, the SSH host-key identity in `known_hosts`, and the hostname the client verifies. That made native `minimald` and the `minvmd` VM collide (see #739). Now each backend gets a distinct identity:

- native minimald -> `local-minimald<N>`
- minvmd microVM  -> `local-minvmd<N>`

The name is produced in one place, `paths::provider_instance_name(kind, N)` (keyed on a new `paths::ProviderKind`). The client picks the backend's dir via `client_provider_kind(use_minvmd)` (minvmd when `--provider local-minvmd`, or always on macOS); `minimald`/`minvmd` each hard-code their own kind. The attach/forward SSH host alias is read back from the resolved provider dir so client and daemon can never disagree.

## Migration of existing on-disk instances

`paths::migrate_legacy_provider_dirs` renames any pre-split `providers/local-<N>` dir to the kind-tagged scheme so existing instances are not orphaned. Each dir's kind is inferred from its contents:

- minvmd markers: `minvmd.toml`, its alive lock, or `data-vol.raw`.
- native minimald markers: the on-disk `ssh_host_ed25519_key`, or its lock.
- both or neither present -> ambiguous, left untouched.

Best-effort and idempotent: only genuine I/O errors warn; benign skips (ambiguous/empty content, target already exists) are silent so the CLI does not repeat a warning every command. Invoked once where provider dirs are first resolved — native minimald startup, minvmd `run`, and the `min` client's command dispatch.

## Test

`cargo test` on `paths` (53, incl. migration cases) / `minimal` / `minvmd` (122) / `minimald` (166) passes; `minimal` integration `cli` (25) and `git_remote` (1) pass. `bug.rs` is 6 pass / 2 fail and `minimal` lib has 2 fails — all four are **pre-existing** environment failures (no network tooling; running as root), unrelated to this change. `clippy`/`fmt` clean for touched crates (remaining clippy warnings are pre-existing in `switch.rs`/`sandbox2`).

Closes #739

> [!NOTE]
> ### Replace `--minvmd` flag with `--provider {local-minimald|local-minvmd}` and rename provider directories to kind-tagged paths
> - Adds a `--provider` CLI flag (via a new `Provider` enum) replacing the deprecated `--minvmd` boolean, allowing explicit selection of `local-minimald` (host) or `local-minvmd` (VM) backends.
> - Renames provider instance directories from `providers/local-<N>` to `providers/local-minimald<N>` or `providers/local-minvmd<N>` depending on backend, affecting socket paths, lock files, known_hosts entries, and data volume paths.
> - Adds a best-effort, idempotent migration routine (`paths::migrate_legacy_provider_dirs`) that classifies and renames legacy `providers/local-<N>` directories to the new kind-tagged scheme on startup, both in `minimald` and `minvmd`.
> - SSH host aliases used in known_hosts are now derived from the provider instance directory name (e.g. `local-minimald0`) instead of being hardcoded as `local-0`.
> - Risk: On-disk provider directory names change for all existing installations; the migration routine handles this automatically but only when the kind can be inferred from directory contents.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8647606.</sup>
> 
>

### DIFF
--- a/.github/workflows/ci-linux-kvm.yml
+++ b/.github/workflows/ci-linux-kvm.yml
@@ -347,7 +347,7 @@ jobs:
         # The SAME unified proof every target lane runs (scripts/session-e2e.sh)
         # against Deployment Model 3 (native Linux + VM — the target this
         # lane owns; see docs/specs/03-spec-networking): from a clean state
-        # `minimal --minvmd activate` must auto-spawn minvmd, boot the microVM,
+        # `minimal --provider local-minvmd activate` must auto-spawn minvmd, boot the microVM,
         # and round-trip a session exec over the vsock bridge — the Linux/KVM
         # counterpart of the macOS lane's CLI gate. Since #748 activate uploads
         # the project into the guest; E2E_PROJECT_DIR=/tmp still names a guest
@@ -355,7 +355,7 @@ jobs:
         env:
           MINVMD_BOOT_LOG: ${{ runner.temp }}/cli-e2e-boot.log
           MINVMD_GVPROXY_BIN: ${{ runner.temp }}/gvproxy
-        run: E2E_VM=1 E2E_MINIMAL_ARGS=--minvmd E2E_PROJECT_DIR=/tmp ./scripts/session-e2e.sh
+        run: E2E_VM=1 E2E_MINIMAL_ARGS="--provider local-minvmd" E2E_PROJECT_DIR=/tmp ./scripts/session-e2e.sh
 
       - name: Upload guest boot console logs
         if: always()
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -132,7 +132,7 @@ jobs:
           MINVMD_INITRAMFS: ${{ runner.temp }}/initramfs.cpio
           MINVMD_GVPROXY_BIN: ${{ runner.temp }}/gvproxy
           E2E_VM: "1"
-          E2E_MINIMAL_ARGS: --minvmd
+          E2E_MINIMAL_ARGS: --provider local-minvmd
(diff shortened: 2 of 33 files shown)


twitchyliquid64 (approved):
SGTM, but given we have minimalists with existing sessions it might be good to rename `local-0` to the appropriate new name on `min` startup or something

PR #936, opened 2026-07-23
fix(minimald): bound diag stream on reader idleness, not total transfer

### LINKED ISSUE #915: minimald: diag STREAM_TIMEOUT caps the whole bundle transfer, not reader idleness
Follow-up to [gominimal/minimal#878](https://github.com/gominimal/minimal/pull/878), from CodeRabbit review comment [discussion_r3632311157](https://github.com/gominimal/minimal/pull/878#discussion_r3632311157). Validated against the merged code.

## Problem

In `crates/minimald/src/diag.rs`, `stream_diag_bundle` bounds the transfer with:

```rust
tokio::time::timeout(STREAM_TIMEOUT, tokio::io::copy(&mut rx, &mut writer)).await
```

`STREAM_TIMEOUT` (300s) is documented as "the bound on the peer rather than on ourselves" — but wrapping the whole `io::copy` caps **build + stream**, not reader idleness. The copy only reaches EOF when `build_bundle` finishes and drops its end of the duplex, so slow collectors count against the client's clock.

`build_bundle` runs **15** sequential `collect_step!` collectors, each budgeted up to `COLLECTOR_TIMEOUT` (30s), so a legitimate worst case is **450s > 300s**. A wedged host — the exact scenario a diagnostic bundle exists for — is when collectors run long, and it's also when the bundle gets truncated.

Two defects:

1. A slow-but-progressing bundle is cut off mid-stream even while the client is actively reading.
2. On expiry the error unconditionally blames the peer: `"client stopped reading the bundle for {STREAM_TIMEOUT:?}"` — misleading when the stall was server-side collection.

## Fix directions (from the review)

- Make the bound idle-based: a progress watchdog on the copy (reset on each write) instead of a whole-transfer deadline, or
- Keep the whole-transfer form but derive the ceiling from the collector budget (`n_collectors × COLLECTOR_TIMEOUT` + streaming margin), and
- Reword the expiry error so it doesn't assert which side stalled.

Severity: real but edge-case (needs ≥10 collectors wedged to trip today); CodeRabbit rated it Major because it fails precisely in the wedged-host scenario the subsystem targets.

norrietaylor:
## What

Replace the diag bundle pump's whole-transfer `STREAM_TIMEOUT` with an
idleness bound. `serve_stream_diag_bundle` previously wrapped
`tokio::io::copy` in a single 300s `tokio::time::timeout`, so the timeout
capped the *entire* transfer rather than reader inactivity.

`build_bundle` runs up to fifteen collectors back to back, each deadlined
individually at `COLLECTOR_TIMEOUT` (30s) — a worst case of 450s. On a
wedged host (the exact host a diagnostic bundle is wanted from), a bundle
the client is actively reading could outlast the fixed 300s total and be
cut off mid-stream, delivering a truncated archive. The expiry error also
blamed the client (`"client stopped reading the bundle …"`) regardless of
whether the stall was server-side collection.

## How

Add `pump_until_idle`, a manual read/write copy loop whose deadline resets
on every chunk that moves and fires only when a single read-or-write step
makes no progress for `STREAM_TIMEOUT`. A slow-but-progressing transfer now
runs as long as bytes keep flowing; a genuine stall — a client that stops
reading, or a build that emits nothing at all for the window — is still
bounded. The drop-order and cancellation invariants around `rx`/`writer`
that the original relied on are preserved. Reword the expiry error to name
the stall without asserting which side caused it.

Progress is measured on the compressed bytes reaching the channel; the doc
comments state this honestly rather than claiming total decoupling from the
collection budget.

## Tests

Two unit tests cover the new semantics: a slow-but-progressing transfer
whose total duration exceeds the idle bound completes without a timeout,
and a stalled reader trips a `TimedOut` error whose message asserts no
blame. `cargo test -p minimald` passes (20 diag tests); `cargo fmt --check`
and `cargo clippy -p minimald` are clean for the touched file.

Closes #915

> [!NOTE]
> ### Bound `minimald` diagnostic stream timeout on reader idleness rather than total transfer duration
> - Replaces `tokio::time::timeout` wrapping `tokio::io::copy` in `stream_diag_bundle` with a new `pump_until_idle` utility that resets the deadline on each chunk of progress.
> - Transfers that keep making progress can now run indefinitely; the timeout only fires after `STREAM_TIMEOUT` (300s) of inactivity on either side.
> - Timeout error messages now report lack of progress neutrally instead of blaming the client.
> - Two unit tests are added to [diag.rs](https://github.com/gominimal/minimal/pull/936/files#diff-ed60839e8685efd73f65ba91318d4d49927b99b4dbd27171a941fc02fecd5a34): one verifying slow-but-progressing transfers complete, one verifying stalled readers time out with `ErrorKind::TimedOut`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7c6f752.</sup>
> 
>

### DIFF
--- a/crates/minimald/src/diag.rs
+++ b/crates/minimald/src/diag.rs
@@ -61,11 +61,23 @@ const MAX_SESSION_FILE_BYTES: u64 = 8 * 1024 * 1024;
 const LOG_FILES_MAX: usize = 5;
 /// Entry cap for the recursive state-dir listing.
 const LISTING_MAX_ENTRIES: usize = 100_000;
-/// Ceiling on the whole client-facing transfer. Every collector is deadlined
-/// individually, but those deadlines only bound work — a client that holds the
-/// channel open and stops reading it parks the pump instead, and through the
-/// bounded duplex that stalls collection and finalization behind it. This is
-/// the bound on the peer rather than on ourselves.
+/// Idleness bound on the client-facing transfer: the longest the pump may make
+/// *no* progress before it is abandoned, not a ceiling on the whole transfer.
+///
+/// A whole-transfer bound is wrong here. Every collector is deadlined
+/// individually, but `build_bundle` runs up to fifteen of them back to back, so
+/// a legitimately slow-but-progressing bundle on a wedged host — the very host a
+/// diagnostic bundle is wanted from — can outlast any fixed total while the
+/// client reads every byte. What actually needs bounding is a stall: a client
+/// that holds the channel open and stops reading parks the pump, and through the
+/// bounded duplex that stalls collection and finalization behind it. So the
+/// deadline resets on every chunk that moves and only fires when nothing moves
+/// for this long.
+///
+/// Progress is measured on the compressed bytes reaching the channel, so this
+/// primarily bounds a client that has stopped reading, and as a backstop a build
+/// that emits nothing at all for this long — not the total collection time, which
+/// a healthy client is welcome to outlast as long as bytes keep flowing.
 const STREAM_TIMEOUT: Duration = Duration::from_secs(300);
 /// Duplex buffer between the bundle writer and the channel pump.
 const PIPE_BUF: usize = 256 * 1024;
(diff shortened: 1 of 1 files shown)


norrietaylor:
Superseded by #927, which landed the same idle-based diag stream timeout on `main` (commit 7fd582e6) while this was in flight. #927 bounds each write to the client on `STREAM_IDLE_TIMEOUT` (per-write, not whole-transfer) — the exact fix this PR targeted. Closing as duplicate; #915 is resolved by #927.

PR #863, opened 2026-07-21
docs(reference): add CLI command reference and config/spec pages

norrietaylor:
Fifth PR in the `oss/integration` decomposition — the CLI & config **reference** section (WS8, first half). Pure docs. **12 files, +779/−207.**

## What

- **Per-binary CLI reference:** `cli-min.md`, `cli-mip.md`, `cli-minimald.md`, `cli-minvmd.md`, plus the `cli.md` overview.
- **Config & spec pages:** `minimal-dot-toml.md`, `tasks.md`, `build-specs.md`, `sandbox-operations.md`, and the `harness-specs.md` → **`stack-specs.md`** rename.
- **`manifest.json`** — the reference-page index.

## Accuracy fix

Removed a stale "known issues" note in the mip reference that cited `dump --stacks` (`todo!()`) and `patched-build --remote-addr`. Both were **since fixed** (dump --stacks by #842) and their issues **closed**, so citing them as current known issues was inaccurate.

## Verification

- **Links:** every relative `.md`/`.json` link resolves; no danglers. Nothing in this set links to `loadouts.md`, so deferring it is clean.
- **Manifest:** valid JSON; all 11 listed pages are present in this PR.
- **Scrub:** no leaked emails/secrets/inbox refs.
- **Staleness handled:** the branch's `cli.md` rewrite already omits the `--hostpath-deps` row that #842 removed (no reversion); `harness-specs` has no lingering references.

Depends on nothing already merged; branches off current `main`. The loadouts page (367 lines) follows as PR6.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add CLI command reference pages and update config/spec docs for mip, min, minimald, and minvmd
> - Replaces the monolithic [cli.md](https://github.com/gominimal/minimal/pull/863/files#diff-975000db2bb150ebdc1a762c178c531c19747c5c75a8147bec4a31e69945e711) with an overview page that splits the session plane (`min`) from the package/build plane (`mip`) and links to new dedicated reference pages for each binary.
[...]

Issue #769, opened 2026-07-14
Upgrade Minimal while sessions are active breaks access to existing sessions


jessie-minimal:
related to gominimal/minimal#662

If Minimal sessions are running on the host when an upgrade is performed, ie uninstall and reinstall with the curl bash, it can break the user's access to existing sessions, and forcing users to destroy existing sessions. 

Options of expected behaviours
- Ask the user to terminate sessions before uninstall / upgrade can proceed
- on a straight upgrade without uninstall, the sessions should still work as before the upgrade

jessie-minimal:
For now, we should just have a message notifying the user that Minimal is shutting down.

Future options: snapshot VMs and resume. Though snapshots are something we hope to stay away from.

PR #941, opened 2026-07-23
docs(guide): lift the getting-started, install, and setup guides

norrietaylor:
Lift the Start pages (getting-started, install, setup) into docs/guide/, de-numbered and GitHub-first: em-dashes removed, harness renamed to stack ([stack] TOML block, verified against mfile), package/build commands mapped to mip (init/update/--version), and the install guide uses the stable channel (https://go.minimal.dev/stable) in every install command. No profiles content, no private repo names. Cross-links resolve once #863 and sibling docs-lift PRs merge; verification greps (em-dash, harness/profile/unstable, root-relative doc links) all clean.

> [!NOTE]
> ### Add getting-started, install, and setup guides to the Minimal docs
> Adds three new documentation pages to the guide section covering the full onboarding flow for Minimal.
>
> - [getting-started.md](https://github.com/gominimal/minimal/pull/941/files#diff-d11497d93c4e9a068e9d45153a3b33747c860d802b11d3f7cd9e5286916e0857) introduces what Minimal is and links to a download.
> - [install.md](https://github.com/gominimal/minimal/pull/941/files#diff-c749a8248e06dd9d3f555251b87a61d355c79c44f4df0123837be8da9838b7a4) covers system requirements and the curl-based installer for Linux (x86_64/aarch64) and macOS on Apple Silicon.
> - [setup.md](https://github.com/gominimal/minimal/pull/941/files#diff-54f34e030ab7aee172f5584498c6f09f0754a72b2dc37084ef2857e9280dce51) explains `min init`, the `minimal.toml` config structure, and how to enter a session with `min activate --attach .`.
>
> 
> #### Changes since #941 opened
>
> - Updated installation requirements documentation [a33e616]
> - Added installation lifecycle management documentation [a33e616]
> - Updated project initialization and configuration documentation [a33e616]
> - Updated session activation sandbox behavior description [a33e616]
> - Wrapped frontmatter description value in double quotes [a33e616]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 82ef622.</sup>
> 
>
>


docs/guide/setup.md
```diff
@@ -0,0 +1,62 @@
...
+# Generated by `minimal init`
+
+[upstream] # Source of software & tooling
+repo = "https://github.com/gominimal/pkgs"
+branch = "main"
+locked_commit = "9d36aa3d0fec09658fe3bbc676beeae49a89c53d"
+
+[stack]
+use = "rust"
+
```
jessie-minimal, line 29: question: why does [defaults] and [session] need to be separate? Don't both mean "base the new environment off this spec"?

docs/guide/install.md
```diff
@@ -0,0 +1,51 @@
...
+curl --proto "=https" --tlsv1.2 -fsSL https://go.minimal.dev/stable | sh
+```
+
+### Verify
+
+```shell
+min --version
+```
+
+If installed correctly, it will provide you with the version.
```
jessie-minimal, line 52: Should we document the upgrade and uninstall instructions here?

docs/guide/setup.md
```diff
@@ -0,0 +1,62 @@
...
+- `[upstream]` pins the software supply chain: the package registry and the exact commit it is read from.
+- `[stack]` declares how the project builds. See the [stack reference](../reference/stack-specs.md).
+- `[session]` is the one that matters for day-to-day work: it lists the tools every contributor's session gets. See the [`minimal.toml` reference](../reference/minimal-dot-toml.md) for the full schema.
+
+## Declare your session's tools
+
+Edit the `packages` list under `[session]` to add whatever your team works with. Extend the existing list; do not add a second `[session]` table, since TOML does not allow a table to be defined twice.
+
+```toml
+[session] # min attach
```
jessie-minimal, line 47: should the comment be `min activate` instead? I see this in the init toml file also

docs/guide/setup.md
norrietaylor: This is a great question! 

Right now we have both... if we add multiple session definitions in the same toml this would make sense, but I don't think that is part of the plan

docs/guide/install.md
norrietaylor: That's a good idea, I Don't think we have that anywhere else.

docs/guide/setup.md
norrietaylor: TBH; at a glance this comment doesn't seem very helpful

docs/guide/setup.md
norrietaylor: They are different planes rather than two specs for the same environment: `[defaults]` sets defaults for task/build sandboxes (`state_key`, `profile` — each task gets its own fresh sandbox), while `[session]` declares the one interactive session environment for the project. Addressed in a33e616c: the section list in setup.md now has a `[defaults]` bullet spelling out that split so readers don't conflate them. Whether the two tables should eventually merge if multi-session configs never happen is a schema question worth its own issue if you want to pursue it.

docs/guide/install.md
norrietaylor: Done in a33e616c — install.md now ends with Upgrade and Uninstall sections: re-running the installer upgrades in place (components are hash-checked, only changed ones re-download, atomic swap), and `... | sh -s -- --uninstall` removes everything, with `--force`, `--purge`, and `--dry-run` documented. Also added the caution that upgrading restarts the daemon and interrupts active sessions (#769), since we hit exactly that while testing.

twitchyliquid64 (approved):

Issue #749, opened 2026-07-14
chore: ADR — build-surface migration (rexec/remote-build to minimal CLI or builds-in-sessions)


0chroma:
## Context

The legacy `mip` CLI has `rexec`, `remote-build`, `run`, `build`, `test`, and `shell` commands. The new `minimal` CLI is session-only (activate/attach/destroy/ls/etc.) with no build or task execution surface. We need to decide how the build surface migrates.

**What exists today:**
- `rexec` / `remote-build` — in `mip`, hidden behind `MINIMAL_SCIENCE_MODE`, using the `remote-client` gRPC crate to talk to remote build servers
- `run <task>` / `build` / `test` / `shell` — in `mip` (Linux-gated), run tasks in a local sandbox
- `remote-client` crate — still a workspace member, consumed only by `mip`, not deprecated
- No ADR exists for this decision

**Options (to be fleshed out in the ADR):**
- **Option A**: Keep remote build servers as a parallel path — port `rexec`/`remote-build` to `minimal`, users invoke them directly
- **Option B**: Builds happen inside sessions via `run <task>` — `rexec`/`remote-build` are dead, `remote-client` crate can be removed

## Blockers

Per @twitchyliquid64, there are pieces that will take more effort than we have time for before the 15th. This ADR is deferred to after the 15th.

## Stakeholders

- @bryan-minimal — owns build-servers, knows whether `remote_client` is still needed
- @twitchyliquid64 — sessions/daemon architecture, whether builds go through sessions
- @norrietaylor — `run <task>` (#161) is the session-side build execution path

## Deliverable

An ADR in `docs/decisions/` documenting the chosen approach, alternatives considered, and migration plan.

---

## Summary

Decide and document the architectural choice for how the build surface migrates from the legacy `mip` CLI to the new `minimal` CLI. Two options exist: (A) port `rexec`/`remote-build` into `minimal` directly, keeping remote build servers as a parallel execution path; or (B) route all builds through sessions via `run <task>`, retiring `rexec`/`remote-build` and allowing the `remote-client` crate to be removed.

## Problem

[...]

PR #960, opened 2026-07-24
feat(minimald): side-op subsystem, package builds

twitchyliquid64:
* Session `SideOp` subsystem and lifecycle plumbed into session actor
 * `min build [--verbose] [--rebuild] <pkgs>...` command in the session shell
 * `min build [--verbose] [--rebuild] <pkgs>...` command over an ssh exec request

> [!NOTE]
> ### Add `min build` side-op subsystem to `minimald` for in-session package builds
> - Adds a `session_sop` module implementing cancellable background build side-ops, with `BuildUpdate`/`BuildOutcome` event types and a `BuildSink` for streaming progress to callers.
> - Extends `SessionHandle` with `start_build()` and `patches_upload_lock()` via actor message passing; introduces `WeakSessionHandle` so in-sandbox `min` commands can reference the owning session without creating a retain cycle.
> - Adds `min build [--verbose] [--rebuild] pkgs...` as an RPC command in [env.rs](https://github.com/gominimal/minimal/pull/960/files#diff-ece548260b2c5052569423263e9b54827b352ed496728bed2f1bf4c52c251dc1) and as an SSH exec command in [exec.rs](https://github.com/gominimal/minimal/pull/960/files#diff-d7bb305c5820ba4cc947bbf8a89b209dfdd48561cadd311f40fa0149e5ef7c8b), both streaming build output and returning structured exit codes.
> - Adds a `BuildRenderer` in the `orchestrator` crate to convert `BuildEvent` streams into attributed line output; reuses it in `mip` and both `minimald` build handlers.
> - `Session::stop_running` now tears down all active side-ops before stopping the host process, fixing a leak where background build tasks could outlive the session.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1c3174d.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added `min build` support for in-sandbox builds, including `--verbose` and `--rebuild`.
  * Stream build progress back to clients with hydration and “up-to-date”/status updates, with stderr rendering enabled only for `--verbose`.
  * Enhanced SSH `exec` handling to use `min <subcommand> ...` (including `min run ...` and `min build ...`), with clearer validation and routing.
* **Bug Fixes**
  * Improved reliability and user-facing error reporting when sessions end during patch upload and when build execution can’t be started (e.g., session gone).

### DIFF
--- a/crates/minimald/src/env.rs
+++ b/crates/minimald/src/env.rs
@@ -73,6 +73,10 @@ pub struct EnvArgs {
     network_mode: NetworkMode,
     own_ip_tap: Option<sandbox2::config::OwnIpTap>,
     own_ip_dns: Option<std::net::Ipv4Addr>,
+    /// Weak handle to the owning session actor, wired into the command channel
+    /// so in-sandbox `min` commands can drive session side-ops (e.g. builds).
+    /// Every session env has one — this `Env` is always session-scoped.
+    session: crate::session::WeakSessionHandle,
     /// When true (default), [`Env::build`] consumes package
     /// [`SetupForPackages`] output for `env_vars`/`fs_mappings`.
     /// When false, the caller supplies both via `env_vars` /
@@ -85,12 +89,14 @@ pub struct EnvArgs {
 
 impl EnvArgs {
     /// Creates args for an environment named `name`, with working directory
-    /// `cwd` and `/state` backed by `state_base_dir`.
+    /// `cwd` and `/state` backed by `state_base_dir`. `session` is a weak
+    /// handle to the owning session actor.
     pub fn new(
         name: impl Into<String>,
         cwd: impl Into<DaemonAbsPath>,
         home: impl Into<DaemonAbsPath>,
         state_base_dir: impl Into<DaemonAbsPath>,
+        session: crate::session::WeakSessionHandle,
     ) -> Self {
         Self {
             name: name.into(),
@@ -105,6 +111,7 @@ impl EnvArgs {
             network_mode: NetworkMode::HostNet,
             own_ip_tap: None,
             own_ip_dns: None,
+            session,
             include_package_attr_wiring: true,
         }
     }
@@ -367,6 +374,7 @@ impl Env {
             working: args.cwd.clone(),
             has_packages: transitives.keys().copied().collect(),
             ot: args.ot.clone(),
+            session: args.session.clone(),
             ctx,
             graph,
             rx,
@@ -499,6 +507,9 @@ struct SessionChannel {
     /// Packages already materialized into the rootfs.
     has_packages: HashSet<BuildSpecRef>,
     ot: Option<OpTracker>,
+    /// Weak handle to the owning session actor, used by session-scoped commands
+    /// (e.g. `min build`) to drive side-ops.
+    session: crate::session::WeakSessionHandle,
     rx: mpsc::Receiver<ChannelRequest>,
 }
 
@@ -571,6 +582,10 @@ impl SessionChannel {
             Some(("run", args)) => {
                 let (name, rest) = args.split_once(' ').unwrap_or((args, ""));
                 self.run_task(stream, name, rest).await;
+                None
+            }
+            Some(("build", args)) => {
+                self.run_build(stream, args).await;
                 None
             }
             _ => {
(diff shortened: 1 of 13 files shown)


crates/minimald/src/session_sop.rs
```diff
@@ -0,0 +1,168 @@
...
+                        }
+                    }
+                });
+
+                // Do the build.
+                if let Err(e) = ctx
+                    .build_graph_with_cancel(&graph, rebuild, Some(log_tx), cancel)
+                    .await
+                {
+                    tracing::warn!("session build failed: {e}");
```
bryan-minimal: **Build failure never leaves the daemon.**

Result is warn-logged and dropped, then `sinks.clear()` closes the stream identically for pass and fail — so neither caller can tell them apart:

- `run_build_exec`: `status` only becomes `1` on a *channel write* failure, so a real failure drains normally → **`exit_status(0)`**.
- `run_build`: unconditional `msg:build finished`, no `error:` terminator.

Also inverted: client disconnect → 1, broken build → 0.

Same class as `min check` exiting 0 on failing checkers (`mctx/src/env.rs`) — CI can only key on exit status, so a second instance is worth avoiding.

Suggest a terminal `BuildEvent::Finished { ok }` (or the result on `SideOp`) so both callers can set exit status / emit `error:`.

crates/minimald/src/env.rs
```diff
@@ -906,6 +921,45 @@ impl SessionChannel {
...
+        };
+
+        let mut renderer = orchestrator::BuildRenderer::new(flag_verbose);
+        while let Some(event) = events.recv().await {
+            if let Some(line) = renderer.render(event) {
+                let _ = writeln!(stream, "msg:{}", line.text);
+            }
+        }
+
+        let _ = writeln!(stream, "msg:build finished");
```
bryan-minimal: Companion to the `session_sop.rs` comment: this is unconditional, so a failed build still prints `build finished` and no `error:`. Worth noting this fn already emits `error:` for the session-gone and start-failure cases — the protocol supports it, the build's own outcome just isn't using it.

crates/minimald/src/exec.rs
```diff
@@ -807,33 +807,123 @@ pub(crate) async fn handle_exec(
...
+
+            spawn(async move {
+                let exec_task = ExecTask {
+                    conn,
+                    serv,
+                    session: session_handle,
+                    channel_id: id,
+                    exec: TaskExec {
+                        args: None,
+                        task: rem.strip_prefix("run ").unwrap().to_string(),
```
bryan-minimal: **Remotely-triggerable panic.** `rem` is everything after `min `, so argv `min run` (no task) gives `rem == "run"` → this arm matches → `strip_prefix("run ")` is `None` → `.unwrap()` panics.

Regression: the old `argv.strip_prefix("min run ")` returned `None` and fell through to a clean `channel_failure`. The `build` arm just below is already defensive (`.unwrap_or("")`) — so this is an inconsistency between the two arms. Reject empty/missing task before the `spawn`.

bryan-minimal (commented):
`min build` in the daemon is the #749/#751 direction landing — for our pkgs clean-room recipe it would drop the 29 MB `mip` upload entirely. Design reads well: actor-owned side-ops, `WeakSessionHandle` to avoid cycles, `graph_from_all_packages` on the blocking pool, cancellation tied to shutdown, non-blocking sinks, one shared `BuildRenderer`.

Two fixes inline. **#1 is the blocker: a failed build reports success on both surfaces.**

Minor: unknown `--flags` silently become package names; and I didn't see coverage for the new `run_build` / `run_build_exec` paths — the exit-status contract is worth pinning.

_Reviewed via [Claude Code](https://claude.com/claude-code)._

crates/minimald/src/session_sop.rs
twitchyliquid64: Fixed - new message `Finished` which holds the outcome, plumbed through.

crates/minimald/src/exec.rs
twitchyliquid64: Fixed

bryan-minimal (approved):
Both fixes verified at `1c3174dd` — and handled more thoroughly than I flagged. 👍

**Build outcome now propagates** via terminal `BuildUpdate::Finished(BuildOutcome)`:
- `run_build_exec`: `Success` → 0, `Failed` → 1 (+ stderr), `Cancelled` → 130, `None` → 1. Only success exits 0.
- `run_build`: only `Success` prints `build finished`; `Failed`/`Cancelled`/`None` emit `error:`.

Two things I didn't ask for and like: the `None` arm (channel closed without a terminal update) fails closed rather than silently passing, and `Finished` is explicitly protected from being dropped under sink backpressure — which is where this class of fix usually leaks.

**Panic fixed**: empty/missing task is rejected before the ack, no `.unwrap()`.

Non-blocking nits, take or leave: still no coverage pinning the exit-status contract (that's the bit most likely to silently regress), and unknown `--flags` still fall through into the package list. CI has jobs pending — worth landing on green.

_Reviewed via [Claude Code](https://claude.com/claude-code)._

PR #961, opened 2026-07-24
fix(minvmd): sync time for guests on MacOS

twitchyliquid64:
KVM has a paravirt clock that keeps everything in check, so no issue on Linux hosts. For MacOS hosts, libkrun has its own solution but the guest side only runs if you use `libkrunfw` / their init, which we dont.

So we do the same thing: listen on their magic vsock timekeeping port `123` when we are an arch that could be macOS, and update the guest clock when we get a periodic message or the time has drifted due to system power states.

## Summary by CodeRabbit

* **New Features**
  * Added optional guest time synchronization using host-provided timestamps.
  * The guest clock now corrects significant time drift automatically.
  * Added architecture-specific configuration to enable time synchronization where supported.

* **Bug Fixes**
  * Malformed updates and temporary communication interruptions are handled without stopping synchronization.
  * Permission failures when adjusting the clock are logged while allowing future updates to continue.

> [!NOTE]
> ### Sync guest CLOCK_REALTIME from host time updates over vsock in minimald
> - Adds `run_timekeep_listener` in [guest.rs](https://github.com/gominimal/minimal/pull/961/files#diff-f0dd3922eb27e86f3e4783c7469e18c89fd5165f3a119ef03217d32b6eb51296) that opens an AF_VSOCK datagram socket and receives 8-byte little-endian host nanosecond timestamps on the configured port.
> - Steps the guest `CLOCK_REALTIME` via `clock_settime` when drift exceeds 80ms; skips malformed packets and logs a one-time warning if `CAP_SYS_TIME` is missing.
> - On aarch64, [main.rs](https://github.com/gominimal/minimal/pull/961/files#diff-94e641147dfc527a957e34ac84f21d47f330c992b120da13441a31765c72f981) automatically spawns this listener on port 123 for init-process builds; other architectures leave it disabled unless `--timekeep-listener-port` is set.
> - Risk: `clock_settime` causes a discrete clock step (not a slew), which may affect time-sensitive applications running in the guest.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a27200f.</sup>
> 
>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,9 @@ lzma-rs = "0.3"
 nickel-lang-core = { git = "https://github.com/nickel-lang/nickel.git", rev = "e8a0ab95e52ab31ac9a8ae1a3bef854d8de73cea", default-features = false, features = ["format"] }
 
 opentelemetry-semantic-conventions = "0.32"
-nix = { version = "0.31", features = ["fs", "user"] }
+# `socket` + `time`: the guest's vsock timekeep listener binds an AF_VSOCK
+# datagram socket and steps CLOCK_REALTIME from the host's updates.
+nix = { version = "0.31", features = ["fs", "socket", "time", "user"] }
 nutype = { version = "0.7", features = ["serde"] }
 object = { version = "0.39", default-features = false, features = ["archive", "elf", "read_core", "std"] }
 oci-spec = "0.10"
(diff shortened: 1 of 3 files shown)


msample (approved):
Nice. Where did the 80ms threshold come from?

twitchyliquid64:
> Nice. Where did the 80ms threshold come from?

I pulled it out of the aether, basic logic was we needed to make it long enough that the delay is unlikely to be caused by vsock or scheduling latency, and 80ms is a solid 6.5 typical timeslices.

jtnkminimal (approved):

PR #981, opened 2026-07-27
fix(det-init): a hung sample must not discard the whole event feed

bryan-minimal:
The feed is drained **after** the sample returns (step 8), so a sample that never returns takes the **entire feed** with it: the host reaps the VM at its timeout, `drain_monitor` and `emit_feed_to_stdout` never run, and a run that should have been `BLOCK` is reported as `substrate_failed: empty observer capture`. The observer had captured everything — honeytoken reads included — and simply never shipped it.

Found by a corpus run, not by inspection. Two executions of the **same** poisoned sample, differing only in whether the C2 connect was bounded:

| C2 connect | Result |
|---|---|
| unbounded (`/dev/tcp` to a blackholed IP) | `ERROR` — empty capture |
| bounded (`timeout 5 curl --max-time 4`) | `BLOCK`, complete, signed |

**Not an exotic input.** A poisoned package whose C2 is unreachable — sinkholed, taken down, firewalled — blocks on `connect()` exactly like this. The perverse consequence: *the better the network defense, the more likely the whole feed is lost*, because the hang is caused by the connection **not** succeeding. It's also a one-line evasion — append a sleep and `BLOCK` degrades to `ERROR` after the payload has run.

## Fix

Bounds the sample at `SAMPLE_DEADLINE` (90s, under the host's 120s VM timeout) via `try_wait` polling, SIGKILLs on expiry, reaps, and falls through to the normal drain. SIGKILL rather than SIGTERM because this is untrusted code that already declined to finish and can trap the softer signal into the same hang.

Killing at the deadline costs only what the sample would have done *after* it; keeping the events it already produced is strictly better than losing all of them. The timeout is logged on the console the host collects, so a hang is distinguishable in triage from a real boot/attach failure — the old message blamed the substrate for something it did correctly.

## Verification status — please read

**This is not compile-verified.** `det-init` is Linux-only (`procfs`) with C deps, so it can't be built from macOS without a cross-toolchain. `rustfmt` parses it cleanly and the region is fmt-clean, but that is syntax, not typecheck.

It also **can't be validated by a golden-image bake as things stand**: the bake builds det-init from the pinned substrate tarball (`source-root-detinit-b34e1be1.tgz`), which predates this commit. Picking it up needs a re-stage on a Linux box with `minimal`.

## Not addressed here, deliberately

A timed-out run yields a **partial** feed, which arguably should cap the verdict at REVIEW as a coverage-honesty signal rather than producing a confident verdict over truncated evidence. That needs an event/schema change and belongs in its own PR.

Refs gominimal/minimal-detonation#272

> [!NOTE]
> ### Add `det-init` microVM init that emits the event feed even when the sample hangs
> Introduces the [`det-init`](https://github.com/gominimal/minimal/pull/981/files#diff-f322ecac9a0fe2e76da1d909c7e8a7c15c8eb44dd4917e084f1d85ca02765fee) binary, which runs as pid-1 inside a detonation microVM and orchestrates the full sample execution lifecycle.
>
> - Mounts `/dev`, securityfs, and bpffs; raises `RLIMIT_MEMLOCK`; then starts a root BPF-LSM observer and waits for its readiness signal before running the sample
> - Enforces a 90-second sample deadline with `SIGKILL` so a hung sample does not stall teardown — the event feed is always drained and emitted to stdout afterward via `emit_feed_to_stdout`
> - Runs the untrusted sample as uid 65534 (`nobody`) via the `droppriv` helper, with nonce- and uid-boundary self-checks and optional DNS rerouting to a local collector
> - Shuts down the guest VM via `minimald::guest::shut_down_vm` on completion when running as pid-1; parks if shutdown fails to avoid a kernel panic
> - Behavioral Change: a hung or killed sample no longer causes the JSONL event feed to be discarded — `drain_monitor` bounds the observer shutdown with a `SIGKILL` backstop before the feed is read
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a0cf81.</sup>
> 
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1581,6 +1581,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "det-init"
+version = "0.5.0-rc1"
+dependencies = [
+ "libc",
+ "minimald",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "diagnostics"
 version = "0.0.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/checkouts",
     "crates/common",
     "crates/decode",
+    "crates/det-init",
     "crates/diagnostics",
     "crates/graph",
     "crates/mctx",
--- a/crates/det-init/Cargo.toml
+++ b/crates/det-init/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "det-init"
-version = "0.0.1"
+version.workspace = true
 edition.workspace = true
 publish.workspace = true
 
@@ -17,7 +17,12 @@ name = "det-init"
 path = "src/main.rs"
 
 [dependencies]
-minimald = { version = "0.0.1", path = "../minimald" }
+# Path dep with NO version requirement, deliberately. Pinning `version = "0.0.1"` here broke the
+# build the moment the workspace bumped to 0.5.0-rc1 — a self-inflicted drift trap that has nothing
+# to do with API compatibility (all six `guest::` fns we call were unchanged). Cargo resolves a
+# bare path dep to whatever is in the tree, so version bumps can never break us again; the API
+# surface is what actually matters and the compiler checks that.
+minimald = { path = "../minimald" }
 tokio.workspace = true
 libc.workspace = true
 tracing.workspace = true
--- a/crates/det-init/Cargo.toml
+++ b/crates/det-init/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "det-init"
+version = "0.0.1"
+edition.workspace = true
+publish.workspace = true
+
+# `det-init` is the minimal-detonation guest pid-1 (`/init`). It reuses
+# `minimald`'s public guest-boot library (mounts, rootfs entry, root egress,
+# VM shutdown) instead of forking minimald, then stands up the in-guest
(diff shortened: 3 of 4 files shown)


bryan-minimal:
**Now verified — undrafted.**

Compiled on Linux via Cloud Build (`rust:1.97-bookworm`, matching the repo's pinned toolchain):

```
rustc 1.97.0 (2d8144b78 2026-07-07)
    Checking det-init v0.5.0-rc1 (/workspace/crates/det-init)
    Finished `dev` profile ... in 7m 08s
```

Two warnings, both pre-existing dead-code notes on `MainError`, unrelated to this change.

**The build also caught a latent bug that inspection had missed:** `det-init` was added to the workspace members but **never entered `Cargo.lock`**, so any `--locked` build failed outright with `cannot update the lock file`. That's invisible to a non-locked local build, which resolves it on the fly. Fixed in the follow-up commit — it would have broken CI here regardless of this PR.

Which is the argument for compiling rather than reasoning: I'd have shipped a PR that couldn't build under `--locked`, for a reason entirely unrelated to the change I was making.

bryan-minimal:
Superseded by #983 — same change, squashed to one commit.

This PR was genuinely hard to read and shouldn't have been opened as-is: six commits introducing one new crate, including fixes to its own earlier commits, plus two headers over commitlint's 100-char limit (pre-existing, from the July prototype work).

#983 is the same content as a single pure addition, with the Linux `cargo check --locked` result and the live PASS/BLOCK detonation stated up front.

PR #983, opened 2026-07-27
feat(det-init): guest pid-1 that detonates untrusted packages under a BPF-LSM observer

bryan-minimal:
Adds `det-init`, the minimal-detonation guest `/init`. It **embeds** `minimald::guest` rather than forking minimald, then stands up an in-guest BPF-LSM observer and runs an untrusted package install as `nobody` under it, streaming the event feed out for a host-side verdict.

It deliberately does **not** run minimald's SSH server — a detonation is one-shot, so the sample runs directly and the guest is discarded.

**Pure addition:** one new crate (+664 lines), plus its `Cargo.toml`/`Cargo.lock` workspace entries. No existing file is modified.

## Three load-bearing trust properties, each self-checked at boot

| Property | How it's proven |
|---|---|
| `nobody` cannot write the feed | sample stdio is **nulled**, not inherited — otherwise it could print forged `{"event":…}` lines onto the console the host scrapes |
| the binding nonce is root-only | leak-check asserts `nobody` gets EPERM |
| the event file is root-0400 | uid-boundary self-check |

## The sample is bounded

The feed is drained **after** the sample returns, so an unbounded sample that never returns takes the entire feed with it — the host reaps the VM, the drain never runs, and a detonation that should have been BLOCK reports as an empty capture.

Not an exotic input: a poisoned package whose C2 is unreachable blocks on `connect()` exactly like this, so *the better the network defense, the more likely the whole feed is lost*. On expiry the sample is SIGKILLed and the events it already produced drain normally.

## Verification

- `cargo check -p det-init --locked` on Linux, rustc 1.97.0 (the repo's pinned toolchain) — clean
- Live detonation on x86_64/KVM: 16/16 LSM hooks + 3/3 tracepoints attach; clean `npm install` → **PASS**; poisoned sample (honeytoken reads + non-registry beacon) → **BLOCK**, signed

Supersedes #981, which carried six commits — including fixes to its own earlier commits and two headers over commitlint's 100-char limit. Squashed to one, since this introduces a single new crate and that history was noise to review.

Refs gominimal/minimal-detonation#194, gominimal/minimal-detonation#272

> [!NOTE]
> ### Add `det-init` guest pid-1 to detonate untrusted packages under a BPF-LSM observer
> - Introduces a new `det-init` binary crate ([Cargo.toml](https://github.com/gominimal/minimal/pull/983/files#diff-81a248dc8c4f239df90392c77711214f7992169c45e47e500d6382eff60275ed), [main.rs](https://github.com/gominimal/minimal/pull/983/files#diff-f322ecac9a0fe2e76da1d909c7e8a7c15c8eb44dd4917e084f1d85ca02765fee)) that runs as pid-1 inside the microVM and orchestrates the full detonation flow.
> - Mounts `/dev`, rootfs, securityfs, and bpffs; raises the memlock rlimit; then starts a root BPF-LSM observer (`detonation-monitor`) and waits up to 10s for it to signal readiness.
> - Executes the untrusted sample as uid 65534 (`nobody`) via a `droppriv` helper with a 90s deadline and SIGKILL backstop; emits the drained JSONL event feed to stdout for host collection.
> - Verifies security boundaries before sample launch: checks that `nobody` cannot read the binding nonce or write the event feed, and optionally repoints DNS to a local collector when egress is up.
> - Risk: as pid-1, any unhandled exit would cause a guest kernel panic; the process parks in an infinite sleep loop after `async_main` completes to prevent this.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb15b54.</sup>
> 
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1574,6 +1574,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "det-init"
+version = "0.5.0-rc1"
+dependencies = [
+ "libc",
+ "minimald",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "diagnostics"
 version = "0.0.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/checkouts",
     "crates/common",
     "crates/decode",
+    "crates/det-init",
     "crates/diagnostics",
     "crates/graph",
     "crates/mctx",
--- a/crates/det-init/Cargo.toml
+++ b/crates/det-init/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "det-init"
+version.workspace = true
+edition.workspace = true
+publish.workspace = true
+
+# `det-init` is the minimal-detonation guest pid-1 (`/init`). It reuses
+# `minimald`'s public guest-boot library (mounts, rootfs entry, root egress,
+# VM shutdown) instead of forking minimald, then stands up the in-guest
+# BPF-LSM observer and runs the untrusted sample as `nobody` under it. It does
+# NOT run minimald's SSH server: a detonation is a one-shot, so the sample is
+# executed directly and the observer streams the event feed out. See
+# gominimal/minimal-detonation#194.
+
+[[bin]]
+name = "det-init"
+path = "src/main.rs"
+
+[dependencies]
+# Path dep with NO version requirement, deliberately. Pinning `version = "0.0.1"` here broke the
+# build the moment the workspace bumped to 0.5.0-rc1 — a self-inflicted drift trap that has nothing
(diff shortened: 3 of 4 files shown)


bryan-minimal:
Closing — this should not have been opened.

Landing det-init on minimal's main is **gated on a minimal-team sync** (observer position + untrusted-package policy), per the re-architecture decision tracked in gominimal/minimal-detonation#194. det-init deliberately lives on `det/det-init-prototype`, and the detonation substrate stages *from that branch* rather than from minimal's main — so nothing was blocked on merging it.

I opened this (and #981) without checking that gate. The CI churn on those PRs was the visible problem; the real one was proposing an architectural change to a shared repo that hasn't had its design conversation yet.

The det-init work itself is unaffected and stays on the branch: the hang fix (gominimal/minimal-detonation#272), the removed `minimald` version pin, and the missing `Cargo.lock` entry are all on `det/det-init-v2`. When the sync happens, that branch is what to propose from.

Issue #980, opened 2026-07-27
release: Linux install omits minvmd and gvproxy (own-ip dead on installs, no VM mode)


norrietaylor:
## Problem

The Linux install ships only `bin/min`, `bin/mip`, `bin/minimald`, the `git-remote-min` symlink, and the AppArmor files (`stage-release.sh` `COMPONENTS`). Two consequences:

- **own-ip is dead on every installed Linux host.** The native (DM2) own-ip datapath is implemented ([#542](https://github.com/gominimal/minimal/issues/542)) and proven by `netns_root_integration.rs` in the `ci-linux-native` netns job — but the daemon resolves the switch at `/usr/lib/minimal/bin/gvproxy` (`crates/minimald/src/server.rs`), and no Linux install has one. CI only passes because it fetches gvproxy itself and sets `GVPROXY_BIN`.
- **No VM-backed sessions on Linux installs.** `minvmd` ships on macOS only, so the KVM deployment mode is unavailable outside a dev checkout.

## What the release already produces

Most artifacts exist in `release.yml` today; they are staged but never mapped to Linux install components:

- `minvmd-linux-amd64` (built against libkrun via `setup-libkrun-linux`)
- `gvproxy-linux-amd64` / `gvproxy-linux-arm64` (pin-verified fetch)
- `vmlinuz-{amd64,arm64}`, `rootfs-{amd64,arm64}.img`, and the initramfs (packed from the `minimald-linux-*` binaries)

## Proposed change

Add Linux rows to the `COMPONENTS` table in `scripts/stage-release.sh`:

```
"gvproxy|linux|amd64|file|bin/gvproxy|gvproxy-linux-amd64"
"gvproxy|linux|arm64|file|bin/gvproxy|gvproxy-linux-arm64"
"minvmd|linux|amd64|file|bin/minvmd|minvmd-linux-amd64"
"minvmd|linux|arm64|file|bin/minvmd|minvmd-linux-arm64"
```

The installer is manifest-driven, so it should pick these up without changes once the manifest carries the rows.

## Open questions

1. **minvmd-linux-arm64**: the release builds only the amd64 minvmd today; an arm64 build job (or cross build) is needed for the arm64 row.
[...]

PR #988, opened 2026-07-27
feat(release): ship the VM stack to Linux installs [blocked on pkgs#533]

norrietaylor:
> **Blocked, and being rebuilt.** Do not review the current diff — it ships a
> `minvmd` that [gominimal/pkgs#533](https://github.com/gominimal/pkgs/issues/533)
> has since shown we should not ship. See "Why this is parked" below.

Part of [#980](https://github.com/gominimal/minimal/issues/980).

## What already split off

| PR | Scope |
|---|---|
| [#994](https://github.com/gominimal/minimal/pull/994) | The switch binary: shipped to Linux (fixes own-ip on installs) and installed as `gvproxy-min` |
| [#999](https://github.com/gominimal/minimal/pull/999) | Hide `--network` / `--ingress` while own-ip is not usable from an install |

That leaves this PR as the VM stack: `minvmd` for both Linux arches, the libkrun it links, and the guest payload.

## Why this is parked

This PR ships `minvmd` as a **native-glibc binary with a dynamically-linked libkrun**, which is what forced everything else in it:

- `lib/libkrun.so.1` as a shipped component (~29 MB of libraries)
- `scripts/rewrite-linux-linkage.sh` + its test harness, to rewrite RUNPATHs for the shipped layout
- a `bin`/`lib` sibling constraint, and an installer warning when a host's prefixes diverge
- a **new glibc version floor** on Linux — the first time an install has had one, since `min`/`mip`/`minimald` are static musl

[gominimal/pkgs#533](https://github.com/gominimal/pkgs/issues/533) has now **proven `minvmd` can ship as a single static-musl binary**: it builds, links, boots a VM, and passes the session e2e — verified against a deliberate stub control, because `build.rs` silently falls back to a stub that links and runs fine and cannot boot anything.

Every item above then disappears. ~29 MB of shipped libraries becomes one ~8 MB binary, and the glibc floor goes with it. Landing the dynamic version first would mean shipping all of that to users and deleting it days later.

The unblocking discovery was @twitchyliquid64's, in review on this PR: `libkrunfw` is dead weight, because it exists to carry a bundled GPL-2 kernel and `minvmd` supplies its own. That mattered beyond one artifact — musl's static libc cannot `dlopen`, so "libkrun must `dlopen` libkrunfw" was the sole reason static musl looked unreachable. #533 confirmed it the strong way: with libkrunfw removed system-wide, the *dynamic* libkrun searched all five loader paths, got `ENOENT` on every one, and booted anyway.

## What comes back

Rebuilt on top of #994, with:

- `minvmd` built for `*-unknown-linux-musl`, statically linked
- a `scripts/build-libkrun-linux.sh` mirroring the macOS from-source build (same pin, same patch series, same trim), plus the archive merge/localize pass #533 worked out
- **no** `lib/` component, **no** RUNPATH rewriting, **no** sibling constraint
- the guest payload, which is only useful once `minvmd` ships

Outstanding before that lands: #533 verified on **aarch64** only, so x86_64 needs proving — which will happen in CI, since the release lane builds both arches.

## Worth keeping from the review here

Both reviews on this PR produced findings that survive the rewrite, and the fixes are already in this branch's history:

- the nightly `ldd` assertion could never match glibc's un-normalized `$ORIGIN/../lib` output (permanently red nightly)
- `pipefail` + `grep -q` silently masking the unresolved-libraries guard
- the pre-swap daemon stop never targeting a running `minvmd` on Linux
- the spec gaps in 07 (`lib` prefix, R5.5's second stop) and 02 ("libkrun is resolved from the system")

The spec text will need another pass once the static build lands, since it currently documents the shipped-`lib/` layout that is going away.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -222,13 +222,28 @@ jobs:
                   mv "$BIN/libkrunfw-linux-amd64.so" "$LIB/libkrunfw.so.5"
                   chmod +x "$BIN/min" "$BIN/minvmd" "$BIN/mip" "$BIN/gvproxy-min"
                   echo "$BIN" >> "$GITHUB_PATH"
-                  ldd "$BIN/minvmd"
-                  if ldd "$BIN/minvmd" | grep -q "not found"; then
-                    echo "::error::shipped minvmd has unresolved libraries in the installer layout" >&2
+                  # Capture once, then match on the string. Piping into `grep -q`
+                  # under `set -o pipefail` can report the pipeline as SIGPIPE
+                  # (141) when grep exits early, which would make the guard below
+                  # silently evaluate false and pass a genuinely broken binary.
+                  ldd_out="$(ldd "$BIN/minvmd")"
+                  printf '%s\n' "$ldd_out"
+                  case "$ldd_out" in
+                    *"not found"*)
+                      echo "::error::shipped minvmd has unresolved libraries in the installer layout" >&2
+                      exit 1 ;;
+                  esac
+                  # ld.so expands $ORIGIN but does NOT normalize the rest of the
+                  # entry, so it reports <bin>/../lib/libkrun.so.1 verbatim.
+                  # Comparing raw strings against "$LIB/libkrun.so.1" could never
+                  # match; canonicalize both sides instead.
+                  resolved="$(printf '%s\n' "$ldd_out" | awk '/libkrun\.so\.1 =>/ {print $3; exit}')"
+                  if [ -z "$resolved" ] || [ ! -e "$resolved" ]; then
+                    echo "::error::shipped minvmd did not resolve libkrun at all (RUNPATH wrong)" >&2
                     exit 1
                   fi
-                  if ! ldd "$BIN/minvmd" | grep -q "$LIB/libkrun.so.1"; then
-                    echo "::error::shipped minvmd did not resolve libkrun from $LIB (RUNPATH wrong)" >&2
+                  if [ "$(readlink -f "$resolved")" != "$(readlink -f "$LIB/libkrun.so.1")" ]; then
+                    echo "::error::shipped minvmd resolved libkrun from $resolved, not $LIB (RUNPATH wrong)" >&2
                     exit 1
                   fi
             - name: Point the microVM env at the shipped guest artifacts
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -141,10 +141,11 @@ jobs:
         # same proof ci-linux-kvm's `test-kvm` runs. Every moving part is a
         # downloaded release artifact: the native-glibc `minvmd`, the guest kernel
         # + ext4 rootfs + initramfs, the pinned `gvproxy-min` switch, and the
-        # shipped libkrun pair — laid out as a `lib/` sibling of `bin/` exactly as
-        # scripts/install.sh drops it, so minvmd resolves libkrun through its
-        # $ORIGIN/../lib RUNPATH and libkrun finds libkrunfw through its own
-        # $ORIGIN (mirrors what smoke-macos proves for the dylib). Deliberately
+        # shipped libkrun — laid out as a `lib/` sibling of `bin/` exactly as
+        # scripts/install.sh drops it, so minvmd resolves it through its
+        # $ORIGIN/../lib RUNPATH (mirrors what smoke-macos proves for the dylib;
+        # like macOS, no libkrunfw is shipped — minvmd brings its own kernel, so
+        # libkrun's bundled-GPL-kernel library is never loaded). Deliberately
         # NO LD_LIBRARY_PATH and no materialize step: this lane is what catches a
         # shipped libkrun whose RUNPATH is wrong, and either would mask exactly
         # that failure.
@@ -173,8 +174,7 @@ jobs:
                   sudo udevadm trigger --name-match=kvm
                   ls -l /dev/kvm
             - name: Download shipped linux-amd64 artifacts
-              # Pulls min, minvmd, mip, gvproxy and the libkrun/libkrunfw pair
-              # (plus minimald, unused on the host — it ships inside the
+              # Pulls min, minvmd, mip, gvproxy and libkrun (plus minimald, unused on the host — it ships inside the
               # initramfs). `*-linux-amd64` is a clean glob here: only darwin
               # artifacts carry the `temp-` prefix.
               uses: actions/download-artifact@v8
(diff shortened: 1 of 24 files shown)


crates/minimald/src/server.rs
```diff
@@ -61,12 +56,15 @@ pub struct Config {
...
+    /// Resolves the configured gvproxy binary path, falling back to the
+    /// installed location when unset: the user-local `bin/gvproxy-min` the
+    /// curl|sh installer stamps, else the system-wide path. The `GVPROXY_BIN`
+    /// env var is scoped to the `#[ignore]` netns proof and is never consulted
+    /// by the daemon.
     fn gvproxy_bin_path(&self) -> PathBuf {
         self.gvproxy_bin
             .clone()
-            .unwrap_or_else(|| PathBuf::from(DEFAULT_GVPROXY_BIN))
+            .unwrap_or_else(::switch::installed_gvproxy_bin)
```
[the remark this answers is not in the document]
norrietaylor: Declining this one — it's accurate about the mechanism but inconsistent with what this code path already does.

`ServerState::new` is not the only sync filesystem work on the async daemon-startup path. `build_russh_config` does `state.host_key().await`, which reaches `Config::host_key()` → `PrivateKey::read_openssh_file(path)` — a synchronous file **read**, and on `create_if_missing` a keygen plus a synchronous **write**. That is materially heavier than the single `stat()` that `installed_gvproxy_bin()` performs, and it is pre-existing, deliberate, and on the same startup path.

Wrapping one `Path::exists()` in `spawn_blocking` while a synchronous key read/generate sits beside it would add machinery without changing the property you're protecting: this path would still block a worker on filesystem I/O. If we want startup off the worker threads, the key handling is the thing to move, and that is its own change rather than something to smuggle into a release-packaging PR.

Happy to revisit if daemon startup latency shows up as a real problem — at which point it should cover the whole path, not just this call.

twitchyliquid64:
> libkrun.so.1 → $ORIGIN (finds its libkrunfw.so.5 sibling)

There is no `libkrunfw` as we bring our own kernel

.github/workflows/nightly.yml
```diff
@@ -197,32 +197,55 @@ jobs:
...
+                  LIB="$RUNNER_TEMP/lib"
+                  mkdir -p "$LIB"
                   mv "$BIN/minimal-linux-amd64" "$BIN/min"
                   mv "$BIN/minvmd-linux-amd64" "$BIN/minvmd"
                   mv "$BIN/mip-linux-amd64" "$BIN/mip"
-                  mv "$BIN/gvproxy-linux-amd64" "$BIN/gvproxy"
-                  chmod +x "$BIN/min" "$BIN/minvmd" "$BIN/mip" "$BIN/gvproxy"
+                  mv "$BIN/gvproxy-linux-amd64" "$BIN/gvproxy-min"
+                  mv "$BIN/libkrun-linux-amd64.so" "$LIB/libkrun.so.1"
+                  mv "$BIN/libkrunfw-linux-amd64.so" "$LIB/libkrunfw.so.5"
```
twitchyliquid64: We bring our own kernel so shouldnt need `libkrunfw` - its purpose is to contain the GPL-2 kernel

scripts/install.sh
```diff
@@ -160,6 +160,21 @@ resolve_prefix() {
 # replaces it.
 bindir="$(resolve_prefix bin)"
 
+# minvmd finds its libkrun through a BINARY-RELATIVE rpath (`$ORIGIN/../lib` on
```
twitchyliquid64, line 163: We might want to update the corresponding spec for the installer (07 iirc) with this logic as well

norrietaylor: oh good call!

.github/workflows/nightly.yml
norrietaylor: good catch

norrietaylor: That matches what I can see from here, and it has bigger consequences than this line — flagging them rather than quietly dropping the artifact.

Corroborating evidence:

- `minvmd` hands libkrun its own kernel explicitly (`ctx.set_kernel(&self.kernel_path, …)`, `crates/minvmd/src/vm.rs:237`), which is the GPL-2 payload libkrunfw exists to provide.
- Nothing in `crates/minvmd/src/` references libkrunfw at all — the only mentions are in build.rs comments, the crate README and the packaging scripts.
- `crates/minvmd/README.md:100` already records that it is loaded "from the prefix (it is not a link-time DT_NEEDED)".
- **macOS already ships libkrun with no libkrunfw**: `stage-release.sh` has exactly one darwin libkrun row (`lib/libkrun.1.dylib`), and the macOS e2e + nightly smoke boot a real VM from that layout. So Linux shipping libkrunfw is the inconsistency, not the other way round.

If you're right, three things fall out beyond this assertion:

1. Drop the `libkrunfw` rows/uploads entirely — one fewer artifact per arch, and one fewer soname to keep in lockstep.
2. `scripts/rewrite-linux-linkage.sh` no longer needs to set `$ORIGIN` on libkrun's own RUNPATH; that entry exists *only* so its dlopen of libkrunfw finds the shipped sibling.
3. It removes the blocker I filed in gominimal/pkgs#533 for building `minvmd` as **static musl** — musl's static libc can't dlopen, so "libkrun must dlopen libkrunfw" was the thing standing in the way of dropping the new glibc floor entirely. If libkrunfw is unnecessary, static musl is back on the table.

The one thing I can't confirm from macOS is whether Linux libkrun *probes* for libkrunfw at init and hard-fails when it is absent, versus only touching it when no external kernel was set. Homebrew has libkrunfw installed here, so the local macOS boot doesn't isolate it.

Happy to drop it in this PR — the nightly Linux smoke now runs with no `LD_LIBRARY_PATH` and no materialize step, so a wrong guess fails loudly there rather than silently. Want me to, or would you rather confirm against a Linux host first?

twitchyliquid64: I’m fairly sure it doesn’t prove for libkrunfw

twitchyliquid64: *probe

norrietaylor: Sorry my agent got carried away here.

norrietaylor: Dropped in 403c7845 — thanks, this was worth more than the one line it was on.

Gone: the artifact, both uploads, the `lib/libkrunfw.so.5` staging rows, and the nightly smoke's placement of it. Also gone is the machinery that existed *only* for it:

- **libkrun's `$ORIGIN` RUNPATH.** Its sole justification was that libkrun `dlopen`s libkrunfw by soname and glibc resolves that against the calling object's RUNPATH. With nothing to find, the entry was pointless — the shipped `.so` is now left verbatim and only leak-checked.
- **The (file, soname) dispatch** I'd added an hour earlier to fix a path-vs-basename bug in the soname check. With one library there is no dispatch, so that bug is now unreachable by construction and its regression test went with it.

The larger consequence is in gominimal/pkgs#533, which I've updated: this **unblocks static-musl `minvmd`**. The chain was musl's static libc can't `dlopen` → libkrun `dlopen`s libkrunfw → so static musl could never get its kernel payload. With no `dlopen` on the path, `minvmd` can be fully static musl rather than the musl-dynamic fallback that issue was hedging on — which would retire the glibc floor entirely instead of tracking it, and take the `lib/` component and the `bin`/`lib` sibling constraint with it.

Also folded in @norrietaylor's point that we already build our own libkrun from source on macOS (`build-libkrun-macos.sh`, vendored + patched, pinned in `vendor/libkrun/libkrun.lock`), so the musl build reuses an existing pipeline rather than starting one.

PR #1060, opened 2026-07-29
feat: always set certain important variables automatically

evanspearman:
Summary

Sessions now come up with the environment variables a usable interactive shell expects — login identity, a styled prompt, and correct locale/timezone — set automatically instead of being left unset or wrong. This also fixes a correctness bug in how the daemon resolves inherited variables: they're now resolved against the user's environment, never the daemon's.

Motivation

A freshly-attached session shell was missing conventional environment a login shell would normally get from PAM//etc/passwd/sshd: no USER/LOGNAME/SHELL, a hardcoded PS1 forced by the launcher, and a locale (en_US.utf8 + LC_ALL) that isn't generated in the session rootfs — producing "cannot set locale" warnings. Worse, project-declared Inherit vars (e.g. LANG, TZ) were resolved daemon-side against the daemon's process environment, so a session would silently pick up the daemon's locale/timezone rather than the connecting user's.

What changed

Session defaults now live in the sandbox, as overridable defaults (crates/sandbox2/src/lib.rs)
- Session sandboxes get a styled default PS1, plus USER/LOGNAME (from the configured username) and SHELL=/usr/bin/bash — mirroring what sshd/PAM would set from /etc/passwd.
- Locale floor is now a safe, always-present C.UTF-8 set via LANG only (no LC_ALL), so a composition var or forwarded client locale can override it. Build/task sandboxes keep their fixed en_US.utf8 + LC_ALL for output stability.
- All set as plain defaults, so composed env_vars win on key collision. The launcher's hardcoded BASELINE_VARS (PS1) is removed accordingly.

Client forwards locale/timezone; daemon accepts an allowlist (crates/minimal/src/lib.rs, crates/minimald/src/session.rs)
- min forwards LANG, LC_*, and TZ into the session via ssh SendEnv.
- The daemon applies an AcceptEnv-style allowlist (inherited_session_env) that keeps only those and drops control-plane vars (MINIMAL_SESSION_ID, TRACEPARENT, etc.) so they never leak into the shell.
- ssh's ProxyCommand shell is pinned to /bin/sh, so a caller whose $SHELL is a bare name (fish) or an absent shell no longer breaks the transport.

Environment is layered by precedence at attach (crates/minimald/src/session_host.rs, session.rs)
- New AttachEnv + layer_session_env: client-forwarded locale/TZ sit below the composition (defaults it may override); per-connection facts (TERM, from the PTY request) sit above it (authoritative, sshd-style). SSH_TTY/SSH_CONNECTION are deliberately not set — nothing in-session backs them.

Inherited vars resolve against the user's env, not the daemon's (crates/sessions/src/core/{compose,primitives}.rs, daemon/composer.rs)
- New deferring_env() for the daemon composer: inherited vars are never resolved against the daemon's process environment.
- ResolvedVar now preserves its pre-resolution spec, and contribution_to_pending ships that original spec (Inherit/InheritWithDefault) to the client, which performs the single authoritative resolution against the user's env and recomputes carries_user_data.

Supporting fixes
- hardlink_dir_contents falls back to fs::copy on EXDEV (cache and destination on different filesystems, e.g. per-VM /state vs. rootfs /home), warning once per process instead of per file (crates/common/src/lib.rs).
- .minimal/minimal.toml adds just to session packages; the example project's locked_commit pin is refreshed and claude-code dropped.

Tests

- Allowlist keeps only LANG/LC_*/TZ and drops control-plane vars; LC_ is a prefix match, not substring.
- layer_session_env precedence: inherited < composition < connection, with non-colliding keys from every layer surviving.
- Daemon ships Inherit/InheritWithDefault specs and the client resolves them from the user's env (both hit and default-fallback paths), with carries_user_data set correctly.

> [!NOTE]
> ### Set PS1, USER, LOGNAME, SHELL, and LANG automatically in session shells and forward locale/timezone from client
> - Session containers now receive a baseline environment including `PS1`, `USER`, `LOGNAME`, `SHELL=/usr/bin/bash`, and `LANG=C.UTF-8` by default in [sandbox2/src/lib.rs](https://github.com/gominimal/minimal/pull/1060/files#diff-8079192560465d3b964c34d0ed8d6073e934202f55f1be88691ff5d8a810478b); `LC_ALL` is no longer set for sessions.
> - The SSH launcher in [minimal/src/lib.rs](https://github.com/gominimal/minimal/pull/1060/files#diff-1b62c63af5a5bd2dbc4b550bf4bc6d0fa87e649b54dc3a587da5a8fa947b3ce1) now runs `ProxyCommand` under `/bin/sh` and forwards `LANG`, `LC_*`, and `TZ` via `SendEnv`.
> - The server-side attach handler filters client-provided channel env vars to an allowlist (`LANG`, `LC_*`, `TZ`, `TERM`) and layers them into the session with composition vars and connection facts taking precedence.
> - Daemon-side composition in [sessions/src/daemon/composer.rs](https://github.com/gominimal/minimal/pull/1060/files#diff-7cf0a51983ffce8f78ac08cfec5718be1cafcbd265029a3dfbf1660fcf5d3564) now uses a no-op env resolver so inherited vars are resolved against the client's environment rather than the daemon's.
> - `hardlink_dir_contents` in [common/src/lib.rs](https://github.com/gominimal/minimal/pull/1060/files#diff-84e11d13acaf28044921bb5a427cfbf511c70a0b3881949137ba4093df58e266) now falls back to copying when source and destination are on different filesystems (`EXDEV`), emitting a one-time warning.
> - Behavioral Change: sessions that previously had no `SHELL`, `USER`, or `PS1` will now see these set; composition-defined vars still override the defaults.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7ad863b.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added `just` and `protobuf` to the default session packages.
  - Interactive shells now get consistent defaults (shell, prompt, and user/locale handling).
- **Bug Fixes**
  - SSH session attachment is more robust (uses a safe ProxyCommand shell and forwards locale/timezone settings).
  - Network diagnostics now capture interfaces and routes using `/proc/net` when system tools are unavailable.
  - Directory materialization continues across filesystem boundaries by copying when hardlinking isn’t possible.
  - Inherited environment variables are now resolved correctly on the client side across the RPC boundary.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -208,18 +208,40 @@ pub fn hardlink_dir_contents(src: &Path, dst: &Path) -> Result<(), HardlinkError
         } else if metadata.is_file() {
             match fs::hard_link(&path, &dst_path) {
                 Ok(()) => Ok(()),
-                Err(e) => {
-                    if e.kind() == std::io::ErrorKind::AlreadyExists {
+                Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+                    warn!(
+                        "Not linking {} => {}, already exists",
+                        path.display(),
+                        dst_path.display()
+                    );
+                    Ok(())
+                }
+                // The cache and destination can live on different filesystems
+                // (e.g. a per-VM `/state` volume vs. the rootfs holding
+                // `/home`), where hardlinks are impossible (`EXDEV`). Fall back
+                // to a copy so materialization still succeeds — slower and no
+                // longer deduplicated, but correct.
+                Err(e) if e.raw_os_error() == Some(libc::EXDEV) => {
+                    // Every file in a cross-device tree hits EXDEV, so warn only
+                    // on the first — a per-file log would flood with thousands
+                    // of identical lines. Once-per-process is enough: the cause
+                    // is a fixed filesystem-layout fact, not a per-file
+                    // condition.
+                    use std::sync::atomic::{AtomicBool, Ordering};
+                    static WARNED: AtomicBool = AtomicBool::new(false);
+                    if !WARNED.swap(true, Ordering::Relaxed) {
                         warn!(
-                            "Not linking {} => {}, already exists",
+                            "Copying instead of hardlinking: cache and \
+                             destination are on different filesystems; further \
+                             cross-device copies this run are silent (first: {} \
+                             => {})",
                             path.display(),
                             dst_path.display()
                         );
-                        Ok(())
-                    } else {
-                        Err(e)
                     }
+                    fs::copy(&path, &dst_path).map(|_| ())
                 }
+                Err(e) => Err(e),
             }
             .map_err(|e| HardlinkError::HardlinkFailed(path.to_path_buf(), dst_path, e))?;
         } else if metadata.is_symlink() {
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -221,7 +221,7 @@ pub fn hardlink_dir_contents(src: &Path, dst: &Path) -> Result<(), HardlinkError
                 // `/home`), where hardlinks are impossible (`EXDEV`). Fall back
                 // to a copy so materialization still succeeds — slower and no
                 // longer deduplicated, but correct.
-                Err(e) if e.raw_os_error() == Some(libc::EXDEV) => {
+                Err(e) if e.kind() == std::io::ErrorKind::CrossesDevices => {
                     // Every file in a cross-device tree hits EXDEV, so warn only
                     // on the first — a per-file log would flood with thousands
                     // of identical lines. Once-per-process is enough: the cause
--- a/.minimal/minimal.toml
+++ b/.minimal/minimal.toml
@@ -65,6 +65,9 @@ path = "usr/share/microvm-rootfs/rootfs.img"
 type = "oci-image"
 packages = ["libkrun"]
 
+[session]
+packages = ["just", "protobuf"]
+
 [tasks.vim]
 profile = "demo"
 state_key = ""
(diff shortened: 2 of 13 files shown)


crates/common/src/lib.rs
```diff
@@ -208,18 +208,40 @@ pub fn hardlink_dir_contents(src: &Path, dst: &Path) -> Result<(), HardlinkError
...
+                // to a copy so materialization still succeeds — slower and no
+                // longer deduplicated, but correct.
+                Err(e) if e.raw_os_error() == Some(libc::EXDEV) => {
+                    // Every file in a cross-device tree hits EXDEV, so warn only
+                    // on the first — a per-file log would flood with thousands
+                    // of identical lines. Once-per-process is enough: the cause
+                    // is a fixed filesystem-layout fact, not a per-file
+                    // condition.
+                    use std::sync::atomic::{AtomicBool, Ordering};
+                    static WARNED: AtomicBool = AtomicBool::new(false);
```
twitchyliquid64, line 231: Statics make me sad but I guess theres no way to do this cleaner without it being more verbose than its worth

crates/common/src/lib.rs
```diff
@@ -208,18 +208,40 @@ pub fn hardlink_dir_contents(src: &Path, dst: &Path) -> Result<(), HardlinkError
...
+                             cross-device copies this run are silent (first: {} \
+                             => {})",
                             path.display(),
                             dst_path.display()
                         );
-                        Ok(())
-                    } else {
-                        Err(e)
                     }
+                    fs::copy(&path, &dst_path).map(|_| ())
```
twitchyliquid64, line 242: Im also not sure how I feel about this, maybe we are far enough along that its better to be slow and fallback to a copy, but also my other thought is this might mask when your paths are wrong

twitchyliquid64 (approved):

crates/common/src/lib.rs
evanspearman: Yeah, I don't really like it either. I don't think we want the log spammed with every single file being copied though. There's probably a better way to do this, but we don't have a ton of time today.

crates/common/src/lib.rs
evanspearman: That's a good point. This is really only required for nested sandboxes, and there may be a better way. We should probably change it to be a preference, but I'll cut an issue for it given timing.

Issue #1065, opened 2026-07-29
Ship minvmd to Linux as a single static-musl binary (link libkrun statically, drop libkrunfw)


norrietaylor:
`minvmd` is the only one of the four binaries that cannot ship to Linux users today. The three musl binaries (`min`, `mip`, `minimald`) are self-contained; `minvmd` is a native-glibc build that links `libkrun.so` and dlopens `libkrunfw.so.5`. `release.yml` states the constraint directly:

> minvmd links libkrun's KVM backend dynamically, so unlike the three binaries above it can't join the static-musl cross build — it needs a native glibc build against a real libkrun.

Shipping it as-is means ~29 MB of libraries on users' disks plus RUNPATH machinery, a `bin`/`lib` sibling constraint, and a glibc floor. This issue tracks removing the constraint instead: build `minvmd` as a single static-musl binary with libkrun linked in.

This re-creates a tracking issue that was deleted (`gominimal/pkgs#533`, now HTTP 410). The proof-of-concept branch (`scratch/static-musl-minvmd-533`) was never pushed to either remote, so the findings below are the surviving record.

## Status: proven feasible, not productionized

A Linux agent took a scratch branch through the full build. **All six phases passed on aarch64** — the static-musl `minvmd` builds, links libkrun, and runs. Binary size ~29 MB → **~8 MB**.

**Not yet verified on x86_64.** The proof is aarch64-only. This must be settled in CI, not on a workstation.

## What it takes

**1. `libkrunfw` is dead weight — drop it.** This is the finding that unblocks everything else: musl static binaries cannot `dlopen`, so a `libkrunfw` dependency would have been fatal. It turns out `minvmd` never needed it — it supplies its own kernel via `ctx.set_kernel`, and libkrunfw exists only to carry a bundled GPL-2 guest kernel. Disproved the strong way (removed it; VM still boots). Dropping it also removes a GPL-2 blob from the shipped closure.

**2. Patch `crate-type`.** At the pinned commit, `src/libkrun/Cargo.toml` declares:
```toml
[lib]
name = "krun"
crate-type = ["cdylib", "lib"]
```
No `staticlib`, so there is no `libkrun.a` to link. Needs `["cdylib", "staticlib", "lib"]`. (Note musl targets are `crt-static`, so `cdylib` is not produced there at all.)

[...]

PR #1070, opened 2026-07-29
feat(minvmd): build against a static musl libkrun

norrietaylor:
First step of [#1065](https://github.com/gominimal/minimal/issues/1065): make `minvmd` buildable as a self-contained static-musl binary. **This does not wire it into the release yet** — that is the next PR, and it is where x86_64 gets proven.

## Why

`minvmd` is the only one of the four binaries that cannot ship to Linux users. `min`, `mip` and `minimald` are self-contained musl builds; `minvmd` is a native-glibc build that links `libkrun.so` and dlopens `libkrunfw.so.5`. `release.yml` states the constraint outright:

> minvmd links libkrun's KVM backend dynamically, so unlike the three binaries above it can't join the static-musl cross build — it needs a native glibc build against a real libkrun.

Shipping that as-is puts ~29 MB of libraries plus RUNPATH machinery, a `bin`/`lib` sibling constraint, and a glibc floor onto users' disks. This removes the reason to.

## `scripts/build-libkrun-linux.sh`

The Linux twin of `build-libkrun-macos.sh` — same vendored pin, same patch series, same trim, different link model. Three things it must do that the dylib build does not:

**Add `staticlib` to the crate-type.** Upstream declares `crate-type = ["cdylib", "lib"]`, so cargo emits no archive to link. Applied as a guarded `sed` rather than a carried `.patch`: this is not an upstream fix but a property of how *we* consume the crate, and a context-free edit survives a pin bump that would reject a line-anchored patch. The guard fails loudly if upstream's declaration ever changes shape.

**Merge, *then* localize — the order is load-bearing.** `libkrun.a` carries its whole Rust dependency closure, including a copy of libstd that collides with `minvmd`'s own. `ld -r --whole-archive` merges into a single object first; only then does `objcopy` cut it down to the `krun_*` globals. Doing it the other way round fails subtly: `objcopy` over an archive rewrites each member independently, so a symbol defined in one member and referenced from another gets localized at the definition and becomes unresolvable from the reference.

**Drop `libkrunfw` entirely.** It exists only to carry a bundled guest kernel that `minvmd` never uses — it supplies its own via `ctx.set_kernel`. This is what makes the whole thing possible: a static musl binary *cannot* `dlopen`, so a real `libkrunfw` dependency would have been fatal rather than merely wasteful. It also takes a GPL-2 kernel blob out of the shipped closure.

## Build-side changes

`build.rs` treats a `libkrun.a` in the resolved prefix as the selector for a static link: it emits `minvmd_libkrun_static` and records **no rpath**, since a static link has nothing to resolve at load time. `raw.rs`'s `#[link]` becomes conditional on that cfg.

Linux only, deliberately. macOS builds and ships a dylib, and quietly switching that proven path because some unrelated libkrun install happened to drop an archive in `/opt/homebrew/lib` would be a surprise, not a feature.

## Drive-by: a latent bug in the macOS script

It ran cargo with `--manifest-path` from the *minimal* repo root. Cargo resolves `.cargo/config.toml` from the **CWD**, not from the manifest's directory — so it read ours and ignored libkrun's. At the pin, that file contains:

```toml
[target.'cfg(target_env = "musl")']
rustflags = ["--cfg", "musl_v1_2_3"]
```

On darwin this only drops a test runner, which is why nobody noticed. On musl it silently drops `--cfg musl_v1_2_3`, which is load-bearing. Fixed in both scripts so the two cannot disagree about what they compiled.

## Verification

Built end-to-end on aarch64 (`rust:alpine`, GNU binutils 2.45.1):

| check | result |
|---|---|
| `libkrun.a` builds | ✅ — cargo confirms the crate-type edit landed by warning it dropped the now-inert `cdylib` for a `crt-static` target |
| symbol surface after localization | **67** `krun_*` globals, **0** non-`krun_` globals |
| `minvmd` links it | ✅ — 8,335,392-byte binary |
| `file` | `ELF 64-bit LSB executable, ARM aarch64, statically linked` |
| `ldd` | `not a valid dynamic program` — no dynamic deps at all |
| real backend, not the stub | `kvm_run`, `kvm_coalesced_mmio`, `KRUN_BLOCK_ROOT_FSTYPE` all present; no stub markers; binary runs |

~29 MB of shipped libraries becomes an 8 MB self-contained binary.

`shellcheck` clean at 0.11.0 and CI's 0.10.0. `cargo clippy -p minvmd --all-targets -- -D warnings` clean, and the static branch typechecks under a forced `--cfg minvmd_libkrun_static`.

**Not proven here:** x86_64, and actually booting a VM (needs KVM). Both belong in CI, in the follow-up that adds the release wiring.

Refs [#1065](https://github.com/gominimal/minimal/issues/1065), [#980](https://github.com/gominimal/minimal/issues/980)

> [!NOTE]
> ### Build minvmd against a static musl libkrun on Linux
> - Adds [scripts/build-libkrun-linux.sh](https://github.com/gominimal/minimal/pull/1070/files#diff-cb8ac7b8a267453b00b6331f43b3d99a36f8a9f8c5cc1832354715cedb84cc3b) to build a static `libkrun.a` from source at a pinned commit, applying patches, merging archive members into a single relocatable object to localize non-`krun_*` symbols, and staging the result into a given prefix.
> - Updates [build.rs](https://github.com/gominimal/minimal/pull/1070/files#diff-adfb9d2ca5918d8192f83cf6ffe3922ac7f7041eab25d6bb29da016e6f7016ae) to detect `libkrun.a` in the prefix and emit `cfg(minvmd_libkrun_static)`; when set, rpath emission is skipped.
> - Updates the FFI extern block in [raw.rs](https://github.com/gominimal/minimal/pull/1070/files#diff-8f2259c33ed7f31bd14cb658b4ad27065d4b433dd99ff03703c24b663b8d30f9) to link `krun` with `kind = "static"` when `minvmd_libkrun_static` is set, falling back to dynamic linking otherwise.
> - Fixes the macOS build script to `cd` into the libkrun source directory so its `.cargo/config.toml` is respected, matching the Linux behavior.
> - Risk: static linking localizes non-`krun_*` symbols, which may cause conflicts if other crates in the binary also statically link libraries with overlapping symbol names.
>
> 
> #### Changes since #1070 opened
>
> - Added environment variable-based build requirement enforcement and static library detection to the `minvmd` build script [697b3a9]
> - Modified `libkrun` build scripts to detect and export the active rustup toolchain from the repository root before changing directories [697b3a9]
> - Replaced bash-specific indirect expansion for reading target-specific linker environment variable with eval-based assignment to a new variable, updated conditional to test the new variable, and added a comment explaining the avoidance of bash-specific substitution for POSIX sh compatibility [9338e89]
> - Added composite action for building and staging static musl `libkrun.a` [bc6b20e]
> - Changed release workflow to build static musl `minvmd` for `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` targets [bc6b20e]
> - Updated CI workflow to build and test static musl `minvmd` for `x86_64-unknown-linux-musl` [bc6b20e]
> - Updated nightly workflow to verify static linkage of shipped `minvmd` [bc6b20e]
> - Updated staging script and documentation to reflect static musl builds for Linux [bc6b20e]
> - Updated CI workflow validation checks to accept both 'statically linked' and 'static-pie linked' outputs from `file(1)` when verifying `minvmd` binary linkage [a26b89e]
> - Restricted static linking to musl targets only [4b51864]
> - Added cross-architecture build support with automatic binutils selection [4b51864]
> - Replaced `eval`-based environment variable reading with `printenv` [4b51864]
> - Added validation in `scripts/build-libkrun-linux.sh` to ensure the produced `libkrun.a` contains no undefined Rust symbols by running `nm -u` and filtering for Rust symbol patterns including C++ mangled symbols starting with `_ZN`, Rust v0 mangled symbols starting with `_R`, `__rust*` symbols, and `rust_eh_personality`, failing the build if any are detected [7a055e5]
> - Modified symbol extraction in `scripts/build-libkrun-linux.sh` to wrap the `grep '^krun_'` command in `{ grep '^krun_' || true; }` when generating `keep.syms` from the merged object, allowing the script to continue under `pipefail` mode when no `krun_*` symbols match and letting the subsequent explicit count check emit a diagnostic [7a055e5]
> - Replaced live piped symbol verification commands `strings -a "$bin" | grep -q 'kvm_run'` in both the amd64 and arm64 Linux release build workflows with a two-step process that first redirects `strings -a "$bin"` output to a temporary file (`$RUNNER_TEMP/minvmd-{amd64,arm64}.strings`) and then greps that file for `kvm_run` [7a055e5]
> - Clarified in `docs/internal/release-pipeline.md` that amd64 musl builds may be reported by `file(1)` as either 'statically linked' or 'static-pie linked' [7a055e5]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 69f9dc6.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Linux releases now include statically linked `minvmd` binaries with embedded libkrun support for amd64 and arm64.
  * Added Linux microVM payloads, including initramfs, rootfs, and kernel artifacts.
* **Bug Fixes**
  * Improved reliability of Linux and macOS libkrun builds.
  * Builds now validate required symbols and linkage before producing release artifacts.
* **Documentation**
  * Updated release pipeline documentation to reflect the new Linux components and static packaging.

### DIFF
--- a/.github/actions/build-libkrun-static-linux/action.yml
+++ b/.github/actions/build-libkrun-static-linux/action.yml
@@ -1,0 +1,99 @@
+name: Build static libkrun (Linux/musl)
+description: >
+  Provision the STATIC libkrun.a that minvmd links into a self-contained musl
+  binary — built from source at the commit pinned in vendor/libkrun/libkrun.lock
+  via scripts/build-libkrun-linux.sh — into a commit-keyed prefix, and publish
+  LIBKRUN_PREFIX to GITHUB_ENV. minvmd's build.rs treats a libkrun.a in the
+  prefix as the selector for a static link and records no rpath, so the
+  resulting binary has no runtime library dependency at all. The Linux twin of
+  setup-libkrun-macos, and the alternative to setup-libkrun-linux (which
+  materializes the DYNAMIC upstream package for dev/native builds).
+  Build-on-miss, riding actions/cache keyed by every build input. Requires a
+  Rust toolchain with the musl target, musl-tools, and the repository checked
+  out.
+
+inputs:
+  target:
+    description: musl target triple (e.g. x86_64-unknown-linux-musl)
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Resolve libkrun pin
+      id: pin
+      shell: bash
+      env:
+        TARGET: ${{ inputs.target }}
+      run: |
+        set -euo pipefail
+        case "$TARGET" in
+          *-linux-musl) ;;
+          *) echo "::error::target '$TARGET' is not a musl triple" >&2; exit 1 ;;
+        esac
+        commit="$(sed -n 's/^commit=//p' vendor/libkrun/libkrun.lock)"
+        if [ -z "$commit" ]; then
+          echo "::error::no commit= line in vendor/libkrun/libkrun.lock" >&2
+          exit 1
+        fi
+        script_hash="$(sha256sum scripts/build-libkrun-linux.sh | cut -d' ' -f1)"
+        # Hash the carried patches (name + content, in apply order) so editing,
+        # adding, or removing a patch invalidates the build. Mirrors the macOS
+        # composite, including the `-d` guard: git drops the directory once the
+        # last patch is removed, and a non-zero `find` would become a step
+        # failure under pipefail.
+        patches_hash="$( { [ -d vendor/libkrun/patches ] && find vendor/libkrun/patches -name '*.patch' -type f -exec sha256sum {} + || true; } | sort | sha256sum | cut -d' ' -f1)"
+        {
+          echo "commit=$commit"
+          # The prefix embeds every build input, and the TARGET besides: one
+          # runner arch builds one triple, but keying on it keeps an amd64 and
+          # an arm64 prefix from ever colliding in a shared cache.
+          echo "prefix=$HOME/.cache/minimal-ci/libkrun-static/$TARGET/$commit-${script_hash:0:12}-${patches_hash:0:12}"
+          echo "script_hash=$script_hash"
+          echo "patches_hash=$patches_hash"
+        } >> "$GITHUB_OUTPUT"
+    - name: Cache built libkrun.a
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+      with:
+        path: ~/.cache/minimal-ci/libkrun-static
+        key: libkrun-static-${{ inputs.target }}-${{ steps.pin.outputs.commit }}-${{ steps.pin.outputs.script_hash }}-${{ steps.pin.outputs.patches_hash }}
+    - name: Build static libkrun from source (on miss)
+      shell: bash
+      env:
+        PREFIX: ${{ steps.pin.outputs.prefix }}
+        TARGET: ${{ inputs.target }}
+      run: |
+        set -euo pipefail
+        if [ ! -f "$PREFIX/libkrun.a" ]; then
+          ./scripts/build-libkrun-linux.sh "$PREFIX" "$TARGET"
+        else
+          echo "libkrun.a already staged at $PREFIX (cache hit)"
+        fi
+    - name: Verify staged libkrun.a + export LIBKRUN_PREFIX
+      # Re-assert the API floor and the symbol surface on the staged (possibly
+      # cache-restored) archive, so a corrupt or hand-tampered prefix fails here
+      # with an actionable message instead of a link error — or worse, a
+      # duplicate-symbol collision — much later.
+      shell: bash
+      env:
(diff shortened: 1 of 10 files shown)


twitchyliquid64 (approved):
The object/symbol surgery to make this work feels very fragile, in particular coalescing the rust internal symbols between libkrun and the minvmd binary. Down to give it a go, i guess time will tell how fragile this is.

norrietaylor:
> The object/symbol surgery to make this work feels very fragile, in particular coalescing the rust internal symbols between libkrun and the minvmd binary. Down to give it a go, i guess time will tell how fragile this is.

I agree, fragile indeed. It is cheap to try but if it means that dev, updates become brittle I will walk it back.

PR #1123, opened 2026-07-31
fix(minimald): lead the shell-exit prompt with the files changed since activation

norrietaylor:
Exiting a session asks keep-or-delete without saying what would be lost; the prompt now leads with the files changed since activation and the delete option says when nothing would be lost.

- Baseline snapshot (relative path -> len+mtime) taken before the session process launches; re-walk and diff at shell exit.
- Prompt shows `N files changed since activation:` with up to 10 `A/M/D <path>` rows, or `No files changed since activation.` (delete option gains `— nothing will be lost`).
- Best-effort throughout: an unavailable delta renders the previous prompt verbatim; the exit path never blocks on change detection.
- Covered by unit tests on the snapshot/diff module and an end-to-end prompt test in the session harness.

> [!NOTE]
> ### Lead the minimald shell-exit prompt with files changed since activation
> - Adds a new `session_delta` module that snapshots a session workspace at activation time using `walkdir` and `blake3`, then diffs it on shell exit to produce sorted A/M/D rows.
> - The shell-exit prompt in `session_host` now shows up to 10 changed-file rows with A/M/D markers; if no files changed, the Delete option appends '— nothing will be lost'.
> - Snapshot walk is bounded by a 5s timeout and skips symlink traversal; files over 4 MiB are signed by length and mtime only, without a digest.
> - Risk: if the baseline snapshot times out or fails, `delta` is left as `None` and the prompt falls back to the plain (no-diff) display.
>
> 
> #### Changes since #1123 opened
>
> - Modified `snapshot` function in `session_delta` module to exclude the root-level .git directory from baseline and re-walk operations by adding a `filter_entry` that rejects entries where depth equals 1 and file_name equals ".git", while preserving nested .git directories as ordinary content [90ba487]
> - Refactored shell-exit prompt logic in `Binding.run` method within `session_host` module by extracting inline prompt code into a new generic async associated function `Binding.shell_exit_prompt` that computes changed files via `DeltaSource`, renders either a no-changes banner or a capped list of changed files, and presents a two-item selection menu for Exit or Delete actions [90ba487]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3767383.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Shell-exit prompts now show workspace files added, modified, or deleted during a session.
  * Changes are detected reliably even when file size and timestamps remain unchanged.
  * Changed files are listed consistently, with long lists limited for readability.
  * Cleanup options update based on whether changes are present.

* **Bug Fixes**
  * Sessions continue to exit safely when workspace change information cannot be collected.
  * Unchanged workspaces no longer display unnecessary file-change details.
  * Symbolic links are handled safely without scanning their targets.

### DIFF
--- a/crates/minimald/src/session_delta.rs
+++ b/crates/minimald/src/session_delta.rs
@@ -12,6 +12,13 @@
 //! their signatures carry a content digest, so a same-length edit that
 //! preserves the modification time is still reported. Above the cap the
 //! comparison falls back to (length, mtime).
+//!
+//! The workspace root's `.git` directory is excluded from both walks: the
+//! delta reports working-tree files, and git-internal churn (index, refs,
+//! objects) no longer appears in it — without the exclusion, any in-session
+//! git operation floods the capped row list with `.git` noise. Only the
+//! root-level `.git` is skipped; a nested one (a vendored subrepo) is
+//! ordinary workspace content.
 
 use std::collections::BTreeMap;
 use std::path::{Path, PathBuf};
@@ -100,11 +107,16 @@ impl DeltaSource {
 /// every non-directory entry keyed by its root-relative path. A symlink is
 /// recorded via its own metadata — never traversed, even when it points at a
 /// directory — so links out of the root or link cycles cannot widen or wedge
-/// the walk. Unreadable entries fail the whole snapshot: a partial baseline
-/// would misreport their contents as added later.
+/// the walk. The root-level `.git` directory is skipped entirely (see the
+/// module docs for the tradeoff). Unreadable entries fail the whole
+/// snapshot: a partial baseline would misreport their contents as added
+/// later.
 fn snapshot(root: &Path) -> std::io::Result<Snapshot> {
     let mut out = Snapshot::new();
-    for entry in WalkDir::new(root) {
+    for entry in WalkDir::new(root)
+        .into_iter()
+        .filter_entry(|e| !(e.depth() == 1 && e.file_name() == ".git"))
+    {
         let entry = entry?;
         if entry.file_type().is_dir() {
             continue;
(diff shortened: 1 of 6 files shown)


evanspearman:
Unless there's a reason not to, I would strongly recommend using either the `walkdir` crate (or `ignore` if you want a version that respects .gitignore and similar ignore lists) and use `WalkParallel` for performance.

crates/minimald/src/session_host.rs
```diff
@@ -486,11 +504,49 @@ impl Binding {
             // We presume they didnt want to completely destroy the session, perhaps just detach, but lets prompt
             // to see where they wanted to go from here.
             let _ = w.write_all(b"\r\n").await;
+
+            // Lead with what a "delete" would lose. An unavailable delta (no
+            // baseline, or the re-walk failed) renders the plain prompt — the
+            // exit path never blocks on change detection.
+            let changed = match &self.delta {
```
twitchyliquid64: Nit: move the bulk of this to a helper function, the handler here is getting long

crates/minimald/src/session_delta.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Walks `root` without following symlinks (walkdir's default), recording
+/// every non-directory entry keyed by its root-relative path. A symlink is
+/// recorded via its own metadata — never traversed, even when it points at a
+/// directory — so links out of the root or link cycles cannot widen or wedge
+/// the walk. Unreadable entries fail the whole snapshot: a partial baseline
+/// would misreport their contents as added later.
+fn snapshot(root: &Path) -> std::io::Result<Snapshot> {
+    let mut out = Snapshot::new();
+    for entry in WalkDir::new(root) {
```
twitchyliquid64: A nice optimization here might be to skip a `.git` in the root

norrietaylor:
Review follow-ups landed in 3767383c (and merged into feat/exit-prompt-archive as 4cb51be5):

- **walkdir adopted** for both walks — @evanspearman, `ignore` was considered and rejected: respecting gitignore would hide changed-but-ignored files from what is fundamentally a data-loss report, and "you're about to lose these files" must include the ignored ones.
- **Exit re-walk is bounded**: the spawn_blocking await is wrapped in a 5 s `tokio::time::timeout`; on expiry the prompt degrades to its plain form and teardown never blocks (the abandoned walk finishes detached on the blocking pool).
- **Symlinks are structurally unfollowed**: walkdir's no-follow default means a link out of the root or a link cycle is recorded as an entry, never traversed — covered by a regression test with both an escape link and a cycle.
- **Size-capped blake3 content digest** (4 MiB cap) in each file signature, so a same-length mtime-preserved edit can no longer masquerade as "no files changed" — covered by a regression test that rewrites same-length bytes and restores the mtime via `File::set_times`, then requires an `M` row.

evanspearman (approved):

norrietaylor:
Thanks @twitchyliquid64 — both addressed in 90ba487b (merged forward to feat/exit-prompt-archive as d6400eef):

- **Prompt handler extracted**: `Binding::run`'s ProcessExited epilogue now lives in a `shell_exit_prompt` associated fn (taking the binding's delta/control capabilities piecewise, since `run` has already split the channel out of `self` at that point); channel output is byte-identical and the existing prompt/teardown tests pass unchanged. On the archive branch the helper carries the extra save-then-delete arm.
- **Root `.git` skipped in the delta walk**: both the baseline and exit walks (one shared walk fn) now filter a depth-1 `.git` entry, so any in-session git operation no longer floods the capped row list with index/refs/objects churn — the documented tradeoff is that the delta (and the save archive) reports working-tree files only, never git-internal state. Root-level only: a nested `.git` (vendored subrepo) remains ordinary workspace content. Covered by a unit test asserting `.git` churn is invisible while a working-tree change and a nested `.git` still report.

PR #1139, opened 2026-07-31
feat(minimal): add min task run — declared tasks in ephemeral sessions

norrietaylor:
Running a project task means hand-building a session, attaching, and remembering to destroy it; `min task run <task>` now does the whole loop in an ephemeral box — create (named `task-<task>-<hex>`), exec the task with output streamed through, exit with the task's exit code, destroy — and `--keep` retains the box as an attachable session.

- Exec rides the native channel with `MINIMAL_SESSION_ID` set, sending the canonical in-box `min task run` spelling.
- Errors name the fix: no blueprint → `min init`; unknown task → the declared list. A hidden top-level `run` catches the bare spelling.
- Unit tests cover the CLI surface and name mint; e2e asserts the loop end-to-end.

> [!NOTE]
> ### Add `min task run` command to execute declared tasks in ephemeral sessions
> - Adds `min task run <task>` CLI command that creates a named ephemeral session, uploads the workspace, runs the declared task via SSH exec with `MINIMAL_SESSION_ID` set, streams stdout/stderr to the user, and destroys the session on completion unless `--keep` is passed.
> - Validates the task name against `minimal.toml` client-side before creating a session, with clear error messages for unknown or malformed task names.
> - Adds a hidden `min run ...` catch-all subcommand that always errors with a redirect message pointing to `min task run`.
> - Ctrl-C during a task run triggers best-effort session teardown within a 10s timeout (or retains the session with `--keep`) and exits with code 130.
> - Behavioral Change: `min task run` routes tracing logs to stderr and propagates the task's exit code directly; non-zero exits suppress error printing in main.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9b294d5.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added `min task run <task>` for executing project tasks in temporary sessions.
  - Supports optional project paths and `--keep` to retain sessions after execution.
  - Streams task output and returns the task’s exit code directly.
  - Added session-aware execution support for task runs.

- **Bug Fixes**
  - Improved interruption cleanup and session retention behavior.
  - Added compatibility handling for legacy `min run` arguments.

- **Tests**
  - Added coverage for command parsing, output, exit codes, cleanup, retention, and task errors.

### DIFF
--- a/crates/minimal/src/client.rs
+++ b/crates/minimal/src/client.rs
@@ -302,12 +302,38 @@ impl Client {
         &mut self,
         command: &str,
     ) -> Result<russh::Channel<russh::client::Msg>, anyhow::Error> {
+        self.exec_channel(None, command).await
+    }
+
+    /// Like [`Self::open_exec_channel`], but sets `MINIMAL_SESSION_ID` on the
+    /// channel before the exec request, the routing contract the daemon's
+    /// `min`-prefixed exec forms (`min task run <task>`, `min package build`,
+    /// `min check`) are served under.
+    pub async fn open_session_exec_channel(
+        &mut self,
+        session_id: sessions::SessionId,
+        command: &str,
+    ) -> Result<russh::Channel<russh::client::Msg>, anyhow::Error> {
+        self.exec_channel(Some(session_id), command).await
+    }
+
+    async fn exec_channel(
+        &mut self,
+        session_id: Option<sessions::SessionId>,
+        command: &str,
+    ) -> Result<russh::Channel<russh::client::Msg>, anyhow::Error> {
         let mut channel = self
             .handle
             .channel_open_session()
             .await
             .context("open exec channel")?;
         send_traceparent(&channel).await;
+        if let Some(id) = session_id {
+            channel
+                .set_env(true, "MINIMAL_SESSION_ID", id.to_string())
+                .await
+                .context("set MINIMAL_SESSION_ID env")?;
+        }
         channel
             .exec(true, command)
             .await
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -21,6 +21,7 @@ mod file_upload;
 pub mod git_remote;
 pub mod loadouts;
 pub mod prompt;
+pub mod task;
 pub mod theme;
 
 #[derive(Parser)]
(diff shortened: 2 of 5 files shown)


twitchyliquid64 (approved):
Something for the future might be a new field in the session record to mark ephemeral sessions intended for task exec, along with a TTL etc

norrietaylor:
> along with a TTL etc

Absolutely! That would be fantastic

norrietaylor:
Filed a follow-up tracking the ephemeral/task marking on the session record + TTL (with the reaper deliberately coordinated with the broader session-lifecycle design rather than ahead of it). Thanks @twitchyliquid64.

norrietaylor:
Addressed the exit-code review feedback in 8aabd891:

- **Missing exit status now fails closed.** The exec bridge tracked the remote status as a bare `u32` initialized to 0, so a daemon or transport death that closed the channel before an `ExitStatus` message read as success. The bridge now tracks receipt explicitly (`Option<u32>`); a run that ends without a reported status prints `task ended without reporting an exit status` and exits 1. The status resolution is factored into a pure `exit_outcome(Option<u32>)` with unit coverage for the missing-status path.
- **Ctrl-C behavior documented.** The interrupt guard's doc comment and the `task run` help text now state that SIGINT is consumed on the host — the session is torn down (or kept with `--keep`) and the process exits 130 without relaying the interrupt into the box, so the task gets no chance to handle it, unlike a local exec.
- **Non-interactive stdin documented.** The bridge's doc comment and help text now spell out that tasks run non-interactively: a terminal stdin is half-closed immediately (the daemon exec channel has no PTY — only the shell path does), so input reaches the task only when piped; interactive work belongs in `min session attach`.

The other gap in the analysis — the daemon mapping signal deaths to `unwrap_or(1)` instead of `128+signo` in `crates/minimald/src/exec.rs` — predates this PR and is tracked separately; not touched here.

norrietaylor:
Addressed the three CodeRabbit findings in 8a76e7ac:

- **Interrupt cleanup is now bounded and re-interruptible** (`task.rs` guard). The Ctrl-C handler's connect + `DestroySession` runs under the same 10-second cleanup ceiling as `best_effort_destroy` (with a manual-cleanup hint if it times out), and is raced against a second `ctrl_c()` listener: a second interrupt during cleanup abandons it and exits 130 immediately. An unresponsive daemon can delay the exit, never prevent it, and a second Ctrl-C is no longer silently swallowed. The guard's doc comment now describes this behavior.
- **Untrimmable task names rejected client-side** (`check_task_declared`). The daemon trims the task name out of the exec command string, so a declared name with leading/trailing whitespace could never round-trip into the box. Such names are now rejected before any config lookup, with the name debug-quoted so the whitespace is visible (e.g. `task name " build " has leading or trailing whitespace...`). Unit-tested, including that the rejection fires ahead of the minimal.toml discovery.
- **e2e leftover-session checks fixed for pipefail** (`scripts/session-e2e.sh`). The two `mnl ls | grep -q` checks could falsely pass: `grep -q`'s early exit SIGPIPEs the ls under `pipefail`, turning a genuine leftover into a non-zero pipeline. Both now capture `mnl ls` output to a variable and glob-match it — the same convention the script's attach-proof markers prescribe. Shellcheck stays clean.

PR #1145, opened 2026-07-31
feat: plumb daemon IDs through to identify a process

### LINKED ISSUE #1001: sandbox2: the sandbox leader should own its namespace, not the creating process
## Goal

Make a sandbox's **leader process** the owner of its namespace, and record that
leader's pid where the sandbox directory can be reclaimed against it. Today the
directory records the pid of whatever process *created* it, which is not the
process whose lifetime the sandbox actually tracks.

## Why it matters

`crates/sandbox2/src/config.rs` names each sandbox directory
`{name}-{timestamp}-{attempt}-{pid}` from `std::process::id()` — the creator.

- Under the **CLI** that happens to work: `mip` creates the sandbox and lives
  roughly as long as it, so when `mip` exits the pid dies and the directory is
  reclaimable. `mip cache clean` relies on exactly this
  (`crates/mip/src/cmd_cache.rs`, `cleanup_stale`).
- Under the **daemon** it does not. `minimald` creates the sandbox and outlives
  it, so the recorded pid stays alive for the daemon's whole life. Inside the
  microVM `minimald` is **pid 1**, so every sandbox directory is stamped `-1`
  and `/proc/1` exists by definition — including across daemon restarts, where
  the new daemon is pid 1 too.

Observed directly on a booted VM: every directory under the build base carried
`-1`, e.g. `build-1784239277-0-1`.

twitchyliquid64:
Fixes: #1001

Temporary files are now minted using a `daemon_id` suffix, which if not provided defaults to `PID` like before.

Needed to fix cache/implement clean under minimald.

> [!NOTE]
> ### Plumb daemon IDs through sandbox, build, and check operations to identify processes
> - Generates a unique alphanumeric `daemon_id` in `minimald::ServerState` and propagates it through `mctx::Config` to all downstream components.
> - Adds `daemon_id: Option<String>` to `op::Options`, `CheckCtx`, `sandbox2::config::Config`, and the orchestrator, threading the ID through builds, checks, materializations, and standalone tests.
> - Changes sandbox temp directory naming in `sandbox2::config::Config.build` to use `daemon_id` as a suffix when set, falling back to the current PID.
> - CLI commands (`cmd_check`, `cmd_materialize`, `cmd_pkg_patched_build`) forward `ctx.daemon_id()` into their respective operations.
> - Behavioral Change: sandbox temp directories are now named by daemon ID rather than PID when running under `minimald`, which affects any tooling that parses sandbox directory names.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a1f61f.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added optional daemon identity support across build, check, materialization, and test operations.
  * Server sessions now receive an automatically generated daemon identifier.
  * Sandbox directories use the daemon identifier when available, improving isolation and traceability.
  * Exposed daemon identity through runtime context and server state APIs.

### DIFF
--- a/crates/check/src/lib.rs
+++ b/crates/check/src/lib.rs
@@ -344,6 +344,7 @@ pub struct CheckCtx {
     pub stdlib_dir: PathBuf,
     pub cache: Cache<LocalDir>,
     pub ot: Option<OpTracker>,
+    pub daemon_id: Option<String>,
 
     /// Fires when the caller wants the run to stop.
     ///
@@ -361,6 +362,7 @@ pub struct CheckCtx {
 }
 
 impl CheckCtx {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         filter_names: Vec<String>,
         skip_checkers: Vec<String>,
@@ -369,6 +371,7 @@ impl CheckCtx {
         stdlib_dir: PathBuf,
         cache: Cache<LocalDir>,
         ot: Option<OpTracker>,
+        daemon_id: Option<String>,
     ) -> Self {
         Self {
             filter_names,
@@ -378,6 +381,7 @@ impl CheckCtx {
             stdlib_dir,
             cache,
             ot,
+            daemon_id,
             cancel: CancellationToken::new(),
             semaphore: Arc::new(Semaphore::new(20)),
             check_cache: Arc::new(CheckCache::new()),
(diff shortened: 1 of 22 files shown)


norrietaylor (approved):

norrietaylor:
Lgtm

PR #1148, opened 2026-07-31
feat(minimal,minimald): confirm single-session destroy with the workspace delta

norrietaylor:
`min session destroy` deletes unsaved in-session work without a word; it now lists the files changed since activation and confirms with default No — and headless runs refuse without `--force`, so EOF can never read as consent.

- A small oneshot RPC serves the running host's delta baseline (merged with the exit-prompt work); rows render exactly as the exit prompt's.
- Headless refusal mirrors the existing `--all` precedent; `--force` is unhooked from `--all` and skips the confirm.
- Delta unavailable (stopped session, error, timeout) degrades to the plain confirm; destroy never blocks on the listing. `--all` unchanged.

> [!NOTE]
> ### Add confirmation prompt with workspace delta to single-session destroy
> - Single-session `min session destroy` now prompts for confirmation before destroying, showing changed files since activation when available (up to 10 rows).
> - A new `destroy_gate` function decides whether to prompt or proceed: `--force` skips the prompt; headless runs (no TTY or `--no-input`) without `--force` return an error instructing the user to pass `--force`.
> - A new `SessionDelta` RPC fetches changed files from the daemon with a 6-second timeout; timeouts and errors silently map to `None`, so the prompt still appears without the file listing.
> - The `--force` flag is no longer restricted to `--all` and now works for single-session destroys as well.
> - Behavioral Change: headless single-session destroy without `--force` now fails with an explicit error instead of proceeding.
>
> 
> #### Changes since #1148 opened
>
> - Introduced VCS-aware workspace at-risk assessment in the daemon [87280ea]
> - Replaced `SessionDeltaRequest` enum with struct and introduced tagged `SessionDeltaResponse` enum in `minimald-rpc` [87280ea]
> - Implemented at-risk state classification and destroy gate logic in `minimal` client [87280ea]
> - Updated tests to verify new destroy gate behavior and at-risk classification [87280ea]
> - Updated end-to-end test script to reflect new destroy gate behavior [87280ea]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c10c935.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added confirmation prompts before destroying sessions.
  * Added `--force` support for destroying individual sessions or all sessions.
  * Display up to 10 files changed since session activation before confirmation.
  * Added workspace change detection for active sessions.

* **Bug Fixes**
  * Headless and no-input commands now safely refuse destruction unless `--force` is provided.
  * Unavailable change information now falls back to the standard confirmation prompt.

### DIFF
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -536,8 +536,8 @@ pub struct DestroyArgs {
     /// Destroy all sessions
     #[arg(long)]
     pub all: bool,
-    /// Skip confirmation when destroying all sessions
-    #[arg(long, short, requires = "all")]
+    /// Skip the destroy confirmation
+    #[arg(long, short)]
     pub force: bool,
 }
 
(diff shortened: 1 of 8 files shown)


norrietaylor:
The `native-daemon-e2e` failure was the feature firing: the e2e's kept-session teardown destroys headlessly without `--force`, which this PR now refuses. `c369897d` updates the teardown to assert the refusal (including that it names `--force`) and then destroy with `--force` — the breakage becomes the e2e proof of the gate.

crates/minimal/src/lib.rs
```diff
@@ -2316,9 +2316,100 @@ pub async fn cmd_destroy(global: &GlobalArgs, args: DestroyArgs) -> Result<(), a
         .context("a session or --all is required")?;
     let record = resolve_session(&mut client, session).await?;
 
+    match destroy_gate(args.force, global.no_input, std::io::stdin().is_terminal())? {
```
twitchyliquid64: No need for such verbosity, fold all those conditionals into one match here with different arms

crates/minimald-rpc/src/lib.rs
```diff
@@ -400,6 +400,42 @@ impl OneshotSshRpc for DestroySession {
...
+/// The request for a [`SessionDelta`] RPC.
+///
+/// Serialized examples:
+///
+///  * `{"name": "my-session"}`
+///  * `{"id": "<some-uuid>"}`
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SessionDeltaRequest {
+    Name(String),
```
twitchyliquid64: For listing sessions and fetching sessions it made sense for there to be a query path based on name or id.

In this case this is an internal RPC where you always know the session ID, so just take an ID and lookup via ID.

twitchyliquid64 (approved):

norrietaylor:
Revised per review — thanks @twitchyliquid64 for the catch that the activation-delta gate over-claimed: committed-and-pushed work still listed as "changed".

The gate now fires only for genuinely at-risk work:

- **VCS-exact when possible**: with a real root `.git` and a working `git`, the daemon reports uncommitted files (`git status --porcelain=v1 -unormal`) and unpushed commits (`rev-list --branches --not --remotes --count`). Committed-and-pushed sessions are proven clean and destroy without a word — no prompt, headless or not.
- **Honest fallback**: without usable VCS state the listing degrades to the activation delta with wording that no longer over-claims: "N files differ from activation (may include committed work):".
- **Conservative when unknowable**: a stopped session or failed RPC can prove neither dirt nor cleanliness, so it confirms without a listing (and refuses headless without `--force`). Asymmetry documented in the gate.
- Review follow-ups: the `SessionDelta` request is ID-only (callers have already resolved the record), and the gate is a single match over (force, at-risk state, interactivity). The pre-existing headless-destroy tests (`cli.rs`, session-e2e teardown) now assert the refusal-then-`--force` path they actually exercise.

Verified via cross (aarch64-musl): `session_delta` filter `test result: ok. 11 passed; 0 failed` (incl. the VCS ladder against a real repo + bare remote), `shell_exit` anchor `test result: ok. 1 passed; 0 failed`.

PR #1149, opened 2026-07-31
feat(minimal,minimald): first-prompt orientation banner for attached shells

norrietaylor:
An attached shell starts with zero orientation — which session, which loadout, how to detach; the first prompt now prints a two-line banner naming them, with a `min init` pointer when the project has no `minimal.toml`.

- Static templates + dynamic env vars: the daemon's launcher baseline seeds the session name and a once-only PROMPT_COMMAND banner; the client contributes the loadout list and blueprint flag as plain composed vars; the shell interpolates at print time.
- A loadout that sets its own `PROMPT_COMMAND` overrides the baseline banner cleanly (baseline sits beneath composition); the built-in default loadout's MOTD carries the same orientation lines after the mark.
- TTY-gated, prints once, `NO_COLOR`-safe.
- Daemon-side tests ride the Linux CI lanes.

> [!NOTE]
> ### Add first-prompt orientation banner to attached shell sessions
> - Injects `MINIMAL_SESSION_NAME`, `MINIMAL_LOADOUTS`, and `MINIMAL_BLUEPRINT` env vars into every launched session so the shell can render an orientation banner on the first prompt
> - The banner shows the session name, active loadouts, detach chord, and a conditional `min init` pointer when no `minimal.toml` is found
> - The built-in default loadout's `MINIMAL_MOTD` is updated to a static template that interpolates these vars; a baseline MOTD is also seeded in [session_host.rs](https://github.com/gominimal/minimal/pull/1149/files#diff-ffcaac1a766e1712e9d5e9d4b54b8abaeb63c6137b5865acf411a835d3a8e39e) for sessions without a custom loadout
> - Adds `loadout_display_list` and `push_orientation_vars` helpers in [loadouts.rs](https://github.com/gominimal/minimal/pull/1149/files#diff-ad282685b575b15a3ae014f466c2e2fdc73824ca16264b7603aec71ba7e5be3f), and a `blueprint_present` probe in [lib.rs](https://github.com/gominimal/minimal/pull/1149/files#diff-1b62c63af5a5bd2dbc4b550bf4bc6d0fa87e649b54dc3a587da5a8fa947b3ce1)
> - The e2e script in [session-e2e.sh](https://github.com/gominimal/minimal/pull/1149/files#diff-004d118857e0d3f2629dc99ab9f489236e0cbfcdb0c00fade94c342b824358a7) is made hermetic and extended to assert banner output
>
> 
> #### Changes since #1149 opened
>
> - Removed `blueprint_present` function from `minimal` crate and eliminated propagation of `MINIMAL_BLUEPRINT` environment variable through session composition [35c8677]
> - Replaced environment variable-based banner logic with in-shell filesystem checks in MOTD templates [35c8677]
> - Updated documentation and end-to-end test to reflect filesystem-based banner behavior [35c8677]
> - Added `Orientation` struct to `core::compose` module and extended `Composition` with an `orientation` field and accessor, allowing first-prompt banner facts (specifically `loadouts_display`) to be carried as control-plane data through the composition pipeline [2146a68]
> - Extended `WireContribution` and `WireComposition` with `orientation` field and introduced `WireOrientation` wire primitive to serialize and deserialize orientation data between client and daemon [2146a68]
> - Modified `session_baseline_env` function in `minimald` to accept `Option<&str>` for `loadouts_display` and conditionally seed `MINIMAL_LOADOUTS` environment variable when orientation data is present, and updated `SandboxLauncher::launch` to extract and pass orientation from composition [2146a68]
> - Refactored `compose_user_contribution` function in `crates/minimal` to accept `ActiveLoadouts` instead of `Vec<Loadout>`, compute loadout display list, and set orientation on composer using `UserComposer::with_orientation`, eliminating synthetic `MINIMAL_LOADOUTS` session variable [2146a68]
> - Extended `UserComposer` with `orientation` field and added `with_orientation` builder method to set first-prompt orientation facts, and updated `compose` method and `composition_to_wire` function to include orientation in resulting `WireContribution` [2146a68]
> - Updated documentation in `docs/reference/loadouts.md` to reflect that `MINIMAL_SESSION_NAME` and `MINIMAL_LOADOUTS` are seeded daemon-side from composition orientation field rather than client-side session variable [2146a68]
> - Imported `WireOrientation` from primitives and replaced `Default::default()` calls with `WireOrientation::default()` for the `WireContribution.orientation` field [2a98ee3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dcb9ce3.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added an interactive orientation banner showing the session name, active loadouts, detach guidance, and workspace setup status.
  * Built-in default loadouts now activate automatically when no configuration is provided.
  * Loadout information is preserved across session composition and reconnection.
  * User-defined default loadouts can replace the built-in default.

* **Bug Fixes**
  * Improved handling and display of selected loadouts during activation.
  * Ensured custom prompt settings correctly override banner defaults.

* **Documentation**
  * Expanded loadout documentation with banner behavior, customization, and configuration details.

### DIFF
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -1566,15 +1566,13 @@ pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(),
         let names: Vec<&str> = active.loadouts.iter().map(|l| l.name().as_ref()).collect();
         eprintln!("Applying loadouts: {}", names.join(", "));
     }
-    // The orientation banner's loadout list, contributed as a synthetic
-    // composed var the daemon-seeded banner template (or a loadout's
-    // own MOTD) interpolates in-shell at print time. The banner's other
-    // dynamic clause — blueprint presence — is a session-filesystem
-    // fact, tested by the templates in-shell when they print.
-    let loadout_display = loadouts::loadout_display_list(&active);
-    let (mut contribution, user_policy) =
-        loadouts::compose_user_contribution(active.loadouts, user_policy, compose_options)?;
-    loadouts::push_orientation_var(&mut contribution, &loadout_display);
+    // The contribution carries the banner's loadout display list as a
+    // first-class orientation field (the daemon seeds MINIMAL_LOADOUTS
+    // from it in the launcher baseline). The banner's other dynamic
+    // clause — blueprint presence — is a session-filesystem fact,
+    // tested by the templates in-shell when they print.
+    let (contribution, user_policy) =
+        loadouts::compose_user_contribution(active, user_policy, compose_options)?;
 
     // `--sync` defaults to tarball; `sync_explicit` records whether the
     // user actually typed the flag, which distinguishes a deliberate
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -1497,6 +1497,19 @@ fn resolve_upload_root(dir: &camino::Utf8Path) -> Result<camino::Utf8PathBuf, an
     }
 }
 
+/// Whether the project carries a blueprint (`minimal.toml`), probed with
+/// the same walk [`resolve_upload_root`] uses. Feeds the orientation
+/// banner's `MINIMAL_BLUEPRINT` var, which gates its `min init` pointer —
+/// so only a definite not-found counts as absent; a present-but-broken
+/// mfile is "present" (pointing its owner at `min init` would be wrong,
+/// and the upload path already fails loudly on it).
+fn blueprint_present(dir: &camino::Utf8Path) -> bool {
+    !matches!(
+        mfile::File::from_dir_recursive(dir.as_std_path()),
+        Err(mfile::Error::NotFound)
+    )
+}
+
 /// Create a new session via the `CreateSession` RPC.
 pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(), anyhow::Error> {
     ensure_daemon(global)?;
@@ -1562,12 +1575,21 @@ pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(),
     let compose_options = loadouts::compose_options_from_config(&cfg);
     let selection = loadouts::LoadoutSelection::from_flags(&args.loadout, args.no_loadouts);
     let active = loadouts::resolve_active_loadouts(selection, &cfg, global)?;
-    if !active.is_empty() {
-        let names: Vec<&str> = active.iter().map(|l| l.name().as_ref()).collect();
+    if !active.loadouts.is_empty() {
+        let names: Vec<&str> = active.loadouts.iter().map(|l| l.name().as_ref()).collect();
         eprintln!("Applying loadouts: {}", names.join(", "));
     }
-    let (contribution, user_policy) =
-        loadouts::compose_user_contribution(active, user_policy, compose_options)?;
+    // The orientation banner's dynamic halves, contributed as synthetic
+    // composed vars the daemon-seeded banner template (or a loadout's
+    // own MOTD) interpolates in-shell at print time.
+    let loadout_display = loadouts::loadout_display_list(&active);
+    let (mut contribution, user_policy) =
+        loadouts::compose_user_contribution(active.loadouts, user_policy, compose_options)?;
+    loadouts::push_orientation_vars(
+        &mut contribution,
+        &loadout_display,
+        blueprint_present(&utf8_path),
+    );
 
     // `--sync` defaults to tarball; `sync_explicit` records whether the
     // user actually typed the flag, which distinguishes a deliberate
--- a/crates/minimal/src/lib.rs
+++ b/crates/minimal/src/lib.rs
@@ -1497,19 +1497,6 @@ fn resolve_upload_root(dir: &camino::Utf8Path) -> Result<camino::Utf8PathBuf, an
     }
 }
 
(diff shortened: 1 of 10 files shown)


jtnkminimal (approved):

crates/minimal/src/lib.rs
```diff
@@ -1497,6 +1497,19 @@ fn resolve_upload_root(dir: &camino::Utf8Path) -> Result<camino::Utf8PathBuf, an
     }
 }
 
+/// Whether the project carries a blueprint (`minimal.toml`), probed with
+/// the same walk [`resolve_upload_root`] uses. Feeds the orientation
+/// banner's `MINIMAL_BLUEPRINT` var, which gates its `min init` pointer —
+/// so only a definite not-found counts as absent; a present-but-broken
+/// mfile is "present" (pointing its owner at `min init` would be wrong,
+/// and the upload path already fails loudly on it).
+fn blueprint_present(dir: &camino::Utf8Path) -> bool {
```
twitchyliquid64: Doesnt this need to happen in minimald? Theres no guarantee that the cwd matches the files in a session

norrietaylor: good eye

norrietaylor: this is minimal lib though right?

twitchyliquid64: minimal lib as in `lib.rs` in the minimal CLI / client

norrietaylor: riiiiiiiight

crates/minimal/src/lib.rs
```diff
@@ -1562,12 +1575,21 @@ pub async fn cmd_activate(global: &GlobalArgs, args: ActivateArgs) -> Result<(),
...
     }
-    let (contribution, user_policy) =
-        loadouts::compose_user_contribution(active, user_policy, compose_options)?;
+    // The orientation banner's dynamic halves, contributed as synthetic
+    // composed vars the daemon-seeded banner template (or a loadout's
+    // own MOTD) interpolates in-shell at print time.
+    let loadout_display = loadouts::loadout_display_list(&active);
+    let (mut contribution, user_policy) =
+        loadouts::compose_user_contribution(active.loadouts, user_policy, compose_options)?;
+    loadouts::push_orientation_vars(
```
twitchyliquid64: These data points should be first-class fields at the composition layer, otherwise we are mixing user env vars with control plane stuff. Handle collecting orientation type stuff in `UserComposer`, add fields to carry it in the corresponding wire type (I think `WireContribution`)

crates/minimal/src/loadouts.rs
```diff
@@ -102,22 +130,85 @@ pub(crate) fn resolve_active_loadouts(
...
+                carries_user_data: false,
+            },
+            source: WireSource::UserLoadout {
+                name: ORIENTATION_SOURCE.to_string(),
+            },
+        });
+    };
+    push("MINIMAL_LOADOUTS", loadout_display);
+    push(
+        "MINIMAL_BLUEPRINT",
```
twitchyliquid64: I would name this something descriptive, blueprint doesnt map to anything. Maybe `MINIMAL_HAS_LOCAL_MINIMAL_TOML`

crates/minimal/src/lib.rs
norrietaylor: Conceded, @twitchyliquid64 — the probe ran client-side against the host cwd at composition time, and the session workspace can diverge from it (skipped uploads, an in-session `min init`, later attaches from unrelated directories).

Fixed in 35c86775 by going one step further than moving the probe into minimald: `MINIMAL_BLUEPRINT` is gone entirely. The banner renders inside the sandbox, so both templates now test the workspace itself at the moment they print — `[ -f /workbench/minimal.toml ] || [ -f /workbench/.minimal/minimal.toml ]` (the daemon-side template derives the root from `sandbox2::SESSION_DEFAULT_WD`, the same constant that sets the attach shell's initial cwd, so it can't drift). That's fresher than a daemon-side probe at launch would be, and immune to in-session drift: a shell launched after an in-session `min init` shows the updated truth. `MINIMAL_LOADOUTS` stays client-contributed, being static composition truth only the client knows.

crates/minimal/src/lib.rs
norrietaylor: Done as prescribed in 2146a686, @twitchyliquid64 — orientation is now first-class at the composition layer, out of the user var lane entirely:

- `WireContribution` (and `Composition` + its persisted `WireComposition` sidecar, so a daemon restart keeps it) carries a serde-defaulted `orientation` field (`WireOrientation { loadouts_display }`); old peers interop — a payload without the field deserializes to the default and seeds nothing.
- `UserComposer` collects it (`with_orientation`) and emits it on the composed contribution; the `min task run` path routes through the same field.
- The daemon seeds `MINIMAL_LOADOUTS` into the launcher baseline env from the typed field, right beside `MINIMAL_SESSION_NAME` — all orientation env is daemon-baseline-seeded now, so user vars and user policy can never collide with it. Templates unchanged.

Re the naming comment on `MINIMAL_BLUEPRINT`: that var is gone entirely — the cwd-probe fix replaced it with an in-shell test of the workspace at print time — so the rename became moot.

twitchyliquid64 (approved):
Last request just trim down the comments, `session_baseline_env` just needs to say its injecting some env vars for the motd, `SESSION_WORKSPACE_ROOT` should just be one line, etc

PR #1162, opened 2026-08-03
test(fuzz): archive + paths targets, and four decoder hardening fixes

bryan-minimal:
A weekend fuzzing campaign on a Linux box (nightly + ASan — coverage this project has never had), landing two new fuzz targets and the four fixes they surfaced.

## New targets

| Target | Crate | Decodes | Trust |
|---|---|---|---|
| `archive_extract` | common | `extract_compressed_tar` — build sources, OCI layers, remote-cache artifacts | NET + SUPPLY |
| `path_invariants` | paths | the `AbsPath`/`RelPath`/`EitherPath` realm gates | NET/OWN |

`archive_extract` asserts *containment after every extraction* rather than trusting the return value — a successful extract that escaped the root trips neither a panic nor a sanitizer, and that is exactly the bug shape this target exists to catch. It is the surface that carried #651.

## Fixes

- **`common::archive`** — the tar link-target check only ran on the `strip_prefix: Some(..)` branch; an escaping symlink got through when no prefix was set.
- **`common::archive`** — cap the xz arm decompressed output (it is the one compression path that buffers the whole tar up front, so a decompression bomb was unbounded).
- **`paths`** — `EitherPath` could mint an unvalidated `RelPath`, bypassing the constructor gate.
- **`minvmd`** — bound oneshot RPC response reads.

## Notes for review

- `crates/paths/fuzz/` is a new fuzz workspace: added to `fuzz-crates` in **both** OS branches, with its own `[workspace]` and `.gitignore`, so `just fuzz-check` guards it.
- Seeds are generated (`scripts/gen-seeds.sh`) and committed — an unseeded byte fuzzer burns ~10^7 execs before stumbling onto a valid ustar header.
- Verified on this branch: `cargo fmt --check` clean, `cargo clippy -p common -p paths -p minvmd --all-targets -D warnings` clean, `cargo test -p common -p paths` 103 passing, `just fuzz-check` green.

A further fix from the same campaign (an SFTP/workspace-uploader chain) follows separately, along with a variant sweep for the same defect class.

## Summary by CodeRabbit

* **Security**
  * Archive extraction now limits decompressed XZ data and blocks links that escape the destination.
  * One-shot RPC responses are capped at 8 MiB.
  * Invalid relative paths are rejected more consistently.

* **Bug Fixes**
  * Improved handling of relative sandbox and working-directory paths.
  * Preserved archive safety checks across compression formats and path-prefix configurations.

* **Tests**
  * Added fuzzing coverage for archive extraction and path invariants.

* **Documentation**
  * Expanded fuzzing setup and coverage guidance.

> [!NOTE]
> ### Add fuzz targets for archive extraction and path invariants, and fix four decoder hardening bugs
> - Adds a `archive_extract` libFuzzer harness in [crates/common/fuzz](https://github.com/gominimal/minimal/pull/1162/files#diff-7640a35570198b51c5506dc7025c05e0c1a97efe1aaf1b1b7c1d6724312db447) that exercises `extract_compressed_tar` across five compression modes and four strip-prefix options, with a post-extraction containment check for symlinks and directory traversal.
> - Adds a `path_invariants` libFuzzer harness in [crates/paths/fuzz](https://github.com/gominimal/minimal/pull/1162/files#diff-9b5642b850ea8ca159d082d45cf7d100a5795e70aadf6f8c7cb92dd522803c8b) that differentially tests `RelPath`, `EitherPath`, and `FromStr` constructors for invariant consistency.
> - Fixes `extract_tar_impl` in [archive.rs](https://github.com/gominimal/minimal/pull/1162/files#diff-c3b59fcaa127815bc487babcda22a8fd129f71e318e19167b1b69ee729bc9daf) to skip escaping symlinks when no `strip_prefix` is set, resolve hardlinks relative to the destination directory instead of process CWD, and skip hardlinks whose targets are missing.
> - Caps xz decompression at 4 GiB via a new `LimitedWriter` wrapper; exceeding the cap returns `ArchiveError::DecompressedTooLarge` before any extraction.
> - Fixes `EitherPath::new` in [paths/src/lib.rs](https://github.com/gominimal/minimal/pull/1162/files#diff-94f295a0746ecfa31f504df7b5ad95cd29b5c2266dfd3a650e0e0250394c2c36) to store a plain `Utf8PathBuf` in the `Rel` variant instead of a `RelPath`, so inputs containing `..` no longer bypass `RelPath` validation. `as_rel` now returns `Option<&Utf8Path>`.
> - Fixes `rpc_client::call_oneshot` in [rpc_client.rs](https://github.com/gominimal/minimal/pull/1162/files#diff-faeefba21c73b8e96e46fcbd97caca0e8cf617515fe573c562c0cf700c5b2b53) to reject oneshot RPC responses larger than 8 MiB instead of buffering unboundedly.
> - Risk: `EitherPath::as_rel` return type changed from `Option<&RelPath<R>>` to `Option<&Utf8Path>`; callers relying on `RelPath` methods must re-validate.
>
> 
> #### Changes since #1162 opened
>
> - Hardened `paths::EitherPath` to reject relative paths containing parent directory segments by replacing infallible `new` constructor with fallible `try_new` method, changing the `Rel` variant to store `RelPath<R>` instead of `Utf8PathBuf`, and updating `FromStr` and serde `Deserialize` implementations to return validation errors [ddfd3e3]
> - Refactored `paths::CwdRelative` to store raw unresolved user input and allow climbing relative paths during resolution by replacing internal `EitherPath<R>` storage with `Utf8PathBuf` and `PhantomData<R>`, and updating `resolve` method to join climbing paths onto current working directory [ddfd3e3]
> - Modified `minimald` crate's `resolve_output` function to accept `&Utf8Path` instead of `SandboxPath`, allowing relative output paths with parent directory segments, explicitly rejecting empty strings, and normalizing absolute paths with `SandboxAbsPath::try_new` [ddfd3e3]
> - Updated `sessions` crate's `FileSet::walk_root` method to use fallible `HostPath::try_new` and return `None` when computed root contains disallowed segments, and modified `FileSet::walk_fs` to skip matching entries that cannot form valid `HostPath` instances [ddfd3e3]
> - Updated test code across `mfile`, `minimal`, `sessions` crates and integration tests to construct `HostPath` values using fallible `HostPath::try_new(...).unwrap()` instead of infallible `HostPath::new(...)` [ddfd3e3]
> - Updated documentation for `env::resolve_output` function [e374757]
> - Updated documentation for path types in `paths` module [e374757]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2f4772b.</sup>
> 
>
>


bryan-minimal:
Two follow-ups pushed.

**1. CodeRabbit review comment — fixed** (`8a89192b`). Confirmed real: the containment walk descended from `root` while comparing against `canonical_root`, so on macOS (`/var` -> `/private/var`) `escapes()` fired on in-tree symlinks. Linux `/tmp` is not symlinked, which is why the campaign that wrote the target never hit it.

**2. This PR introduced a hardlink escape; also fixed** (`4addf1e9`).

Unifying both `strip_prefix` branches onto the per-entry loop closed the symlink hole and opened a hardlink one, because the two tar APIs harden against opposite halves of the threat:

| | symlink | hardlink |
|---|---|---|
| `Archive::unpack` (the old `None` branch) | unsafe | **safe** — routes to `unpack_in`, `target_base = Some(dst)`, validates |
| `Entry::unpack` (per-entry loop) | safe (our check) | **unsafe** — `fields.unpack(None, ..)` uses the link name verbatim, so the kernel resolves it against the process **CWD** |

tar 0.4.46 comments on precisely this asymmetry in `entry.rs`. So the link-target check was necessary but not sufficient: a target of `etc/shadow` normalizes cleanly, passes, and then hardlinks the real file into the destination. Sharing an inode is an escape **no path-based containment can detect** — the resulting path really is inside the root — so it defeats the SFTP resolver and every other daemon check at once, and gives read *and* write.

Reachable from a remote cache mirror (`rcache::remote::materialize`) and upstream tarballs (`op::sources`) — both extract with `strip_prefix: None`, the branch that was hardlink-safe before this PR.

Fix creates hardlinks explicitly, anchored to `dest_dir`, mirroring `unpack_in`. The regression test asserts *through the inode* (plants a secret in the process CWD, requires the extracted entry not to read back as it) and is verified to fail without the change.

Worth flagging for the campaign: the `archive_extract` fuzz oracle is structurally blind to this class — it asserts path containment, and this escape has a contained path. A device+inode check would close that gap.

crates/paths/src/lib.rs
```diff
@@ -983,7 +983,20 @@ pub enum EitherPath<R: Realm> {
     /// Absolute variant.
     Abs(AbsPath<R>),
     /// Relative variant.
-    Rel(RelPath<R>),
+    ///
+    /// Deliberately a bare [`Utf8PathBuf`] and **not** a [`RelPath<R>`].
```
twitchyliquid64: This breaks the model, paths in the paths crate carry the marker type `Realm` to make it difficult to accidentally interpret a path representing a sandbox as something on a host/daemon filesystem, so realm needs to be carried.

Also, it seems like this change was motivated by the whole `..` thing - if thats the case then the fix is elsewhere depending on how RelPath is used/misused today:

1. upstream uses through RelPath that may have traversal need to be resolved before they hit here, this might be a good fit for the `CwdResolvable` / `CwdRelative` types here that were intended for that purpose.
2. Or if thats really not compatible with how its used today, then RelPath needs to be loosened to permit traversal

crates/paths/src/lib.rs
```diff
@@ -983,7 +983,20 @@ pub enum EitherPath<R: Realm> {
...
+    /// `EitherPath` is built from inputs that may legitimately contain `..` —
+    /// a `--output ../artifacts` typed in a sandbox cwd, a
+    /// `--minimal-state-dir ../state` on the daemon command line — so it
+    /// cannot uphold `RelPath`'s no-`..` guarantee. Holding a `RelPath` here
+    /// would mint one that [`RelPath::try_new`] would have rejected, and every
+    /// downstream holder of a `RelPath` is entitled to assume that cannot
+    /// happen. The realm is already carried by `EitherPath<R>` itself, so
```
twitchyliquid64: No, any nested use of the inner `Utf8PathBuf` for the rel variant looses the realm. The only reason its not a compile error for `EitherPath` to have the realm generic is because its used in the other variant, `Abs`. If Abs stopped being generic over realm, so would EitherPath.

So, its more correct to keep the realm as a generic on all variants.

bryan-minimal:
@twitchyliquid64 you're right, and the "got lazy" read is fair — the change weakened the abstraction to accommodate its misuse rather than fixing the misuse. Redone in `ddfd3e33`.

**What changed**

- `EitherPath::Rel` carries a validated `RelPath<R>` again, so the realm survives destructuring and the generic stops being justified by the `Abs` arm alone.
- The actual bad use was `EitherPath::new`: infallible, and struct-literalling **both** variants, so it forged an `AbsPath`/`RelPath` that `try_new` would have rejected. It's now `EitherPath::try_new`, routing through `AbsPath::try_new`/`RelPath::try_new`. `Deserialize` is fallible too, so a climbing path is refused at the wire boundary (that was `WireSource::Project`'s door) instead of becoming a forged `RelPath` downstream.
- Per your second option: a path that legitimately climbs is unresolved user input, so it belongs to `CwdRelative`. It no longer wraps `EitherPath` — that wrapping is what forced `Rel` to degrade in the first place — and owns the raw string, with `resolve()` as the only exit. `FromStr` stays `Infallible`, so clap is unaffected.

**Then rustc did the audit**, which is the part that made your point concrete:

- `minimald::env::resolve_output` took a `SandboxPath` purely to classify absolute-vs-relative, then normalized `..` itself two lines later. It never wanted the guarantee — it now takes the raw path, which says so honestly.
- `sessions`' glob literal-prefix extraction: a climbing prefix isn't a usable base, so it's dropped like an absent one.
- The remaining ~14 were absolute literals in tests.

The test that asserted the old behaviour now asserts the new one: `EitherPath::try_new("../../etc/passwd")` must be `Err(ContainsParentDir)`.

Green locally: `paths` 58, `sessions` 358, `mfile` 53, clippy + fmt clean. **`minimald`/`minimal` don't build on macOS (`procfs`), so the `resolve_output` change is by-inspection — relying on CI for it.**

---

**One design question I'd rather ask than guess at**, since it's the same shape as what you just pushed back on.

The plan for the remaining containment work (the SFTP escape and friends) is a primitive whose operations return **fds, bytes, or metadata — never a path**, so a caller can't obtain the thing they'd misuse. That mechanic seems aligned with what `paths` is doing: make misuse a compile error rather than a check someone forgets. Supporting evidence for the "retype the boundary" approach: `sftp.rs` was in nobody's audit list, but retyping `SessionPaths.working` would have surfaced it mechanically.

The open question is **placement**. The analysis argues for `common`, because `common`/`lcache`/`op` depend on neither `paths` nor camino, and `common::archive` can't pick a realm (`rcache::remote` calls it daemon-side). But that lands a second path-ish abstraction *outside* the realm system — which is arguably the same "stand outside the abstraction for convenience" move you just rejected, and risks being one-rule-two-implementations at the design level.

Options as I see them:
1. `common::contain` with no realm, `paths` re-exporting a realm-typed newtype for daemon callers.
2. Put it in `paths` and take the camino/realm dependency into `common`/`lcache`/`op`.
3. Something you'd rather — this is your model and I'd sooner build the version you'd keep.

Happy to spec (1) or (2) properly before writing any of it.

crates/minimald/src/env.rs
```diff
@@ -1165,21 +1165,25 @@ impl MaterializeArgs {
...
-                return Err("`--output` is empty".to_string());
-            }
-            if !cwd.is_absolute() {
-                return Err(format!(
-                    "cannot resolve `--output {rel}`: \
-                     the working directory `{cwd}` is not an absolute path"
-                ));
-            }
-            cwd.join(rel.as_utf8_path())
+/// Takes the raw user string, not a `SandboxPath`: `--output ../artifacts` is
```
twitchyliquid64: Docstring over commentary: just say it resolves a user path that might be relative to their cwd to the absolute path within the sandbox mount namespace

crates/paths/src/lib.rs
```diff
@@ -982,7 +982,15 @@ impl<'de, R: Realm> serde::Deserialize<'de> for RelPath<R> {
 pub enum EitherPath<R: Realm> {
     /// Absolute variant.
     Abs(AbsPath<R>),
-    /// Relative variant.
+    /// Relative variant — a validated [`RelPath<R>`], so it carries both the
+    /// realm and the no-`..` guarantee.
+    ///
+    /// It must stay a `RelPath` and not a bare path: the realm marker is the
```
twitchyliquid64: Drop the commentary with claude talking to itself about the current pr

twitchyliquid64 (approved):
LGTM, caveat nits about commentary in code comments

bryan-minimal:
Both fixed in `e374757c` — and taken as the general note, not just the two lines.

The pattern was doc comments arguing against a shape that is no longer in the tree: the `Rel` variant explaining why it is *not* a bare path, `EitherPath::try_new` explaining what the old infallible `new` got wrong, `CwdRelative` explaining what it stopped wrapping. Nobody reading the merged code knows any of that existed, so it reads as me narrating the PR rather than documenting the type. Replaced with what each thing is and when to reach for it.

`resolve_output` already had a docstring that said it plainly; I had appended a second paragraph underneath it. Dropped that and kept the original, reworded to name the sandbox mount namespace as you suggested.

Net -35/+12. Tests and clippy green.

PR #1164, opened 2026-08-03
feat(minimald)!: rebase sftp paths at / to support accessing /home

twitchyliquid64:
Rebases paths over sftp as rooted at the same tree as in the corresponding session, so that `/home` and `/workbench` are both accessible. Previously, `/` mapped to `/workbench` which was limiting.

## Summary by CodeRabbit

* **New Features**
  * SFTP sessions now provide separate client-visible roots for workspace and home directories.
  * Added consistent path normalization and synthetic root directory listings.
* **Bug Fixes**
  * Improved protection against directory traversal, symlink escapes, unsafe root operations, and insufficient permissions.
  * Git status checks now verify the resolved work tree, including nested repositories and symlinks.
* **Improvements**
  * Enhanced metadata, file access, directory operations, and real paths across exported locations.
  * Updated session documentation and manifest examples to use the workspace path.

> [!NOTE]
> ### Rebase SFTP paths at `/` to expose both `/workbench` and `/home` exports
> - Introduces a synthetic SFTP root (`/`) containing two exports: `/workbench` (session working tree) and `/home` (user home directory), replacing the previous single-root layout.
> - Rewrites path resolution in [`sftp.rs`](https://github.com/gominimal/minimal/pull/1164/files#diff-dca5edb40762c2e0c2a4bb8639425db7df2be3f39afaf201e36238e56b9b4611) to normalize client paths textually, map them to the correct export, and enforce containment via a symlink-safe `contained` walk.
> - Adds `resolve_below_export` to reject destructive operations (remove, rename, rmdir) targeting the export roots themselves.
> - Updates the integration test upload path from `/minimal.toml` to `/workbench/minimal.toml` to match the new layout.
> - Risk: Breaking change — SFTP clients that referenced paths directly under `/` (e.g. `/foo`) must now use `/workbench/foo`; symlinked ancestors or dangling symlink leaves inside exports are refused with `PermissionDenied`.
>
> 
> #### Changes since gominimal/minimal#1164 opened
>
> - Modified `russh_sftp::server::Handler::setstat` implementation for `SftpSession` to enforce minimum owner permissions by performing `fstat` on the file handle to determine if target is a directory, then calling a new helper function to apply a permission floor of owner read for files or owner read-execute for directories before applying permissions to the filesystem [da91cab]
> - Added `keep_owner_access` function that accepts `std::fs::Permissions` and a directory boolean flag, returning modified permissions with an enforced minimum of owner read-execute for directories or owner read for files while preserving all other permission bits [da91cab]
> - Added `sftp_setstat_leaves_the_owner_able_to_reach_what_it_chmodded` integration test that creates a session, writes test files and directories, exercises `setstat` with various permission modes including 0o000, 0o644, 0o751, and 0o444, and validates that resulting permissions are floored to 0o400 for files and 0o500 for directories when restrictive modes are requested while preserving requested group and other bits [da91cab]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 52b6979.</sup>
> 
>
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3810,7 +3810,6 @@ dependencies = [
  "op",
  "orchestrator",
  "ot",
- "path-absolutize",
  "paths",
  "rand_core 0.6.4",
  "rcgen",
--- a/crates/minimald/Cargo.toml
+++ b/crates/minimald/Cargo.toml
@@ -43,7 +43,6 @@ libc.workspace = true
 nix.workspace = true
 russh.workspace = true
 russh-sftp.workspace = true
-path-absolutize.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sysinfo.workspace = true
--- a/crates/minimald/src/env.rs
+++ b/crates/minimald/src/env.rs
@@ -53,12 +53,12 @@ const MIN_SCRIPT: &str = include_str!("env_min_helper.sh");
 /// Where the session's workspace appears *inside* the sandbox. The daemon sees
 /// the same directory at [`SessionChannel::working`], so this is the prefix
 /// that translates a path typed in the sandbox into one the daemon can write.
-const WORKSPACE_ROOT: &str = constcat::concat!("/", sandbox2::SESSION_DEFAULT_WD);
+pub(crate) const WORKSPACE_ROOT: &str = constcat::concat!("/", sandbox2::SESSION_DEFAULT_WD);
 
 /// Where the session's home directory appears inside the sandbox, backed by
 /// [`SessionChannel::home`]. The second of the two directories a sandbox path
 /// can name on the daemon's side.
-const HOME_ROOT: &str = constcat::concat!("/", sandbox2::SESSION_HOME);
+pub(crate) const HOME_ROOT: &str = constcat::concat!("/", sandbox2::SESSION_HOME);
 
 /// The parameters used to construct an [`Env`].
 ///
@@ -1191,7 +1191,7 @@ fn resolve_output(cwd: &Utf8Path, output: SandboxPath) -> Result<SandboxAbsPath,
 
 /// Resolves `.` and `..` textually. A `..` at the root stays there, so an
 /// escape can only show up as a path failing the workspace prefix check.
-fn normalize_absolute(path: &Utf8Path) -> Utf8PathBuf {
+pub(crate) fn normalize_absolute(path: &Utf8Path) -> Utf8PathBuf {
     let mut out = Utf8PathBuf::new();
     for component in path.components() {
         match component {
--- a/crates/minimald/src/session_delta.rs
+++ b/crates/minimald/src/session_delta.rs
@@ -169,6 +169,14 @@ async fn vcs_at_risk(root: &Path) -> Option<minimald_rpc::SessionDeltaResponse> 
     if !root.join(".git").exists() {
         return None;
     }
+    // A `.git` entry is only a marker. When it is not a real repository, git
+    // does not fail: it keeps walking up and answers for the nearest enclosing
+    // repo, which would report an unrelated tree's uncommitted files as this
+    // session's work at risk. So the work tree git found has to be this one.
+    let top = run_git(root, &["rev-parse", "--show-toplevel"]).await?;
+    if !is_same_dir(Path::new(top.trim()), root) {
+        return None;
+    }
     let status = run_git(root, &["status", "--porcelain=v1", "-unormal"]).await?;
     let count = run_git(
         root,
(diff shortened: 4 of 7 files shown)


bryan-minimal:
Reviewed with the escape from the gominimal/minimal#1162 campaign in mind (the `resolve()` `NotFound` branch that fell back to a merely lexical check, letting a planted `workspace/link -> /` be written through). **This closes it**, and closes it the right way rather than by adding another check:

- Roots canonicalized once at construction, with the `/var` -> `/private/var` reason stated — so both sides of every comparison are symlink-free.
- The deepest-existing-ancestor walk, canonicalizing *that* and re-appending the missing tail, is the thing the old code got wrong: it refuses a symlinked parent **before** anything is created through it, which canonicalizing only existing paths cannot do. Same shape as `env::create_artifact`, which is the one implementation in the tree that already had this right.
- Termination is argued explicitly rather than assumed.

Two things, one worth acting on:

**1. No `O_NOFOLLOW` on the final open (`opts.open(host.as_str())`).**

`resolve()` proves the path is contained, then the open re-resolves it by name. Between those two steps the leaf can be swapped for a symlink, and the open follows it. That is a narrow race, but the racer here is not remote — `sandbox2` bind-mounts **workspace and home read-write into the sandbox** (`crates/sandbox2/src/lib.rs:713-740`), so an in-session shell can spin on it locally against a daemon-uid write.

`.custom_flags(libc::O_NOFOLLOW)` on the `OpenOptions` closes the leaf case for the cost of one flag. It does not close a mid-path swap, but the ancestor walk already covers the interesting part of that, and the leaf is the one an attacker can actually aim at.

Same consideration for the `mkdir` / `rename` / `remove` / `setstat` handlers if they re-resolve by name after the check.

**2. Observation, not an objection — the export surface roughly doubles.**

`/` previously mapped to `/workbench` alone; it now exposes `home` as well. Both are RW bind-mounted into the sandbox, so the plant-a-symlink primitive now applies to both roots. The containment logic looks symmetric across the two (both canonicalized, both routed through the same `resolve_below_export`), so I think this is fine — flagging it so the symmetry is a deliberate invariant rather than an accident, since a future third export root would need the same treatment.

---

Context that may be useful: this bug class has now recurred six times in the tree as "one containment rule, two implementations, only one hardened" — most recently in gominimal/minimal#1162, where switching tar extraction to a per-entry loop closed the symlink hole and opened a hardlink one (`Entry::unpack` passes `target_base: None`, so the kernel resolves hardlink targets against the **process CWD**; `Archive::unpack` routed through `unpack_in` and did not). Tracked in gominimal/inbox#534, including a proposed `clippy.toml` `disallowed-methods` entry that would have caught it at compile time.

Two related items from the same audit are still open and independent of this PR: `op::materialize::extract_raw_file` (`crates/op/src/materialize.rs:256`) has **no** containment check at all — `[outputs.x] type="raw-file" path="../../../etc/shadow"` streams daemon-side bytes to the client — and `lcache::LocalDir`'s write paths are lexical-only. Happy to take those.

bryan-minimal:
Re-reviewed after the update. The `O_NOFOLLOW` work looks right, and the reasoning is the part I'd have gotten wrong:

> `resolve` hands back a canonical path, so a real symlink is already resolved to its target by the time we get here and there is no link left to refuse.

That's the argument that makes `O_NOFOLLOW` safe here rather than a behaviour regression. Adding it to a handler that resolves by name *without* canonicalizing would break legitimate symlink access; because `resolve` canonicalizes, the flag can only ever reject a leaf that appeared **after** the check — which is exactly the race and nothing else. Good, and the comment says so plainly, including that a mid-path swap is the ancestor walk's job rather than this flag's.

**The `O_NONBLOCK` catch is better than what I raised.** A FIFO planted in the workspace makes a read-only open block until a writer appears, wedging the dispatch loop — a denial of service that needs no traversal at all, just a `mkfifo` from the in-session shell. I flagged the symlink race and missed that one entirely.

The remaining handlers check out for the reasons that make them uninteresting rather than because of new code, which is worth stating so it's a deliberate conclusion:

- `unlink` / `rename` operate on the link itself, not its target, so a swapped-in leaf gets removed/renamed rather than followed.
- `mkdir` onto an existing symlink is `EEXIST`; a symlinked *parent* is caught by the ancestor walk.
- `stat` uses `symlink_metadata` on an already-canonical path, so it is equivalent to `stat` without the following.

**One residual, lower severity than the open, and possibly a deliberate accept:** `fs::read_dir(host.as_str())` still opens by path and follows the leaf. Same window as the `open` had — swap the resolved directory for a symlink between `resolve` and `read_dir` and the listing comes from outside the export root. It is an information disclosure rather than a write, and std has no `O_NOFOLLOW` equivalent for `read_dir` (it needs `O_DIRECTORY|O_NOFOLLOW` + `fdopendir`, i.e. `nix`), so "not worth the machinery for a listing" is a reasonable call — I'd just want it to be a stated one rather than an oversight, since the next reader will notice the asymmetry with `open`.

Nothing blocking from me. LGTM.

bryan-minimal (approved):

PR #1178, opened 2026-08-05
fix(op): use a BTreeSet for override_deps so rootfs assembly is deterministic

bryan-minimal:
## Summary

`SpecBuild.override_deps` was an `Option<HashSet<PathBuf>>`. That collection feeds rootfs assembly, which hardlinks entries first-writer-wins, so its iteration order decides which package provides a path that several packages install — and `HashSet` iteration follows a per-process `RandomState` seed, so the winner changed from one run of the binary to the next.

## Change

`Option<BTreeSet<PathBuf>>`: iteration is deterministic across processes, first write still wins. Packages installing overlapping paths remains a packaging bug to fix in the packages themselves — this only keeps the symptom deterministic (per review: no caller-chosen precedence).

Companion: #1180 makes the downstream `sandbox2::Config::rootfs` collection ordered too, which covers every rootfs producer.

### DIFF
--- a/crates/op/src/specs.rs
+++ b/crates/op/src/specs.rs
@@ -33,9 +33,8 @@ pub struct SpecBuild<'a, SF: crate::SourceFetcher> {
     /// Does not override `Local` or `Source` variant build_deps - those will
     /// always be made present automatically.
     ///
-    /// Ordered, because `rootfs_mapped` passes these to the sandbox, which hardlinks them
-    /// first-writer-wins: with a `HashSet` the winner of a path two deps both install was
-    /// decided by a per-process `RandomState` seed and so varied between runs.
+    /// Ordered: `rootfs_mapped` passes these to the sandbox, which hardlinks
+    /// them first-writer-wins, so iteration order must be deterministic.
     pub override_deps: Option<BTreeSet<PathBuf>>,
 
     /// Optional async writer that receives a copy of the sandbox's stdout stream.
--- a/crates/op/src/specs.rs
+++ b/crates/op/src/specs.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashSet, path::PathBuf, time::Instant};
+use std::{
+    collections::{BTreeSet, HashSet},
+    path::PathBuf,
+    time::Instant,
+};
 
 use crate::{Error, Materialized, Options, Runnable, SubsetBuild};
 use anyhow::anyhow;
@@ -28,7 +32,11 @@ pub struct SpecBuild<'a, SF: crate::SourceFetcher> {
     ///
     /// Does not override `Local` or `Source` variant build_deps - those will
     /// always be made present automatically.
-    pub override_deps: Option<HashSet<PathBuf>>,
+    ///
+    /// Ordered, because `rootfs_mapped` passes these to the sandbox, which hardlinks them
+    /// first-writer-wins: with a `HashSet` the winner of a path two deps both install was
+    /// decided by a per-process `RandomState` seed and so varied between runs.
+    pub override_deps: Option<BTreeSet<PathBuf>>,
 
     /// Optional async writer that receives a copy of the sandbox's stdout stream.
     pub stdout_writer: Option<Box<dyn tokio::io::AsyncWrite + Unpin + Send + Sync>>,
--- a/crates/op/src/patched.rs
+++ b/crates/op/src/patched.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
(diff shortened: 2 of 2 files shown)


twitchyliquid64 (changes requested):
Just change `HashSet` to `BTreeSet`, deterministic iteration order makes sense but i dont think we want to make it possible for someone to choose 'winning' packages because its an issue if files overlap and that should be fixed downstream.

crates/op/src/specs.rs
```diff
@@ -28,7 +32,17 @@ pub struct SpecBuild<'a, SF: crate::SourceFetcher> {
...
+    /// sandbox hardlinks them first-writer-wins, so position decides which provider wins a
+    /// contended path. Measured: mes-libc (36 headers), musl (90) and glibc (109) all ship
+    /// `/usr/include/stdio.h`; a bootstrap rung needs ITS OWN libc to win, and the recipe
+    /// cannot disambiguate because all three land on the same path.
+    /// This was a `HashSet`, whose RandomState is seeded per-process — the winner therefore
+    /// changed on every RUN, which is the stage0 "mes-libc SIGSEGV lottery" that went
+    /// undiagnosed for two months. A `BTreeSet` made it deterministic but sorted by path,
+    /// which picks arbitrarily. A Vec preserves the caller's order, so precedence is a
+    /// decision the caller makes rather than an accident of hashing or alphabetisation.
+    pub override_deps: Option<Vec<PathBuf>>,
```
twitchyliquid64: This should be a BTreeSet, also trim commentary which is now mostly incorrect

crates/op/src/specs.rs
```diff
@@ -55,7 +69,7 @@ impl<'a, SF: crate::SourceFetcher> SpecBuild<'a, SF> {
         for input in build.build_deps.iter() {
             match input {
                 BuildDep::Local { full_path, .. } => {
-                    build_deps.push(SandboxMapped::FileCopy(full_path.to_path_buf()))
+                    build_deps.push(SandboxMapped::File(full_path.to_path_buf()))
```
twitchyliquid64: This will break builds where cwd is on a different filesystem to the minimal / cache dir

bryan-minimal:
Done — rebuilt the branch on current `main` and reduced it to just the `BTreeSet` change:
`specs.rs` field type plus the collection in `patched.rs`, +12/−4. Commentary trimmed to
two lines. `local_backend.rs` didn't need touching after all, its `.collect()` infers from
the field.

Rebuilding on `main` also dropped three accidental reverts my older base had carried in,
including the `FileCopy` → `File` one you flagged on line 72 — good catch, that would have
broken exactly the cross-filesystem case you describe.

On the wider point, agreed and taking it downstream. For context on why I'd reached for
caller-chosen order: in our bootstrap ladder three libc packages each install
`/usr/include/stdio.h` (mes-libc, musl, glibc), and a stage0 build needs its own to win.
An earlier attempt of mine that ordered by path was deterministic but selected the wrong
one and failed 9/9 runs. Ordering by cache path is likewise unrelated to which libc the
build needs, so I'd expect this PR to make those builds consistent rather than
consistently correct. I'll verify that by rebuilding the affected rungs and will fix the
overlap in the specs — the real bug is that all three libcs are in that sandbox at all.

One related spot I left alone: `rootfs_mapped`'s own `HashSet<SandboxMapped>` on the
non-override path has the same run-to-run variance. Converting it needs `Ord` on
`SandboxMapped`. Happy to add it here or in a separate PR — whichever you prefer.

twitchyliquid64:
Feel free to fix rootfs_mapped and add `Ord` to `SandboxMapped`

crates/op/src/specs.rs
```diff
@@ -28,7 +32,11 @@ pub struct SpecBuild<'a, SF: crate::SourceFetcher> {
     ///
     /// Does not override `Local` or `Source` variant build_deps - those will
     /// always be made present automatically.
-    pub override_deps: Option<HashSet<PathBuf>>,
+    ///
+    /// Ordered, because `rootfs_mapped` passes these to the sandbox, which hardlinks them
+    /// first-writer-wins: with a `HashSet` the winner of a path two deps both install was
```
twitchyliquid64: Now that HashSet is gone theres no need to mention it, that doesnt help a future reader

bryan-minimal: Trimmed in the follow-up commit — it now states just the property (ordered because assembly is first-writer-wins), no history.

twitchyliquid64 (approved):
Honestly theres a lot of cruft in here, maybe rebase and just

PR #1180, opened 2026-08-05
fix(op): sort standalone-test dependencies so the test rootfs is deterministic

bryan-minimal:
## Summary

Same class of issue as #1178, on the standalone-test path.

`StandaloneTest::dependencies` collects into a `HashSet<SandboxMapped>`, fed by an unsorted
`HashMap` iteration. The sandbox hardlinks these first-writer-wins, so when two
dependencies install the same path the winner is decided by iteration order, which comes
from a per-process `RandomState` seed. A test therefore ran against a different rootfs from
one invocation to the next.

## Change

Collect into a `Vec` sorted by spec hash, deduplicating by path and keeping the first
occurrence. One file.

## Why this one is worth doing

Unlike a build, a test that composes its rootfs differently can silently report the wrong
result rather than failing.

Concretely: `packages/gawk`'s `matching` test has `test_deps = [base-bootstrap]`, and that
closure contains `gawk-bootstrap` 5.3.2, which installs `usr/bin/gawk`, `usr/bin/awk` and
`usr/lib/gawk/*.so` — the same paths gawk 5.4.0 installs. `sandbox2` also symlinks
`/bin → usr/bin`, so command lookup resolves to whichever won. On an unlucky iteration
`minimal check gawk` asserts against the 5.3.2 binary and reports a pass for 5.4.0; it
would report a pass even if 5.4.0 could not start. `packages/mtools` has the same shape
with `bash`.

This PR makes that outcome consistent rather than random. It does not decide *which*
package ought to win — as with #1178, the overlapping paths are the underlying problem and
belong in the specs. Filing this separately so the variance is gone while that is sorted
out.

## Note

The queue builder does not run standalone tests, so no published artifact depends on these
verdicts today.

> [!NOTE]
> ### Sort standalone-test rootfs dependencies deterministically using `BTreeSet`
> - Replaces `HashSet<SandboxMapped>` with `BTreeSet<SandboxMapped>` in both [`Config.rootfs`](https://github.com/gominimal/minimal/pull/1180/files#diff-cae0fcb053d236a17a0fa33c43b824a9dba69e6c3b14956501a09d010016b3d5) and [`StandaloneTest.dependencies`](https://github.com/gominimal/minimal/pull/1180/files#diff-f45bf8f82ce4c76fd24aeaa1b80efccfd3a2fdfdf1f8462b3b3684998c6eddce) to give deterministic iteration order.
> - Adds `PartialOrd`/`Ord` implementations for `SandboxMapped`, ordering by variant (`File < Dir < TempDir < FileCopy`) then by associated path.
> - Behavioral Change: anything iterating over `Config.rootfs` now sees a stable, sorted order instead of hash-determined order.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized afdbc22.</sup>
> 
>

### DIFF
--- a/crates/op/src/standalone_test.rs
+++ b/crates/op/src/standalone_test.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
 use anyhow::anyhow;
 use graph::{BuildSpec, BuildSpecRef, SpecTest, SubsetInput, Transitives};
@@ -34,7 +34,7 @@ impl<'a> StandaloneTest<'a> {
         build: &BuildSpec,
         test: &SpecTest,
         opts: &Options<'a>,
-    ) -> Result<(Vec<SandboxMapped>, bool, bool), Error> {
+    ) -> Result<(BTreeSet<SandboxMapped>, bool, bool), Error> {
         let transitives = Transitives::for_toplevels(
             opts.graph,
             {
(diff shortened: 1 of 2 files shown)


twitchyliquid64 (commented):
Theres no need to sort and/or remove dupes if the upstream collection of paths into a rootfs is a `BTreeSet` - it will be deterministic order, and first write will win. A lot of this diff seems superfluous (tho s/HashSet/BTreeSet seems fine)

bryan-minimal:
Done in bbab0ff3 — took it one level further per your note: `Config::rootfs` itself is now a `BTreeSet` (with an `Ord` on `SandboxMapped` mirroring its manual `Eq`/`Hash` key), so assembly is deterministic for every producer, not just this call site. The op-layer sort/dedup is reverted; the standalone-test path keeps the plain `HashSet`→`BTreeSet` swap.

twitchyliquid64 (approved):

Issue #1192, opened 2026-08-10
Replace lzma-rs with a streaming xz decoder (lzma-rust2) — recommendation, open for review


bryan-minimal:
# Replace `lzma-rs` with a streaming xz decoder (`lzma-rust2`) — recommendation, open for review

*Assessment + recommendation. The decision is not yet made — this is a proposal to weigh before doing the work.*

## TL;DR

`common::archive::extract_compressed_tar` uses `lzma-rs 0.3` for `.tar.xz`. Two independent reasons to move off it, weighed separately because their strength differs a lot:

- **Weak (nearly cosmetic):** a fuzzer found a reachable arithmetic-overflow panic in `lzma-rs`. It's benign in release (overflow-checks off → wraps → clean `XzError`), so it is **not a vulnerability in shipped `minimal`**. But it is unfixed upstream, and it currently **blocks the xz branch of the `archive_extract` fuzz target** (skipped precisely because it panics under overflow-checks).
- **Strong (structural):** `lzma-rs` has **no streaming decoder**. That single API gap is the direct cause of the only awkward arm in our extraction path — the xz arm must materialize the whole decompressed tar to a tempfile before validating a single entry, which is why it needed a bespoke 4 GiB cap + counting writer that no other format needs.

**Recommendation: move to [`lzma-rust2`](https://github.com/hasenbanck/lzma-rust2)** — pure Rust (matters for our musl cross-builds), actively maintained + fuzzed, and it provides `XzDecoder<R>: Read` so the xz arm can stream like gzip/zstd/bz2 and we can *delete* the tempfile, the cap, and the counting writer. That's a simplification, not just a swap.

## Current state (verified against `main`, 2026-08-10)

| Piece | Status on main |
|---|---|
| Disk-fill cap (`MAX_XZ_DECOMPRESSED_BYTES` 4 GiB + `LimitedWriter`) | **Present** (#1162, `archive.rs:18,271`) |
| Panic containment (`catch_unwind` around the xz decode) | **Merged** — #1193 (`fix(common): contain lzma_rs panics on malformed xz streams`), reproducer committed |
| xz branch in the `archive_extract` fuzz target | **Re-enabled** — #1193 (the containment made it fuzzable) |
| The overflow reported upstream | **No** — no `lzma-rs` issue, no RustSec advisory (`cargo deny check advisories` passes) |

## The bug

[...]

Issue #2, opened 2025-08-19
Cycle detection


brandonweeks:
Creating a cycle in the graph causes `minimal` to use all available RAM and then crash my machine.

Expected behavior: `minimal` doesn't crash my machine, that's my job.

PR #1193, opened 2026-08-10
test(fuzz): campaign 2 — tarball/xz/path/redact hardening + minimald remote-write fix

bryan-minimal:
## Summary

Fuzzing campaign gominimal/minimal#2: hardens the untrusted-input decode paths (tar/xz extraction, path normalization, diagnostics redaction) and hardens the `minimald` workspace uploader against escaping links. Eight commits, kept separate on purpose — each is an independent fix or target with its own rationale.

## Workspace-uploader hardening

**`fix(minimald): validate the uploaded workspace tarball entry by entry`** — `unpack_workspace_files` handed the client's tarball to `async_tar::Archive::unpack`, which writes link targets verbatim, so `link -> /` plants an escaping symlink in the shared workspace. Defense-in-depth: the sftp path already canonicalizes and refuses to follow such a link (`SftpSession::contained` — thanks @twitchyliquid64 for the correction; this is **not** a live remote-write), but the workspace's *other* consumers (sandbox mount, attach shell, `op::materialize`) shouldn't have to rely on each canonicalizing — so don't plant it. Adds `unpack_validated` (escaping entry paths fatal — tar-slip; escaping link targets skipped-with-warning so legitimate tarballs still extract), unifying the symlink-vs-hardlink base rule through `normalize_link_target` (one definition instead of two). `unpack_workspace_patches` already validated per-entry; this sibling path didn't.

## The lzma containment (closes the loop on the assessment)

**`fix(common): contain lzma_rs panics on malformed xz streams`** — this is the `catch_unwind` guard referenced as the "forthcoming PR" in the `lzma-rs` assessment issue, **gominimal/minimal#1192**. `extract_compressed_tar` promises malformed → `ArchiveError`, never panic; `lzma_rs`'s `backward_size + 1` footer check overflows at `u32::MAX`. Release wraps harmlessly (overflow-checks off) but any overflow-checked build — tests, dev, the fuzz target — aborts, which is why the xz fuzz arm was dark. Guarded here + committed reproducer (`xz_backward_size_overflow.tar.xz`). Makes the "never panics" contract independent of the dependency, and is the safety net that lets gominimal/minimal#1192's decoder swap proceed without re-litigating panic-safety.

## Fuzz infrastructure + oracles

- **`fix(fuzz): mirror the root workspace patch into each fuzz workspace`** — `just fuzz-check` had been silently failing since the topiary patch landed: each `fuzz/` dir is its own `[workspace]` (the isolation that keeps the nightly/sanitizer build off the main tree), and `[patch]` doesn't cross a workspace boundary, so four of six fuzz workspaces stopped building. `just ci` doesn't run `fuzz-check` — the exact silent rot the docs warn about.
- **`test(fuzz): give archive_extract an inode oracle`** — a path oracle structurally cannot see a hardlink escape (the escape is by *identity*, not location — a hardlink is a second name for one inode). Adds a sentinel-inode check; moves the sandbox under `fuzz/target` so cross-device EXDEV doesn't silently mask escapes. Retargets `path_invariants` at the `EitherPath::try_new` routing claim.
- **`test(common): fuzz normalize_within_root directly`** — three crates now trust this contract and none re-checks it; fuzzed directly it's a pure function (150k vs 1.7k exec/s) and explores path shapes a ustar header can't encode (incl. `/srv/work` vs `/srv/workbench`, where a string-prefix check wrongly passes). 96M execs, no findings.
- **`test(diagnostics): fuzz redact for leaks, not just panics`** — asserts the module's stated asymmetry (masking a harmless value is fine; leaking a secret is not): no leaf under a sensitive key survives redaction, across JSON and TOML. Deliberately monotonic, not idempotent (a second pass re-masks the placeholder by key — asserting idempotence fired wrongly against behavior the module never promised). 18.1M execs, no findings.
- **`docs(fuzzing)`** and **`fix(diagnostics): reject redacted TOML that will not parse back`** round it out.

## Notes

- The lzma commit fixes a real never-panic contract violation; the minimald commit is defense-in-depth on the shared workspace; the rest restore and deepen fuzz coverage that had silently rotted.
- CI is the compiler here — the campaign was developed and fuzzed locally; the series is content-identical to that work.

## Summary by CodeRabbit

- **Bug Fixes**
  - Malformed XZ archives now return a clear error instead of crashing.
  - Archive extraction better prevents path and hardlink escapes.
  - Redacted TOML output is validated to ensure it remains parseable.

- **Tests**
  - Added fuzz coverage for path normalization, archive extraction, diagnostics redaction, and path invariants.
  - Expanded checks for containment, idempotency, masking, and malformed inputs.

- **Documentation**
  - Added guidance for reviewing security claims and designing fuzzing checks.

- **Chores**
  - Included diagnostics in macOS fuzzing workflows.

> [!NOTE]
> ### Harden tarball extraction, XZ decompression, and redaction fuzzing in campaign 2
> - Wraps `lzma_rs::xz_decompress` in `catch_unwind` in [`extract_compressed_tar`](https://github.com/gominimal/minimal/pull/1193/files#diff-c3b59fcaa127815bc487babcda22a8fd129f71e318e19167b1b69ee729bc9daf) so malformed XZ streams return a `CompressionError`/`IO` error instead of panicking.
> - Adds inode-escape detection to the [`archive_extract`](https://github.com/gominimal/minimal/pull/1193/files#diff-626a054431c2aa4f5622cca37c502d54229e41cdaacd5c5e8eee0e2be3077008) fuzz harness by writing a sentinel file and asserting no extracted entry shares its inode/device.
> - Adds a new [`normalize_within_root`](https://github.com/gominimal/minimal/pull/1193/files#diff-fdd4c7f50694ecbede2de8d7e8e0d1231ba2146a078c2db9266197ca4aef2e45) fuzz target asserting path normalization is relative, parent-dir-free, non-escaping, and idempotent.
> - Adds a new [`redact_roundtrip`](https://github.com/gominimal/minimal/pull/1193/files#diff-b8e7e4ece57c3ebada3873d0910fd3527a65bd45a6b05bb32b8c761dd3685079) fuzz target asserting sensitive keys are masked and outputs remain parseable; `redact_toml` now validates its output re-parses as TOML.
> - Updates [`path_invariants`](https://github.com/gominimal/minimal/pull/1193/files#diff-23e17690f176691ab18a1c37238e07a43134b009a9d54e92213b5c176eaf7d88) to use fallible `EitherPath::try_new` and tighten consistency checks.
> - Risk: `redact_toml` can now return an error when the pretty-printed output fails TOML re-parsing, which is a new failure mode for callers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 239ea40.</sup>
> 
>


crates/minimald/src/rpc.rs
```diff
@@ -935,14 +935,91 @@ async fn unpack_workspace_files(
...
 
     Ok(())
 }
 
+/// Unpacks a client-streamed tarball entry by entry, rejecting anything that
+/// would land outside `dest`.
+///
+/// `async_tar::Archive::unpack` cannot be used here. Like the synchronous
+/// `tar` crate it writes link targets verbatim, so a tarball carrying
+/// `link -> /` plants an escaping symlink in the workspace — and the SFTP
```
twitchyliquid64: Is that true tho? I thought we had code in the sftp subsystem that would see if any link it was following would escape.

bryan-minimal: You're right — I checked `contained()` in the sftp subsystem and it blocks this. A write to `workspace/link/passwd` with `link -> /` canonicalizes the link to `/`, sees it's outside the export root, and returns `PathTraversal`; it handles the dangling-symlink and not-yet-created-leaf cases explicitly too. So the "write through it via SFTP" half is already defended — this isn't a live remote-write.

The fix still earns its place as defense-in-depth: it stops an escaping symlink being *planted* in the shared workspace at all (protecting the other consumers of that dir — sandbox mount, attach shell, `op::materialize` — rather than relying on each to canonicalize), and it rejects escaping tar entry paths (tar-slip). Reframing the commit + PR from "remote arbitrary write" to "harden the uploader" and dropping the SFTP claim. Good catch.

crates/minimald/src/rpc.rs
```diff
@@ -935,14 +935,87 @@ async fn unpack_workspace_files(
...
+///
+/// `async_tar::Archive::unpack` writes link targets verbatim, so `link -> /`
+/// plants an escaping symlink in the shared workspace. The sftp path
+/// canonicalizes and refuses to follow it (`SftpSession::contained`), but the
+/// other consumers of that dir (sandbox mount, attach shell, `op::materialize`)
+/// shouldn't have to — so don't plant it. Escaping entry paths (tar-slip) are
+/// rejected the same way. `unpack_workspace_patches` below already validates
+/// per-entry; this path didn't, and they share an upload channel. Containment
+/// rules come from `common::archive`, not a second copy.
+async fn unpack_validated<R>(reader: R, dest: &std::path::Path) -> Result<(), String>
```
twitchyliquid64: I think we can just drop this entirely, as its totally valid to have a symlink to some rootfs component like `/` or `/bin`, and it will just get interpreted to the `/bin` in the sandbox rootfs at runtime.

bryan-minimal: Agreed — dropped it (reverted to `async_tar::Archive::unpack`).

Traced async_tar 0.6.1 to be sure: `unpack_in` calls `self.unpack(Some(canon_target), ..)`, so hardlink targets are validated (`p.join(src)` + `validate_inside_dst`) and symlink write-through is caught by the parent-canonicalize before that — no unguarded vector. And you're right that the symlink *entry* to `/bin` is legitimate: it resolves against the sandbox rootfs at runtime, so skipping it corrupted valid uploads. Net-negative, gone.

(The shared `normalize_link_target` was just a behavior-preserving extract of `extract_tar_impl`'s own check — its only external caller was this reverted path, so I restored that inline. `extract_tar_impl` genuinely needs it: unlike `Archive::unpack` it uses the per-entry `tar::Entry::unpack`, which is the unvalidated `target_base=None` arm.)

Good catch — that's the second premise on this PR I built on without tracing the dependency first.

twitchyliquid64 (approved):

PR #1202, opened 2026-08-11
feat: make session detach key configurable (ctrl-] then d)

0chroma:
## Summary

The session detach chord is now configurable and negotiated per attach
channel, retiring the hardcoded `ctrl-w` default. `ctrl-w` is
termios-special (`VWERASE`): the kernel line discipline consumes it
before the app reads the byte, so a leaked or verbatim-forwarded leader
triggers line editing instead of reaching the app. The new default
leader is `ctrl-]` (`0x1d`), which is not termios-special and is lightly
bound by editors, tmux, and zellij.

Detach is now a two-key chord: press the leader to enter command mode,
then a subcommand. `d` detaches; a second leader verbatim-forwards a
leader byte down the PTY (a primitive for nested sessions); any unbound
key cancels command mode, mirroring tmux.

Refs gominimal/inbox#475.

## How it works

- A new `sessions::keys` module owns the key types, the reject sets
  (termios-special and wrapping-ambiguous), and the env-var contract.
- The client config gains a `[session-keys]` section. The leader is
  validated loudly at load; the daemon re-validates it as a silent
  backstop (log and fall back to default, never garble the screen).
- The daemon matcher derives the full encoding set (plain byte, kitty,
  modifyOtherKeys) from the negotiated leader, so a remapped leader gets
  its CSI forms automatically. No hardcoded key encodings remain.
- Negotiation reuses the existing per-channel env-var channel: the
  interactive attach path sends `MINIMAL_SESSION_LEADER` and friends
  alongside `MINIMAL_SESSION_ID`; exec channels send none. Two clients
  with different configs on the same session each get their own chord.

## Changes

| Area | What |
|---|---|
| `sessions::keys` (new) | `Key`, `SessionKeys`, reject sets, env-var contract, 30 unit tests |
| `sessions::client::config` | `[session-keys]` section, loud validation, `ConfigError::Validation` |
| `minimal-client::attach` | resolve session keys, send env vars with matching `SendEnv` |
| `minimal` (CLI) | `session_via_ssh` resolves keys for the interactive path; TUI threads `config_dir` |
| `minimald` | command-mode state machine, key-aware MOTD, re-validation gate, AcceptEnv doc |
| `minimal-tui` | `DashOptions.config_dir`, attach-time key resolution |
| docs, e2e | banner updated to `ctrl-] then d`, `[session-keys]` schema docs, spec F1 note |

## Breaking change

The default detach chord changes from `ctrl-w` to `ctrl-]` then `d`. The
daemon no longer intercepts `ctrl-w` (`0x17` and its CSI encodings)
anywhere. It is forwarded to the shell like any ordinary key.

## Verification

- `just fix` (fmt, clippy with `-D warnings`) is clean.
- `cargo test` on `sessions`, `minimal`, `minimald`, `minimal-client`,
  `minimal-tui` is green. 398 sessions tests, 259 minimal tests, 21
  minimal-client tests including two new attach-negotiation tests.
- Doctests pass.
- VM/daemon path: the session e2e banner assertion is updated. The full
  `just e2e` run should be confirmed on a KVM/HVF host before merge.

## Notes for reviewers

- The leader is always swallowed on match, never forwarded, so a leaked
  leader does not trigger the app `ctrl-]` binding. `ctrl-]` is not
  termios-special, so a stray leader past the deepest nesting layer is
  non-destructive.
- The daemon re-validation failure mode is load-bearing: a bad chord
  must never garble the screen. It logs and falls back to the default
  rather than rejecting the attach or writing to the channel.

> [!NOTE]
> ### Replace hardcoded `ctrl-w` detach with a configurable leader chord defaulting to `ctrl-]` then `d`
> - Introduces a new `[session-keys]` config section in `config.toml` allowing users to remap the session leader key, detach subcommand, forward-leader subcommand, and an optional bell-on-leader flag via [`crates/sessions/src/keys.rs`](https://github.com/gominimal/minimal/pull/1202/files#diff-2a693e0f0d603eb9ddea1dd57386c7d25bb3c7321f7818c885ed9135040fdfdd).
> - Adds a streaming `ChordMatcher` that scans stdin byte-by-byte across chunk boundaries, supporting plain, kitty, and modifyOtherKeys encodings, replacing the old single-byte `ctrl-w` check in the host mainloop.
> - Session keys are negotiated per-channel via SSH env vars (`MINIMAL_LEADER`, `MINIMAL_DETACH_KEY`, etc.) set by the client at attach time and validated server-side, with unsafe or ambiguous leaders silently falling back to defaults.
> - The MOTD detach hint is now driven by `MINIMAL_DETACH_HINT` (injected per-channel) with a fallback of `ctrl-] then d`, replacing the hardcoded `ctrl-w` string.
> - Behavioral Change: `ctrl-w` is no longer the detach key and is forwarded to the shell as a normal byte; the new default detach chord is `ctrl-]` then `d`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d1ae5e3.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added configurable session keys for leader, detach, forwarding, and terminal bell behavior.
  * Interactive attachments now negotiate and apply key bindings per connection.
  * Added customizable detach hints, defaulting to `Ctrl-]`, then `d`.

* **Bug Fixes**
  * Rejected unsafe or conflicting key bindings.
  * Improved handling of split, combined, stale, and renegotiated session-control input.
  * Configuration errors are reported before starting an attachment.

* **Documentation**
  * Documented session-key configuration, validation, defaults, and detach behavior.

### DIFF
--- a/crates/minimal-client/src/attach.rs
+++ b/crates/minimal-client/src/attach.rs
@@ -4,24 +4,9 @@
 //! command, replacing itself) and the `min dash` TUI (which spawns it as a
 //! child while the TUI is suspended and resumes when ssh exits).
 
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use anyhow::Context as _;
-
-/// Resolve the minimal config directory: `<config_dir>/minimal` when an
-/// override is given, else the platform default
-/// ([`paths::minimal_config_dir`]). The base under which `config.toml`,
-/// `loadouts/`, etc. live.
-pub fn minimal_config_dir(config_dir: Option<&Path>) -> PathBuf {
-    if let Some(dir) = config_dir {
-        dir.join("minimal")
-    } else {
-        paths::minimal_config_dir()
-            .as_utf8_path()
-            .as_std_path()
-            .to_path_buf()
-    }
-}
 
 /// Read and validate the session-key config, returning the resolved
 /// [`sessions::keys::SessionKeys`] to negotiate at attach. A missing config
@@ -35,7 +20,7 @@ pub fn minimal_config_dir(config_dir: Option<&Path>) -> PathBuf {
 pub fn resolve_session_keys(
     config_dir: Option<&Path>,
 ) -> Result<sessions::keys::SessionKeys, anyhow::Error> {
-    let cfg_path = minimal_config_dir(config_dir).join("config.toml");
+    let cfg_path = paths::minimal_config_dir_with_override(config_dir).join("config.toml");
     let cfg = sessions::client::config::read_config_or_default(&cfg_path)
         .map_err(|e| anyhow::anyhow!("reading session-keys config {cfg_path:?}: {e}"))?;
     cfg.session_keys
--- a/crates/minimal-client/src/attach.rs
+++ b/crates/minimal-client/src/attach.rs
@@ -4,9 +4,44 @@
 //! command, replacing itself) and the `min dash` TUI (which spawns it as a
 //! child while the TUI is suspended and resumes when ssh exits).
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anyhow::Context as _;
+
+/// Resolve the minimal config directory: `<config_dir>/minimal` when an
+/// override is given, else the platform default
+/// ([`paths::minimal_config_dir`]). The base under which `config.toml`,
+/// `loadouts/`, etc. live.
+pub fn minimal_config_dir(config_dir: Option<&Path>) -> PathBuf {
+    if let Some(dir) = config_dir {
+        dir.join("minimal")
+    } else {
+        paths::minimal_config_dir()
+            .as_utf8_path()
+            .as_std_path()
+            .to_path_buf()
+    }
+}
+
+/// Read and validate the session-key config, returning the resolved
+/// [`sessions::keys::SessionKeys`] to negotiate at attach. A missing config
+/// file yields the shipped defaults; a present-but-invalid one (e.g. a
+/// termios-special leader) surfaces its error loudly so the user fixes the
+/// config rather than silently attaching with the wrong chord.
+///
+/// Only the interactive attach path needs the keys (non-interactive exec
+/// channels have no detach); callers pass `None` to [`attach_command`] for
+/// those.
+pub fn resolve_session_keys(
+    config_dir: Option<&Path>,
+) -> Result<sessions::keys::SessionKeys, anyhow::Error> {
+    let cfg_path = minimal_config_dir(config_dir).join("config.toml");
+    let cfg = sessions::client::config::read_config_or_default(&cfg_path)
+        .map_err(|e| anyhow::anyhow!("reading session-keys config {cfg_path:?}: {e}"))?;
+    cfg.session_keys
+        .to_session_keys()
+        .map_err(|e| anyhow::anyhow!("invalid session-keys config: {e}"))
+}
 
 /// Shell-quote a string for safe interpolation into `sh -c`.
 pub fn shell_quote(s: &str) -> String {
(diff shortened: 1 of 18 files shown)


crates/minimal-client/src/attach.rs
```diff
@@ -4,10 +4,45 @@
...
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anyhow::Context as _;
 
+/// Resolve the minimal config directory: `<config_dir>/minimal` when an
+/// override is given, else the platform default
+/// ([`paths::minimal_config_dir`]). The base under which `config.toml`,
+/// `loadouts/`, etc. live.
+pub fn minimal_config_dir(config_dir: Option<&Path>) -> PathBuf {
```
twitchyliquid64: Surely we already have logic for this somewhere in the loadouts code, can we share the method?

crates/minimal-client/src/attach.rs
```diff
@@ -4,10 +4,45 @@
...
+        dir.join("minimal")
+    } else {
+        paths::minimal_config_dir()
+            .as_utf8_path()
+            .as_std_path()
+            .to_path_buf()
+    }
+}
+
+/// Read and validate the session-key config, returning the resolved
```
twitchyliquid64, line 11: If i understand correctly, for session keys we always read the current state of your loadout config, but for the rest of your session we use the state of your loadout at activate time? that behavior seems surprising?

crates/minimald/src/session_host.rs
```diff
@@ -1307,6 +1305,22 @@ pub(crate) struct Host<P: SessionProcess, G: SessionGuard> {
...
 
+    // The per-channel session-key chord matcher: the negotiated leader chord
+    // that enters command mode, the detach/forward subcommand keys, the bell
+    // flag, plus the command-mode state and the pending split-candidate
+    // buffer. Refreshed from the channel's env vars on every attach (so two
+    // clients with different configs on the same session each get their own
+    // chord, and a reattach never inherits a stale awaiting-subcommand state
+    // or half a split candidate); defaults to `ctrl-]` / `d` when a client
+    // sends no keys.
+    chord_matcher: ChordMatcher,
```
twitchyliquid64, line 1633: Newtype YAY

crates/sessions/src/keys.rs
```diff
@@ -0,0 +1,1430 @@
...
+
+/// The negotiated per-channel session-key config: the leader chord that
+/// enters command mode, the subcommand keys, and whether entering command
+/// mode rings the terminal bell.
+///
+/// Defaults match the shipped config: `leader = ctrl-]`, `detach = d`,
+/// `forward = ctrl-]` (forward defaults to the leader, so a double-press
+/// verbatim-forwards), `bell_on_leader = false`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SessionKeys {
```
twitchyliquid64, line 461: This is very cool

twitchyliquid64 (commented):
LGTM, I think the main thing to nail down is if the source-of-truth for the keys config comes from the loadout at the time of activate, or if it comes from the representation on disk at the time of attach (as currently implemented).

crates/minimal-client/src/attach.rs
0chroma: The keybinds come from `config.toml` under the `[session-keys]` section, not from loadouts. My thought process is that if we're connecting to some remote session that we didn't create, we'd still want our own predictable detach binding to be respected. So it's handled on a per-channel basis rather than a per-session basis.

crates/minimal-client/src/attach.rs
0chroma: so, we had `minimal::config::resolve_minimal_config_dir` but we couldn't access it from `minimal-client`. That's why we had to move it here but you're right, it's better to put in the `minimal-client::paths` module not here.

0chroma:
@twitchyliquid64 ready for rereview!

msample:
hoping this change can be merged soon @0chroma @twitchyliquid64

twitchyliquid64 (approved):

twitchyliquid64:
Sorry this slipped off my radar, approved

PR #1205, opened 2026-08-11
feat: lifecycle hooks

evanspearman:
Resolves https://github.com/gominimal/inbox/issues/201

## Summary

Sessions can now run scripts at their four transition points — `on_activate`,
`on_attach`, `on_detach`, `on_destroy` — declared either in a loadout or in a
project's `[session]` block. Hooks execute **inside the running session** by
joining its namespaces, so they see the same packages, variables, files, and
network the user does.

- **Declaration.** Inline bodies or external script paths, each with its own
  `timeout` (default 60s, capped at 300s — a hook is arbitrary code, so an
  unbounded one could hold a session open). Hooks concatenate in contributor
  order and run in reverse for teardown, so a project sets up before your
  loadouts and tears down after them.
- **Interpreter.** POSIX `sh` by default, or whatever a leading shebang names,
  parsed exactly as `execve(2)` would — so a hook can be fish or Python without
  behaving differently here than anywhere else.
- **Consent.** A *project's* hooks are arbitrary code from someone else, so
  they run only once the user allow-lists the project in their
  `[hooks]` policy; a loadout's are the user's own file and are ungated. A
  package cannot declare hooks at all, and the gate denies any that appear.
- **Failure semantics.** A failing `on_activate` fails the activation — a
  development environment whose setup script failed is not the environment
  that was asked for. The other three never block their transition: a session
  must always be attachable and always destroyable. Teardown additionally runs
  under one shared time budget across all its hooks.
- **Introspection.** `min session hooks <session>` lists what a session
  carries and where each hook came from, read from the persisted composition
  snapshot so it answers for stopped sessions and survives a daemon restart.
- **Opt-out.** `min session activate --no-hooks` drops them at composition
  time, so the session records that it has none rather than carrying hooks
  every later transition has to remember to skip.

Also lands the e2e coverage for the above, which found three bugs that unit
tests structurally could not: a project's external hook script could never
activate (the finalize gate demanded an upload the client does not send for
project-sourced scripts), `on_destroy` was silently skipped for any session
whose actor was idle (every session, after a daemon restart), and "Allow
permanent" stored the project path as a glob pattern, so a path containing
metacharacters either matched nothing or matched its siblings.

The session e2e's pty driver moved from Python to socat — see the dependency
note in the checklist below.

## Testing

`just e2e-native` — the full session e2e against a host-native daemon,
including the new hooks blocks and the rewritten pty driver:

```
::group::lifecycle hooks: the four transitions
on_activate + shebang dispatch OK
on_attach (on the terminal) + on_detach (headless, after leaving) OK
on_destroy OK (ran, output captured, and a non-zero exit still destroyed)
::group::lifecycle hooks: external scripts
external hook script proof OK (staged, uploaded, resolved, ran)
::group::lifecycle hooks: refusals
gate refuses an un-allow-listed project, naming it OK
a failing on_activate aborts the activation OK
--no-hooks suppresses execution and composition OK
::group::lifecycle hooks: loadout-declared
loadout-declared hooks proof OK (ran, ungated)
::group::sandbox proof (interactive attach via pty: min add jq + run)
sandbox proof: in-sandbox 'min add jq' + run OK, orientation banner rendered (3041ms)
::group::lifecycle hooks: survive a daemon restart
hooks survive a daemon restart OK (listed, and still executable)
session e2e OK
```

Unit and integration tests across the touched crates:

```
$ cargo test -p minimald -p sessions -p minimal -p mfile
total passed: 980   (0 failed)
```

Static gates:

```
$ just clippy       # clean
$ just fmt-check    # clean
$ just lint-shell   # 27 script(s) passed shellcheck
```

Each regression test was confirmed to fail against the pre-fix behaviour
before being kept — e.g. reverting the destroy fix yields *"destroying an idle
session skipped its on_destroy hook"*.

**Not run:** `just e2e` (the VM lane). This box has no `/dev/kvm`, so the
VM-backed path is unverified locally; the e2e is lane-agnostic and the native
lane exercises the identical script.

## Checklist

  `docs/reference/loadouts.md` (hook semantics, interpreter/shebang rules,
  metadata variables, teardown budget), `docs/reference/minimal-dot-toml.md`,
  `docs/reference/user-policy.md`, `docs/reference/cli-min.md`,
  `docs/concepts/loadouts.md`, and `AGENTS.md`.

- [x] `BREAKING CHANGE:` footer present if this is a breaking change

  Not breaking. `on_failure` — an earlier spelling that never executed — is
  rejected by name with an error pointing at `on_destroy`, rather than being
  silently ignored.

### Reviewer notes

- ~~**New host dependency: `socat`.** The e2e drives its interactive attaches
  through a real pty, because only a tty can answer the session-exit prompt.
  That driver was Python (an undeclared assumption that happens to hold on
  hosted runners); it is now `scripts/e2e-attach-pty.sh`. socat is present on
  stock Linux but **not** stock macOS — the self-hosted Apple Silicon runner
  needs `brew install socat` before the macOS lane will pass. Documented in
  AGENTS.md; I could not add an install step because `.github/workflows/` is
  frozen.~~
- `crates/sessions/src/core/policy.rs` gains `HooksPolicy` as a near-copy of
  `PatchesPolicy`. Left deliberately un-unified — the two are incidentally the
  same shape today and are expected to diverge.

> [!NOTE]
> ### Add lifecycle hooks (on_activate, on_destroy, on_attach, on_detach) to sessions
> - Introduces four lifecycle hook transitions that run scripts inside the session at activation, destruction, attach, and detach; hooks are declared in loadout or project `minimal.toml` files and composed alongside vars and patches.
> - Adds a user-level `[hooks]` policy (allow/deny/ignore) that gates project-declared hooks; interactive mode prompts for approval and persists decisions to `user_policy.toml`; packages cannot declare hooks.
> - External hook scripts (file-backed, not inline) are validated and staged locally by the client then uploaded to the daemon via a new `WorkspaceHookScriptsTarZst` SSH subsystem before finalization.
> - `min activate --no-hooks` disables all hooks for a session; the choice persists across attach, detach, and destroy transitions.
> - Adds `min session hooks` CLI subcommand to list composed hooks for a running or idle session, with `--json` output support.
> - On-activate hook failures abort activation with a detailed error; detach/destroy hooks run best-effort under a shared 120s teardown budget and are skipped (with a warning) if a host cannot be started within 60s.
> - Risk: sessions with external hook scripts now require a hook-script upload step before `FinalizeSession` will succeed; missing the upload returns an `InvalidInput` error instructing the client to upload scripts first.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfd3e6d.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle hooks for activation, attachment, detachment, and destruction, including ordering, timeouts, output handling, and failure reporting.
  * Added `session hooks` to view configured hooks in table or JSON format.
  * Added policy-based hook approval, denial, and ignore controls.
  * Added `session activate --no-hooks`, persisted across later transitions.
  * Added safe staging and execution of external hook scripts.

* **Documentation**
  * Updated lifecycle hook configuration, policy, timeout, security, and migration guidance.
  * Added example configurations and end-to-end coverage.

### DIFF
--- a/crates/minimal/src/loadouts.rs
+++ b/crates/minimal/src/loadouts.rs
@@ -196,6 +196,134 @@ pub(crate) fn compose_options_from_config(
         .with_follow_symlinks(cfg.loadouts.follow_symlinks)
 }
 
+/// Resolve and validate every external hook script the active loadouts
+/// declare, ready to upload.
+///
+/// Run before the daemon connection is opened so a mistyped script path
+/// fails on this machine, naming the file, instead of surfacing after a
+/// session already exists on the daemon.
+///
+/// Only *loadout* scripts are staged here. A project's external scripts
+/// live in the project tree, which the daemon already has from the
+/// workspace upload, so it resolves those itself —
+/// [`check_project_hooks`] validates them without staging them.
+///
+/// Returns an empty list when hooks are disabled — `--no-hooks` means
+/// nothing is staged, uploaded, or run.
+pub(crate) fn stage_loadout_hook_scripts(
+    active: &ActiveLoadouts,
+    global: &GlobalArgs,
+    project_root: &camino::Utf8Path,
+    hooks_enabled: bool,
+) -> Result<Vec<sessions::client::hookscripts::StagedScript>, anyhow::Error> {
+    use sessions::client::hookscripts::{ScriptAnchors, stage_external_scripts};
+    use sessions::core::source::{ProvenancedHook, Source};
+
+    if !hooks_enabled {
+        return Ok(Vec::new());
+    }
+    let loadouts_dir =
+        camino::Utf8PathBuf::from_path_buf(resolve_minimal_config_dir(global).join("loadouts"))
+            .map_err(|p| {
+                anyhow::anyhow!("loadouts directory is not valid UTF-8: {}", p.display())
+            })?;
+
+    let hooks: Vec<ProvenancedHook> = active
+        .loadouts
+        .iter()
+        .flat_map(|l| {
+            let source = Source::UserLoadout {
+                name: l.name().as_ref().to_owned(),
+            };
+            l.lifecycle_hooks()
+                .iter()
+                .map(move |h| ProvenancedHook::new(h.clone(), source.clone()))
+        })
+        .collect();
+
+    let anchors = ScriptAnchors {
+        loadouts_dir,
+        project_root: project_root.to_owned(),
+    };
+    stage_external_scripts(&hooks, &anchors).map_err(|e| anyhow::anyhow!("{e}"))
+}
+
+/// Validate the hooks a project declares, before a session exists.
+///
+/// The daemon composes these from the uploaded mfile, so nothing here is
+/// staged or sent — but every reason a hook would fail later is knowable
+/// now, on the machine that owns the file:
+///
+/// - an **external** script that is missing, is not a regular file, or
+///   traverses a symlink. The daemon would surface this only when the
+///   hook ran, which for `on_destroy` is at teardown, long after the
+///   mistake.
+/// - an **inline** script that is empty, which can only ever be a typo:
+///   it declares a transition and then does nothing at it.
+///
+/// A project with no mfile, or one that does not parse, is not this
+/// function's problem — the activation has its own handling for both, and
+/// duplicating it here would report the same fault twice.
+pub(crate) fn check_project_hooks(project_root: &camino::Utf8Path) -> Result<(), anyhow::Error> {
+    use sessions::client::hookscripts::{ScriptAnchors, stage_external_scripts};
+    use sessions::core::lifecyclehook::{HookScript, HookScriptBody};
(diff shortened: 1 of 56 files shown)


crates/minimal/src/loadouts.rs
```diff
@@ -213,10 +213,139 @@ pub(crate) fn compose_options_from_config(
 /// allow/ignore/deny rules and the caller wants to persist them.
 ///
 /// [`UserPolicy`]: sessions::core::policy::UserPolicy
```
twitchyliquid64: I think this docstring meant to stay with `compose_user_contribution`

twitchyliquid64 (approved):

crates/minimal/src/loadouts.rs
evanspearman: Fixed

Issue #1109, opened 2026-07-30
Introduce Kani bounded verification over the pure security-load-bearing cores


bryan-minimal:
## Summary

Introduce [Kani](https://model-checking.github.io/kani/) (a bit-precise bounded model checker for Rust) as a CI-grade verification tool over a handful of minimal's pure, security-load-bearing cores. Proof harnesses look like fuzz targets, run in CI, and prove properties like "never panics on hostile input", "never reads out of bounds", and small algebraic laws — over **all** inputs within a bound, not just sampled ones.

This is deliberately scoped as **additive and non-breaking**: no `BuildSpec` changes, no cache-key changes, no format changes. Harnesses live behind `#[cfg(kani)]`, land independently, and need no maintainer decisions to start. It's the pragmatic floor of the verification story from the formal-methods audit — value today, and a concrete stepping stone toward the deeper Aeneas/Lean question without committing to that toolchain.

## Why Kani, why here

The audit identified a set of pure "kernel" cores that parse or combine **untrusted input** — bytes from a bucket/CDN/mirror our own trust model explicitly treats as hostile, or path decisions where a reorder could be an escape. These are the textbook Kani sweet spot: small, pure, owned data in, bounded input space. AWS runs Kani in CI on Firecracker, s2n-quic, and the Rust std-lib verification campaign (16k+ harnesses/change), so the CI pattern is well-trodden.

Two of the targets below already ship *defenses* (a varint-overflow guard; a `MAX_VARINT_LEN` DoS cap). Kani's job there is to **prove the defense holds for every input in-bound**, turning "we added a check" into "the check is provably sufficient" — and to lock it against regression.

### 1. `rcache::index_file` — parsing untrusted cache-index bytes *(do first)*
`IndexFile::from_reader` (`crates/rcache/src/index_file.rs:104`) parses fixed 68-byte records (`spec_hash(32) | flags(4) | sha256(32)`, `WIRE_RECORD_LEN` at :29) from storage the signed-index design (#86) exists *because* we don't trust. This same parser feeds build-servers' seal path, so one harness guards the hot format in both repos.
- **Harnesses:** arbitrary `&[u8]` → `from_reader` returns `Ok`/`Err`, never panics; `merge` (:146) on two small bounded indices returns the correct conflict count and preserves canonical (`BTreeMap`) order; round-trip `from_reader(write_to(x)) == x` on bounded indices.
[...]

PR #1217, opened 2026-08-13
test(kani): bounded verification for the index_file codec and PathDecision lattice

bryan-minimal:
First Kani slice per **#1109**'s PR-1 scope: the two anchor harness sets plus the tooling. Additive and non-breaking — everything lives behind `#[cfg(kani)]`, no format/API changes, and the CI lane is **advisory** until promoted.

## What is now *proven* (not tested — proven, over all inputs in the stated bounds)

**`sessions` — the `PathDecision` lattice, exhaustively.** Six proofs with no bounds and no assumptions (4-variant `Copy` enum → Kani enumerates the whole space):

- `combine` **is** max-by-severity under the documented precedence `Denied > Ignored > NeedsApproval > Allowed` — the one lemma that entails the rest, kept alongside the named laws so a regression names what it broke
- `Denied`-absorbing, `Allowed`-identity, commutative, associative, idempotent

Together these discharge the reorder-attack defense on `ExpandedPatchPolicy::check`: **no permutation or regrouping of per-path decisions can downgrade a deny.** Note `severity()` in the harness is deliberately *not* the enum's discriminant order — declaration order disagrees with precedence (`NeedsApproval` is numerically greatest but not most restrictive), which is exactly the confusion these proofs pin against. This is also the Aeneas/Lean warm-up target the issue names, previewed at zero toolchain commitment.

**`rcache` — the 68-byte record codec, the layer that eats hostile bucket/CDN bytes:**

- decoding an arbitrary record **at any truncation** never panics, and success consumes exactly 68 bytes
- the `flags` forward-compat gate **always** fires on any nonzero flags — and fires before the sha256 bytes are consumed
- encode→decode round-trips every record exactly, emitting exactly 68 bytes with zero flags — output this code writes can never trip its own gate

## The one scoping decision to review

The issue sketched file-level harnesses (`from_reader`/`merge`). The first cut did that and **OOM'd CBMC**: symbolic 32-byte keys inside `BTreeMap` are a classic bounded-model-checking blow-up, and it wouldn't fit CI runners either. This PR deliberately proves the **record codec totally** and leaves file-level composition (`from_reader`'s error-latch, `merge`, canonical ordering) to the existing fuzz target and unit tests — std's `BTreeMap` is not our proof obligation. The division of labor is stated in the harness module docs. If we later want file-level proofs, the path is stubbing the map for a small model, as its own PR.

## Tooling (`scripts/kani.sh`, `just kani`, `ci-kani.yml`)

- **Kani pinned exactly 0.67.0** — not hygiene, correctness: older releases give *spurious verification failures on arrays >64 elements* (kani#2416/#4408, fixed by the CBMC 6.8.0 upgrade), and one wire record is 68 bytes.
- **The MSRV wall**: Kani 0.67.0 bundles a 1.93-nightly toolchain, numerically below our declared `package.rust-version` floor. The gate is declarative only — the nightly compiles the tree and all 9 proofs verify — but cargo hard-errors and `cargo-kani` has no `--ignore-rust-version`. `scripts/kani.sh` therefore runs from a scratch rsync copy with the floor relaxed, loudly documented, to be deleted when Kani ships a ≥floor toolchain.
- **Sequential on purpose** — `-j` OOM'd CBMC; the whole suite solves in ~7s sequentially.
- **`ci-kani.yml` is its own commit** for code-owner review per the frozen-workflows policy (docs/ci-strategy.md §10). House pattern: always-triggered, `changes`-filtered in-workflow (no trigger-level paths), `ci-kani-success` aggregator. **Advisory** — becomes required only when someone adds `ci-kani-success` to the ruleset; no workflow change needed at that point.
- Both proved crates declare `check-cfg` for `cfg(kani)` so the stable clippy `-D warnings` gate stays clean.

## Verification

- `./scripts/kani.sh` end-to-end: **9/9 harnesses verified** (sessions 6 exhaustive in 0.02s; rcache 3 in ~7s), exit 0 — the exact invocation the CI lane runs
- `cargo check`/`clippy`/`fmt` clean on both crates with harnesses cfg'd out (stable)
- One spurious-counterexample lesson learned and recorded in-harness: blake3's constant-time `PartialEq` is opaque to the model checker — the round-trip proof compares key *bytes*, which is the actual property anyway

## Follow-ups (per the issue's sequencing)

Harness sets #2 (varint), #3 (`fixup_spec`), #5 (`normalize_within_root`) as separate PRs; #6 rides the hash-epoch work. The `unwind`-bound playbook and the scratch-copy MSRV dance from this PR carry over directly.

> [!NOTE]
> ### Add Kani bounded-verification proofs for `index_file` codec and `PathDecision` lattice
> - Adds 3 Kani proof harnesses in [`crates/rcache/src/index_file.rs`](https://github.com/gominimal/minimal/pull/1217/files#diff-b50196204a4e8f2dfaca3adb9746243b539d9066d3ceade1dd995efb2185c92d) verifying that `read_wire_kv`/`write_wire_kv` never panic, reject nonzero flags, and round-trip byte-accurately.
> - Adds 6 Kani proof harnesses in [`crates/sessions/src/core/policy.rs`](https://github.com/gominimal/minimal/pull/1217/files#diff-d86bf7ae6641c5f84d1217ae9c6938d264308ac2bf9e79b21ef3a23a92d67de1) verifying that `PathDecision::combine` is commutative, associative, idempotent, has `Allowed` as identity, and has `Denied` as absorbing element.
> - Adds a GitHub Actions workflow ([`ci-kani.yml`]( file:.github/workflows/ci-kani.yml)) and [`scripts/kani.sh`](https://github.com/gominimal/minimal/pull/1217/files#diff-bf932986711b447aef632366a6f045bb0e27b8e067e32dedde213b79e30b293a) that run proofs in CI, pinning Kani 0.67.0 and asserting exact harness-verified counts (6 for `sessions`, 3 for `rcache`).
> - The script copies the workspace to a scratch directory and relaxes `rust-version` to 1.90 to satisfy Kani's toolchain requirements.
> - The `ci-kani-success` lane is advisory (not a required check) until promoted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d10c8d1.</sup>
> 
>

## Summary by CodeRabbit

* **Tests**
  * Added bounded verification for session policy evaluation and cache index encoding.
  * Added checks for malformed input handling, round-trip consistency, and policy-combination properties.

* **Chores**
  * Added automated verification on code changes and manual runs.
  * Added a local command for running verification.
  * Documented required verification tools and compatibility constraints.
  * Improved confidence in policy decisions and cache data integrity.

### DIFF
--- a/.github/workflows/ci-kani.yml
+++ b/.github/workflows/ci-kani.yml
@@ -1,0 +1,102 @@
+name: CI (Kani proofs)
+
+# Kani bounded-verification harnesses over the pure security-load-bearing
+# cores (#1109): the rcache index_file untrusted-bytes parse path and the
+# sessions PathDecision combination lattice. ADVISORY lane: not in branch
+# protection's required contexts until the harnesses have soaked — flip by
+# adding `ci-kani-success` to the ruleset, no workflow change needed.
+#
+# ALWAYS-TRIGGERED, path-scoped INSIDE the workflow (same pattern as every
+# other lane — do NOT add trigger-level `paths:` here; a required context
+# skipped at trigger level wedges PRs as "Expected" forever).
+on:
+    push:
+        branches: ["main"]
+    pull_request:
+        branches: ["main"]
+    workflow_dispatch:
+
+env:
+    CARGO_TERM_COLOR: always
+    # Pin EXACTLY: 0.66.0 and earlier give spurious verification failures
+    # on arrays with more than 64 elements (kani#2416/#4408, fixed by the
+    # CBMC 6.8.0 upgrade in 0.67.0) — one index_file wire record is 68
+    # bytes, so an older cached Kani would fail proofs that are correct.
+    KANI_VERSION: 0.67.0
+
+concurrency:
+    group: ci-kani-${{ github.ref }}
+    cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+permissions:
+    contents: read
+    pull-requests: read
+
+jobs:
+    changes:
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        outputs:
+            kani: ${{ steps.filter.outputs.kani }}
+        steps:
+            - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+              with:
+                  persist-credentials: false
+            - name: Detect changes to proved crates or this lane
+              id: filter
+              if: github.event_name == 'pull_request'
+              uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v3.0.2
+              with:
+                  filters: |
+                      kani:
+                        - "crates/rcache/**"
+                        - "crates/sessions/**"
+                        - "crates/common/src/spec_hash.rs"
+                        - ".github/workflows/ci-kani.yml"
+                        - "justfile"
+                        - "Cargo.toml"
+                        - "Cargo.lock"
+
+    kani:
+        needs: [changes]
+        if: github.event_name != 'pull_request' || needs.changes.outputs.kani == 'true'
+        runs-on: ubuntu-latest
+        timeout-minutes: 60
+        steps:
+            - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+              with:
+                  persist-credentials: false
+            - uses: ./.github/actions/setup-rust
+              with:
+                  shared-key: kani
+            - name: Cache Kani install
+              id: kani-cache
+              uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+              with:
+                  path: |
+                      ~/.cargo/bin/cargo-kani
+                      ~/.cargo/bin/kani
+                      ~/.kani
+                  key: kani-${{ env.KANI_VERSION }}-${{ runner.os }}
+            - name: Install Kani
+              if: steps.kani-cache.outputs.cache-hit != 'true'
+              run: |
+                  cargo install --locked kani-verifier --version "$KANI_VERSION"
+                  cargo kani setup
+            - name: Run proofs
+              run: ./scripts/kani.sh
+
+    # Lane aggregator, same shape as every other lane: branch protection
+    # keys on this job (once the lane is promoted from advisory), so
+    # adding/removing jobs above stays a workflow-only change. Skipped
+    # counts as pass; failure or cancellation of any needed job fails it.
+    ci-kani-success:
+        if: always()
+        needs: [changes, kani]
+        runs-on: ubuntu-latest
+        timeout-minutes: 10
+        steps:
+            - name: Verify all gating jobs succeeded or were skipped
+              if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+              run: exit 1
+            - run: echo "Kani lane passed"
--- a/.github/workflows/ci-kani.yml
+++ b/.github/workflows/ci-kani.yml
@@ -51,11 +51,16 @@ jobs:
                       kani:
                         - "crates/rcache/**"
                         - "crates/sessions/**"
+                        # In the current proof closure: the rcache record
+                        # harnesses construct SpecHash::from_bytes and
+                        # serialize via as_bytes. Direct spec-hash codec
+                        # proofs are #1109 set #6 (hash-epoch work).
                         - "crates/common/src/spec_hash.rs"
                         - ".github/workflows/ci-kani.yml"
                         - "justfile"
                         - "Cargo.toml"
                         - "Cargo.lock"
+                        - "scripts/kani.sh"
 
     kani:
         needs: [changes]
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-kani.yml
```diff
@@ -0,0 +1,102 @@
...
+              with:
+                  filters: |
+                      kani:
+                        - "crates/rcache/**"
+                        - "crates/sessions/**"
+                        - "crates/common/src/spec_hash.rs"
+                        - ".github/workflows/ci-kani.yml"
+                        - "justfile"
+                        - "Cargo.toml"
+                        - "Cargo.lock"
```
norrietaylor, line 62: `scripts/kani.sh` is missing from this filter. A PR that edits only the proof runner — the file holding the harness-count gate that keeps this lane from going green vacuously — skips the lane entirely on the pull_request path.

```suggestion
                        - "Cargo.lock"
                        - "scripts/kani.sh"
```

.github/workflows/ci-kani.yml
```diff
@@ -0,0 +1,102 @@
...
+            - name: Detect changes to proved crates or this lane
+              id: filter
+              if: github.event_name == 'pull_request'
+              uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v3.0.2
+              with:
+                  filters: |
+                      kani:
+                        - "crates/rcache/**"
+                        - "crates/sessions/**"
+                        - "crates/common/src/spec_hash.rs"
```
norrietaylor, line 58: `crates/common/src/spec_hash.rs` has no harness behind it yet — `scripts/kani.sh` proves only `sessions` and `rcache`; the spec-hash codec proofs are [#1109](https://github.com/gominimal/minimal/issues/1109) set #6, which rides the hash-epoch work. Failure direction is harmless (extra runs, not false greens), but either drop this line until set #6 lands or add a comment marking it forward-looking, so the filter doesn't read as claiming coverage that doesn't exist.

norrietaylor (commented):
Codeowner pass over the frozen-workflow commit. Lane shape is right (always-triggered, in-workflow `changes` filter, `if: always()` aggregator, logic in `scripts/kani.sh` per docs/ci-strategy.md §10), and the harness-count gate in `kani.sh` closes the vacuous-green hole. Two findings on the `changes` filter, inline — the first needs a fix before this merges.

.github/workflows/ci-kani.yml
bryan-minimal: Kept, with a comment added in b12656e4 explaining why: the line is in the *current* proof closure, not only forward-looking — `read_wire_kv` constructs `SpecHash::from_bytes` and `write_wire_kv` serializes via `as_bytes`, so a spec_hash.rs change can alter what the rcache record harnesses prove. The direct codec proofs are indeed set #6, and the comment says so. Happy to drop it instead if you'd rather the filter track harness files strictly.

crates/rcache/src/index_file.rs
```diff
@@ -251,3 +251,97 @@ mod tests {
...
+
+    /// The flags forward-compat gate always fires: any record whose 4
+    /// flag bytes are not all zero fails to decode.
+    #[kani::proof]
+    fn nonzero_flags_always_rejected() {
+        let bytes: [u8; RECORD] = kani::any();
+        kani::assume(bytes[32] != 0 || bytes[33] != 0 || bytes[34] != 0 || bytes[35] != 0);
+        let mut cur = Cursor::new(&bytes[..]);
+        assert!(read_wire_kv(&mut cur).is_err());
+    }
```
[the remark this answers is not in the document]
bryan-minimal: Already present — `assert_eq!(cur.position(), 36)` landed with the mutation-testing hardening commit (index_file.rs:329 at HEAD); this finding reviewed the pre-hardening hunk.

.github/workflows/ci-kani.yml
bryan-minimal: Fixed in b12656e4 — added `scripts/kani.sh` to the filter. Good catch: the file carrying the vacuity gate was itself outside the gate's trigger.

norrietaylor: Fixed in b12656e4 — `scripts/kani.sh` is in the filter now.

.github/workflows/ci-kani.yml
norrietaylor: Resolved in b12656e4 — keeping the entry, and the annotation I asked for turned out to be wrong in one respect: the rcache record harnesses construct `SpecHash::from_bytes` and serialize via `as_bytes`, so this file is in the *current* proof closure, not merely forward-looking. The comment now states that; direct spec-hash codec proofs remain [#1109](https://github.com/gominimal/minimal/issues/1109) set 6.

norrietaylor (approved):
Codeowner approval for the frozen-workflow addition (`ci-kani.yml`): house lane pattern throughout (always-triggered, in-workflow `changes` filter, `if: always()` aggregator, logic in `scripts/kani.sh`), harness-count gate closes the vacuous-green hole, and both filter findings are resolved in b12656e4. Lane stays advisory until `ci-kani-success` is added to the ruleset. The branch also adds the latest-Kani version canary to `nightly-tests` (e1e7e3cf), non-blocking, so the 0.67.0 pin cannot go stale silently.

PR #1225, opened 2026-08-14
fix(dash): match the Preview pane to the session PTY size

### LINKED ISSUE #1197: dash: full-screen TUIs (e.g. btop) break rendering and trap the detach keybind
Reported from dogfooding `min dash`.

When a session is running a full-screen / alternate-screen TUI such as `btop`, the dash **preview pane** renders it with broken wrapping — lines wrap in weird ways instead of displaying cleanly. (May be an easy fix.)

Likely a size mismatch: the session's PTY dimensions (and what `btop` has drawn for them) don't match the preview pane area dash renders the `ScreenSnapshot` into, so rows reflow/crop incorrectly.

## Detach keybind — tracked elsewhere
Attaching to a session with `btop` running also swallows the detach control keypress (the inner TUI captures it; killing `btop` first lets detach work). **That portion is tracked in https://github.com/gominimal/inbox/issues/475** and is out of scope here — this issue is only about the preview rendering.

_From dash feedback (Slack): https://minimal-dev.slack.com/archives/C091HGA43NC/p1786406103071889?thread_ts=1786387349.174509&cid=C091HGA43NC_

0chroma:
## Summary

The `min dash` Preview pane shows a `GetSessionScreen` snapshot of the session's PTY, sized for the last attach terminal. In a narrower dash terminal, a full-screen app (btop and similar) renders clipped on the right and looks broken.

This change makes the TUI resize the session PTY to the pane size:

- New oneshot RPC `SetSessionScreenSize` in `minimald-rpc`, carried through the daemon's actor chain (manager, session, host). `Host::set_size` is already a no-op when the size is unchanged, so steady state costs nothing.
- The TUI stashes the Preview pane's `(rows, cols)` at render time. On each tick, if the last snapshot's size disagrees with the pane, it issues the resize before fetching the next snapshot.
- The resize is best-effort. A daemon that predates this change answers "request subsystem" and the preview keeps its clipped behavior, so the wire change stays back-compat in both directions.

## Changes

| File | What it does |
|---|---|
| `crates/minimald-rpc/src/lib.rs` | New `SetSessionScreenSize` RPC type and request struct |
| `crates/minimald/src/session_host.rs` | `Message::SetScreenSize` arm; `HostHandle::set_screen_size` |
| `crates/minimald/src/session.rs` | `SetHostScreenSize` arm; `SessionHandle::set_screen_size` |
| `crates/minimald/src/sessions.rs` | Manager message, handler, public `set_screen_size` |
| `crates/minimald/src/rpc.rs` | Dispatch registration and `serve_set_session_screen_size` |
| `crates/minimal-tui/src/rpc.rs` | `set_screen_size` RPC client |
| `crates/minimal-tui/src/app.rs` | `preview_pane` on the model, `Effect::SetScreenSize`, tick mismatch detection, `exec_effect` arm |
| `crates/minimal-tui/src/render.rs` | Stash the pane size during `render_detail` |

## Verification

- `just ci` green: fmt, clippy, cargo-deny, 1728 tests, doctests. The one `test-ignored` failure (`mctx tests::task_env`) also fails on clean `main` and is unrelated.
- New daemon integration test: `set_session_screen_size_resizes_the_live_terminal` (resize reaches the screen snapshot; hostless session is a no-op).
- New TUI unit test: `tick_resizes_the_pty_when_the_pane_and_snapshot_disagree` (resize only on mismatch).

Fixes #1197

## Summary by CodeRabbit

* **New Features**
  * Terminal sessions now automatically resize to match the Preview pane.
  * Preview dimensions update as the interface layout changes.
  * Resizing works for active sessions and safely handles unavailable sessions or hosts.

* **Bug Fixes**
  * Prevented mismatches between the displayed terminal preview and the session’s actual terminal size.
  * Resize failures are handled gracefully without interrupting normal screen updates.

### DIFF
--- a/crates/minimal-tui/src/app.rs
+++ b/crates/minimal-tui/src/app.rs
@@ -138,6 +138,10 @@ pub enum Effect {
     Refresh,
     FetchDetail(SessionKey),
     FetchScreen(SessionKey),
+    /// Resize the session's terminal to the Preview pane's `(rows, cols)` so
+    /// the next snapshot matches the pane. Best-effort; a failure (old daemon)
+    /// is ignored and the preview keeps its clipped behavior.
+    SetScreenSize(SessionKey, u16, u16),
     Destroy(SessionKey),
     Rename(SessionKey, String),
     Create {
@@ -173,6 +177,12 @@ pub struct Model {
     pub filter: FilterState,
     pub details: HashMap<SessionKey, Detail>,
     pub screens: HashMap<SessionKey, ScreenFetch>,
+    /// The Preview pane's `(rows, cols)` from the last render, stashed by
+    /// `render_detail`. Compared against the focused session's last snapshot
+    /// size on each tick; a mismatch triggers a best-effort
+    /// `SetSessionScreenSize` so the daemon's PTY (and thus the next
+    /// snapshot) matches the pane. `None` when no session is focused.
+    pub preview_pane: Option<(u16, u16)>,
     /// The most recent bell timestamp acknowledged for a session; a bell
     /// newer than this lights the `●` indicator.
     pub bells_seen: HashMap<SessionKey, DateTime<Utc>>,
@@ -211,6 +221,7 @@ impl Model {
             filter: FilterState::default(),
             details: HashMap::new(),
             screens: HashMap::new(),
+            preview_pane: None,
             bells_seen: HashMap::new(),
             action: None,
             status: None,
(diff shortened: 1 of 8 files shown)


0chroma:
Closing this unmerged. Two findings changed the picture:

1. The Preview garbling this targeted was root-caused to a parser bug, not pane geometry: `vt100-ctt` (the session's terminal model in `minimald`) doesn't dispatch CSI `f` (HVP), and btop positions every stroke with `ESC[r;cf` — so its absolute positioning is silently dropped and output concatenates inline. Full analysis on #1197.

2. Regardless, resizing the session PTY to fit the pane mutates shared terminal state — any other attached client would have its terminal resized out from under it. The preview should stay read-only; plain truncation of the snapshot is the accepted trade.

PR #1234, opened 2026-08-17
feat(mctx): default cache reads to the R2 mirror (cache.minimal.dev)

bryan-minimal:
**DRAFT until [infra#422](https://github.com/gominimal/infra/pull/422) is deployed and verified** — the cache.minimal.dev custom domain + index.shisha bypass route must be live before this default can ship. I'll mark ready-for-review with the verification transcript once the domain passes the three-curl parity check (artifact bytes ≡ .farm ≡ GCS, fresh index, 404 passthrough).

## What flips

Read-side default only. Precedence after this change: `MINIMAL_REMOTE_CACHE_URL` env override → explicitly configured location → **`https://cache.minimal.dev/`**. Ends the per-repo whack-a-mole (webapp#380, pkgs#613, build-servers#239 — all become harmless no-ops) and covers every laptop and future consumer with zero wiring.

## What deliberately does NOT change

- **Writes**: still derive from the configured `gs://` bucket; an HTTPS mirror remains read-only. The read default cannot leak into the writer (new test pins this).
- **In-cloud readers**: res-servers pin `gs://` explicitly (buildbot sends `remote_cache_gcs_bucket` unconditionally — verified at the single call site), keeping free same-region GCS reads.
- **Escape hatch**: `MINIMAL_REMOTE_CACHE_URL=gs://minimal-staging-cache` returns any client to direct GCS.

## Also in here

The chosen read location + source is now logged at `info` — selection was previously silent, which made "is this client actually on the mirror?" unanswerable from logs (this month's egress attribution took GCS access-log forensics; that gap dies here).

## Safety case

Same objects, byte-identical (R2 is Sippy-fed from authoritative GCS; verified); every artifact is sha256-verified client-side regardless of source; the mutable root `index.shisha` is served fresh from GCS by the edge Worker on both hostnames; per-commit snapshots (the default read path since #912/#995) are immutable. The env-var version of this exact routing has run in webapp CI since July 8 and is canaried by dogfood (#667).

Local validation: `cargo check -p mctx` clean; clippy findings identical to baseline (10 pre-existing); mctx's test target doesn't compile on macOS at baseline either (pre-existing, unrelated) — Linux CI is authoritative for the new test.

> [!NOTE]
> ### Default remote cache reads to the R2 mirror (cache.minimal.dev) in `mctx`
> - `ConfigBuilder.build` now selects the read URL by precedence: env override (`MINIMAL_REMOTE_CACHE_URL`) > explicitly configured value > new `DEFAULT_REMOTE_CACHE_READ_URL` constant pointing at `https://cache.minimal.dev/`.
> - Write bucket derivation is unchanged — it continues to come from the explicitly configured location via `parse_remote_cache_url`.
> - An `info` log is emitted at build time indicating the selected read URL and which precedence level chose it.
> - Behavioral Change: unconfigured builds now read from the R2 mirror instead of the GCS default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d502ef2.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Improved remote cache configuration by using a faster default read source.
  * Preserved existing cache write behavior while separating read and write locations.
  * Added clearer logging to indicate the selected cache source.

### DIFF
--- a/crates/mctx/src/config.rs
+++ b/crates/mctx/src/config.rs
@@ -42,6 +42,18 @@ impl std::error::Error for ConfigError {
 /// bucket. A bare name (no scheme) is treated as a GCS bucket.
 pub const DEFAULT_REMOTE_CACHE_BUCKET: &str = "minimal-staging-cache";
 
+/// Default *read* location: the R2 mirror on the product domain. Cache
+/// reads are immutable, content-addressed, and sha256-verified
+/// client-side, and R2 charges no egress — where the same reads from
+/// the GCS bucket are billed internet egress for every consumer
+/// outside GCP. Writers never use this (writes require the `gs://`
+/// bucket above). An explicitly configured location or the
+/// `MINIMAL_REMOTE_CACHE_URL` env var both take precedence: in-cloud
+/// readers (res-servers) pin `gs://` explicitly for free same-region
+/// reads, and pointing the env var at `gs://minimal-staging-cache` is
+/// the escape hatch back to direct GCS.
+pub const DEFAULT_REMOTE_CACHE_READ_URL: &str = "https://cache.minimal.dev/";
+
 /// Parses a remote-cache location string into (a) the resolved [`AnyUrl`] a
 /// reader fetches from and (b) the GCS bucket name a *writer* would use
 /// (`None` for an HTTPS mirror — you can't write to a read mirror). Accepted:
@@ -180,6 +192,7 @@ impl ConfigBuilder {
     /// Constructs a config object using the given builder.
     pub fn build(self) -> Result<Config, ConfigError> {
         // The one configured cache location (default: the shared GCS bucket).
+        let explicitly_configured = self.remote_cache_url.is_some();
         let remote_cache_url = self
             .remote_cache_url
             .unwrap_or_else(|| DEFAULT_REMOTE_CACHE_BUCKET.to_string());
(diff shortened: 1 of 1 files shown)


bryan-minimal:
## cache.minimal.dev is live — verification transcript (2026-08-17)

Deployed via infra#422 (custom domain) + infra#423 (zone-plane routes; also opts the cache host out of the `*minimal.dev/*` analytics-Worker wildcard). Cloudflare stack re-preview settles clean (68 unchanged).

**1. Artifact byte-parity** — same object from all three hosts, sha256 `f3099f7e66d2cb4b53b3a2ad…` (which is also its object name — content addressing self-verifies):
- `cache.minimal.dev` ✓  · `cache.minimal.farm` ✓ · authoritative GCS ✓

**2. Mutable root index served fresh** — `GET cache.minimal.dev/index.shisha` → HTTP 200, `cache-control: no-store` (the bypass Worker answering, not R2/Sippy), byte-identical to the live GCS object (`506c3a0564d0…`).

**3. Semantics intact** — missing object → clean 404 (the reader's "empty index" signal, unrewritten); per-commit closure snapshot serves correctly (62,424 bytes).

Ready for review — un-drafting is @bryan-minimal's call per the plan. Smoke-test the flip today without waiting for the channel build:
```
MINIMAL_REMOTE_CACHE_URL=https://cache.minimal.dev/ min session activate --name r2test --sync tarball
```

twitchyliquid64 (approved):

PR #1237, opened 2026-08-17
feat(decode): implement nickel/parse logic for container specs

twitchyliquid64:
- New types to support field validation for containers in the embedded nickel stdlib
 - Code in `decode` to parse containers specs into a `decode::Container` struct
 - Tests for parsing

## Summary by CodeRabbit

* **New Features**
  * Added support for defining and decoding container specifications in layers.
  * Added container image metadata, commands, environment variables, ports, volumes, labels, users, signals, and architecture settings.
  * Added validation for container fields, image names, duplicate definitions, and target architecture compatibility.
  * Added support for TCP and UDP exposed ports.
  * Container specifications are now included when loading projects.

* **Bug Fixes**
  * Invalid values, collection shapes, and command syntax now return clear, structured errors instead of failures.
  * Improved error messages with field-specific context.
  * Unsupported container objects are now reported as unexpected types.

### DIFF
--- a/crates/decode/src/container.rs
+++ b/crates/decode/src/container.rs
@@ -1,0 +1,1040 @@
+//! Decoding of container declarations, laid out at `<layer>/containers/<name>/spec.ncl`.
+
+use common::{Target, target::Arch};
+use nickel_lang_core::{
+    eval::{cache::CacheImpl, value::NickelValue},
+    program::Program,
+    term::{IndexMap, RuntimeContract},
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{Error, ObjTy, eval_if_closure, packages_array_from_term, record_data_from_val};
+
+/// The transport protocol of an [ExposedPort].
+///
+/// OCI recognises only `tcp` and `udp`, and spells them lowercase in the
+/// `ExposedPorts` keys of the image config.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum Proto {
+    #[default]
+    Tcp,
+    Udp,
+}
+
+impl std::fmt::Display for Proto {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Proto::Tcp => write!(f, "tcp"),
+            Proto::Udp => write!(f, "udp"),
+        }
+    }
+}
+
+/// Word-splits the string form of an argv field into OCI exec form.
+///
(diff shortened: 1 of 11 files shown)


crates/decode/src/container.rs
```diff
@@ -0,0 +1,1040 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{Error, ObjTy, eval_if_closure, packages_array_from_term, record_data_from_val};
+
+/// The transport protocol of an [ExposedPort].
+///
+/// OCI recognises only `tcp` and `udp`, and spells them lowercase in the
+/// `ExposedPorts` keys of the image config.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum Proto {
```
evanspearman: Nit: might be worth making the name more descriptive

crates/decode/src/container.rs
```diff
@@ -0,0 +1,1040 @@
...
+                        )?)
+                        .unwrap(),
+                    ))
+                },
+            )
+            .collect::<Result<IndexMap<_, _>, Error>>()
+    } else {
+        todo!("unexpected term for string map: {:?}", rt)
+    }
+}
```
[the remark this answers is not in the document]
evanspearman: This one probably needs to be addressed.

crates/decode/src/container.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+    /// Labels applied to the image.
+    pub labels: IndexMap<String, String>,
+    /// Additional image-config fields, applied verbatim.
+    pub config: IndexMap<String, String>,
+}
+
+impl Container {
+    /// Deserializes a container structure from the given nickel term tree.
+    pub fn from_term(rt: &NickelValue, program: &mut Program<CacheImpl>) -> Result<Self, Error> {
```
evanspearman, line 264: There are quite a few unwraps in this function on calls to deserialize. If these are protected by an invariant enforced elsewhere, we should document that through `expect`, but otherwise, we should handle the error.

evanspearman (changes requested):
Generally, there are quite a few unwrap calls outside of tests. We should avoid this if at all possible and use `expect` instead when protected from a panic by an invariant enforced elsewhere, but otherwise we should handle the errors.

twitchyliquid64:
I took this opportunity to fix the lazy error handling we've had in decode since forever, and replace it with a small handful of new error variants. Should be a lot harder to panic decodes now.

evanspearman (approved):

PR #1249, opened 2026-08-19
fix(init): use latest pkgs for upstream in min init

twitchyliquid64:
## Summary by CodeRabbit

* **Bug Fixes**
  * Existing project checkouts are now updated before the default packages repository is checked out when no upstream checkout is available.


bryan-minimal (approved):
LGTM

PR #1254, opened 2026-08-21
docs: add build reproducibility reference page

norrietaylor:
## The gap

Nothing in the published docs tells you how to make a package build byte-reproducible.

`docs/reference/build-specs.md` documents the `BuildSpec` field schema and then explicitly hands off the actual compilation:

> Each package also has a `build.sh` script adjacent to `build.ncl` that performs the actual compilation.

...and that is where the documentation stops. There is no page describing what `build.sh` must do to keep that compilation deterministic. Grepping `docs/` for `reproducib|determinis|SOURCE_DATE_EPOCH|byte-identical` finds only assertions *that* Minimal builds reproducibly ([`concepts/packages.md`](../blob/main/docs/concepts/packages.md), [`concepts/sessions.md`](../blob/main/docs/concepts/sessions.md), [`guide/building.md`](../blob/main/docs/guide/building.md)) — never how a package author achieves it.

That is a real gap, because reproducibility is not automatic. The sandbox makes the *environment* reproducible; making the *compiler output* reproducible is the build script's job and is toolchain-specific.

## What this adds

A new reference page, `docs/reference/reproducibility.md`:

- **Why it is required** — the artifact cache is content-addressed, so an artifact that varies between builds cannot be verified against the spec that claims to describe it.
- **What the sandbox already provides** — `SOURCE_DATE_EPOCH=0` and `PYTHONHASHSEED=0`, and the fact that re-assigning either one *fails* the `mip check` build-script audit. This is a genuine footgun: the obvious "belt and braces" move of exporting `SOURCE_DATE_EPOCH=0` yourself is rejected, and until now nothing published said so.
- **Per-toolchain recipes** — C/C++ (`-ffile-prefix-map`, `-gno-record-gcc-switches`, `--build-id=none`, `ARFLAGS=Drc`, deterministic archives, `.la` removal), Go (`-trimpath`, `-ldflags "-buildid="`, `-buildvcs=false`), Rust (`--remap-path-prefix`, and the `codegen-units=1` / `CONST_RANDOM_SEED` fallback), the Linux kernel (`KBUILD_BUILD_*`), and the case of a build that stamps its own wall-clock time regardless of `SOURCE_DATE_EPOCH`. Framed as *pick the one that matches your build system*, not a checklist.
- **How to verify** — build twice, `diff -r` the two `$OUTPUT_DIR` trees, plus a table mapping each common diff signature (embedded date, absolute path, differing `.note.gnu.build-id`, non-deterministic `.a`, varying uid/gid) back to the flag that fixes it.
- A link to [reproducible-builds.org](https://reproducible-builds.org/) for the general background and full taxonomy.

## Verification

The two environment claims are checked against this tree, not taken on faith:

- `crates/sandbox2/src/config.rs` sets `SOURCE_DATE_EPOCH=0` and `PYTHONHASHSEED=0` for isolated (build) sandboxes.
- `crates/check/src/lib.rs` `BUILD_SCRIPT_AUDIT_REGEXES` fails a build script that assigns either, with the message quoted in the page. The same audit rejects `tar -xf` in favour of `tar -xof`, which the page's diff-signature table also covers.

Every relative link and anchor in the new page resolves (`./build-specs.md`, `./cli-mip.md#check`, `../concepts/packages.md`, `../concepts/sandboxing.md`).

## Placement

A new page rather than a section inside `build-specs.md`. That page is 87 lines of intro / example / field-schema table, and this content is neither a field nor a schema — it is `build.sh` guidance, which `build-specs.md` explicitly scopes out. `docs/reference/` already carries topical non-schema pages (`sandbox-operations.md`, `linux-host-setup.md`), so a dedicated page fits how the section is already organised. Registered in `docs/reference/manifest.json` and linked from `build-specs.md` at the point where `build.sh` is introduced.

## Downstream

`gominimal/pkgs` currently carries this material inlined in its `AGENTS.md` (`### Reproducibility (required)`), which is where the text here was adapted from. That is a duplication that has already drifted in other sections, so once this page is live, `pkgs` will replace its inlined copy with a link to `https://minimal.dev/docs/reference/reproducibility`. A follow-up draft PR in `pkgs` stages that routing change plus a docs-drift canary, and is deliberately blocked on this PR merging — linking a URL that still 404s would be worse than the duplication.

Adapted for a public audience: pkgs-specific command spellings and repo conventions were dropped, and the one package-specific worked example (`nspr`) was generalised into the pattern it illustrates.

## Summary by CodeRabbit

* **Documentation**
  * Updated build specification examples to reflect current contracts, dependency formats, output and test definitions, and command syntax.
  * Added guidance on version propagation through build arguments.
  * Added documentation for achieving and verifying byte-reproducible package builds across supported toolchains.
  * Documented common reproducibility issues, diff signatures, and troubleshooting references.


twitchyliquid64 (approved):
Lovely!

norrietaylor:
Downstream consumer, now open: [gominimal/pkgs#636](https://github.com/gominimal/pkgs/pull/636) — replaces the inlined `### Reproducibility (required)` section in `pkgs` `AGENTS.md` with a link to the page this PR adds, and arms a liveness + drift canary over that link. It is a draft and its CI is deliberately red until this merges and deploys, because the URL 404s until then.

One thing worth flagging for this PR specifically: the published pages embed the reference sidebar, so adding a page to `docs/reference/manifest.json` changes the raw HTML of **every** reference page. Anything hashing whole pages will see a full-set drift on deploy — `gominimal/minimal-skills` `evals/docs-snapshots.json` currently reports drift on all 9 of its URLs for what looks like exactly this reason, and will need a refresh after this lands. The pkgs canary hashes only the `<main>` text to avoid that class of false positive.

norrietaylor:
### Added: the `build-specs` example was broken, not just incomplete

Pushed `36be75c` onto this branch. While confirming what the reproducibility page should link back to, I found the primary example on `build-specs.md` could not be copied and run.

**It destructured a symbol that does not exist.** The example did:

```ncl
let { build_spec, .. } = import "minimal.ncl" in
build_spec { ... }
```

`crates/stdlib/minimal-ncl/minimal.ncl` exports the `BuildSpec` contract (line 10) and a `build` constructor (line 39, `fun spec => spec | BuildSpec`). The string `build_spec` appears **nowhere** in the Nickel stdlib; the only repo-wide matches are unrelated Rust locals in `graph.rs` and `cmd_pkg_dep.rs`. Destructuring a missing field fails at evaluation.

**The dependency syntax was stale in the same way.** `{ pkg = "make" }` and `{ src = { ... } }` have no counterpart: `BuildDep` dispatches on a `ty` tag, which is exactly what the `Source` and `Local` contracts set, and a package dependency is a Nickel import of its `build.ncl` used directly in the array.

The example also omitted pieces every working package needs, most importantly the `Local` entry for the `build.sh` that `cmd` invokes.

Rewritten against the real `jq` spec in `gominimal/pkgs`, plus a short note naming the load-bearing shapes. Also corrected the `cmd`/`cmds` schema row: the stdlib types `cmd` as `CmdSpec` (String or Array String) and `cmds` as `Array CmdSpec`, not the two array forms the table listed.

**Verification.** I could not run a Nickel evaluator locally (no host `nickel`, and the task sandbox I normally use for `min check` began hanging part-way through this work). What I did instead was check every construct against the 466 specs in the public registry, all of which pass pkgs CI on every run:

| Construct | Packages using it |
|---|---|
| `} \| BuildSpec` | 466 |
| `{ file = ... } \| Local` | 464 |
| `} \| Source` | 450 |
| `cmd = "./build.sh"` (string form) | 461 |
| `\| Attrs` | 421 |
| `\| OutputBin` / `OutputLib` / `OutputData` | 319 / 210 / 338 |
| `standaloneTest` | 58 |
| **`build_spec`** | **0** |
| **`{ pkg = ... }`** | **0** |
| **`{ src = ... }`** | **0** |

Worth a reviewer with a working evaluator confirming it parses before merge.

### Why this matters beyond the page

We were about to route `gominimal/pkgs`'s `AGENTS.md` at this page, replacing its inlined build.ncl guidance with a link. Had we done that, agents would have been handed an example that does not evaluate, which is strictly worse than the duplication we were removing. That routing is now on hold until this lands.

Still undocumented and worth separate issues: `subsetOf`, and the `OutputBin`/`OutputLib`/`OutputData` and `Local`/`Source` contracts are named in the schema table but defined nowhere.

norrietaylor:
Addressed the review in 2d75d8b. All nine findings were acted on; notes on the two where the underlying claim needed adjusting before I applied it.

**Checksum.** Confirmed and fixed. I downloaded `jq-1.8.1.tar.gz` and hashed it: `2be64e7129cecb11d5906290eba10af694fb9e3e7f9fc208a311dc33ca837eb0`, matching jqlang's published `sig/v1.8.1/sha256sum.txt`. The old value was wrong.

**`--enable-deterministic-archives`** — fixed, but the stated mechanism was off. Autoconf does not *reject* unknown `--enable-*` options; it emits an `unrecognized options` warning and continues. Six packages in the registry pass this flag today, four of them not binutils, and they build. The real problem is that it is a no-op outside binutils, where it sets the default for the `ar`/`ranlib` being installed. I scoped it to binutils and made `ARFLAGS=Drc` the general advice, rather than dropping it outright.

**`tar -xof`** — fixed as described. Worth recording that restoring an archive's recorded mtimes is not itself a reproducibility problem, since those timestamps are fixed by the archive content. The row was wrong because it credited `-xof` with fixing wall-clock mtimes written by build steps, which is a separate cause with a separate fix. Split accordingly.

**`CONST_RANDOM_SEED`** — correct, and verified at the source: const-random's README documents that the crate reads this variable to make builds deterministic. Reattributed to the crate.

Also verified while here: `SOURCE_DATE_EPOCH=0` and `PYTHONHASHSEED=0` are set at `crates/sandbox2/src/config.rs:318-319` and rejected by the build-script audit at `crates/check/src/lib.rs:1170-1175`; the `--no-cache`/`--no-fetch`/`--rebuild` flags now cited in the verification section match `docs/reference/cli-mip.md`; and every URL in both files returns 200.

norrietaylor:
@coderabbitai review

PR #1259, opened 2026-08-21
fix(graph): treat an unparseable layer-cache entry as a miss (#1214)

### LINKED ISSUE #1214: diag: The long-running in‑VM minimald (up 21h40m, started 2026‑08‑12T20:56Z) is executing pre‑upgrade code rel

**Bundle:** `minimal-diag-20260813T183647Z.tar.zst`
**Reporter context:** _(none)_

---

# Diagnosis

## Verdict
The long-running in‑VM `minimald` (up 21h40m, started 2026‑08‑12T20:56Z) is executing **pre‑upgrade code** relative to the binaries installed at 2026‑08‑13T18:03:15Z. Ninety seconds after that install, the first session hit a graph layer‑cache entry written in the *new* format (the `profiles` field was removed upstream), the stale daemon's deserializer required `profiles`, and `LayerCacheDir::get` fails **closed** — aborting the whole graph load, so session `task-vet-7c5f` was created with **no package environment** (no `base/git/go/make/vim` PATH/env wiring). The condition self‑healed (the broken cache file is deleted on read error), so exactly one session was affected.

## Confidence
**Medium‑High.** The timestamp coincidence (install 18:03:15 → failure 18:04:45), the direction of the serde error, and the guest daemon's uptime are all consistent. It would be raised by: the contents of `~/.local/state/minimal/installed` (only its mtime/size is in the bundle), a build hash / `long_version` that differs between binaries (both report bare `0.5.1`, so the version string cannot distinguish builds), and a listing of `cache/lc/*` mtimes in the guest (the state listing was truncated before reaching it).

## Skew check
- `manifest.json` `.version` (host CLI that took the bundle): **0.5.1**
- `providers/local-minvmd0/socket-probe.json` get_version: daemon **0.5.1**, long_version **0.5.1**, stdlib **0.0.18**
- Guest `daemon-diag/meta.json`: **0.5.1**, stdlib **0.0.18**, pid 1, `uptime_secs 78019.66` (start ≈ 2026‑08‑12T20:56:28Z)
- `logs/minvmd.log.2026-08-12`: minvmd **0.5.1** (spawned 2026‑08‑12T20:56:28Z)
- `providers/local-minvmd0/minvmd.toml`: `started_at = 1786568189` = 2026‑08‑12T20:56:29Z; `host/process-tree.txt`: pids 6859/6860/6861 `etime 21:40:19`
- Install marker `state/listing.txt`: `f 2067 1786644195 644 installed` = **2026‑08‑13T18:03:15Z** — i.e. all three running host/VM processes and the guest daemon **predate the currently installed binaries by ~21 h**.

Version *strings* agree everywhere; the skew is a **build/generation skew hidden behind an unchanged version number**, proven by process age vs. install time and by the schema direction of the serde error.

**skew is the root cause** (of the one observed fault; its blast radius was small and self‑limiting).

## Evidence
- `providers/local-minvmd0/guest/daemon-diag/logs/minimald.log.2026-08-13` (and the same lines in raw console `providers/local-minvmd0/boot.log:366-368`):
  - `18:04:45.382562Z WARN graph::loader: "LayerCacheDir::get failed to deserialize: missing field \`profiles\` at line 1 column 991667"`
  - `18:04:45.382771Z WARN minimald::sessions::composables: "graph resolution failed at CreateSession; package contributions will be empty", error="graph: Error: failed to fetch upstream layer: layer-cache fetch: ()", project_path="/var/lib/minimal/sessions/30a28/tree"`
  - `18:04:45.382832Z WARN "project declared packages but the graph is unavailable; package composables will be empty (no env_state_wiring vars, no env_dir/file_mappings patches)", packages="[\"base\", \"git\", \"go\", \"make\", \"vim\"]"`
- Mechanics in source: `crates/graph/src/loader.rs:316` emits that warning and **propagates** the error (`map_err(...)?`) after `std::fs::remove_file(&p)`; `crates/graph/src/loader.rs:423` turns it into `Error::Fetch(format!("layer-cache fetch: {:?}", e))`, so a single poisoned cache entry aborts `load_layer` instead of re‑decoding from source.
- Schema direction: current `Layer` (`crates/decode/src/lib.rs:65-76`) has **no** `profiles` field, and `crates/graph/src/wire.rs:35` says `// profiles were removed, tag is ignored`. "missing field `profiles`" can therefore only be produced by a *pre‑removal* reader parsing a *post‑removal* file → the reading daemon is older than the cache writer.
- Self‑heal confirmed: the very next session `task-test-ba19` at `18:04:56.976822Z` composed successfully — `compose: walking top-level packages ... top_level_names=["base","binutils","git","go","linux_headers","make","vim"]` — as did every session through `18:35:02` (`task-run-a352`, RPCs all `outcome="ok"`).
- Skew timing: `state/listing.txt` `installed` mtime `1786644195` (18:03:15Z) vs `host/process-tree.txt` `6859 … 21:40:19` and guest `meta.json uptime_secs 78019.66`.
- Non‑faults ruled out: `socket-probe.json` all stages ok (stat/connect/handshake/get_version); `status.json` `lifecycle=running, minvmd_alive=true`; guest `disk.json` `/var/lib/minimal` 247 GB free; host 55 GB free; `logs/kmsg.txt` clean (`EXT4-fs (vdb): recovery complete`, no I/O errors, no OOM); `host/power.txt` `kern.sleeptime = 0` (no sleep/wake involvement); `min bug` (pid 26530) sample shows a normal `_pthread_cond_wait` collect, not a hang.

## Failure chain
1. 2026‑08‑12T20:56:28Z — `minvmd run` + gvproxy + VMM start; guest `minimald` (pid 1) boots. This build stays resident for the rest of the incident.
2. 2026‑08‑13T17:18–17:58Z — normal session churn; benign teardown noise (`pty master error; tearing down host … I/O error (os error 5)`, `bash exited 127` for sessions whose `top_level_names=[]`, i.e. no packages declared yet).
3. 2026‑08‑13T18:03:15Z — new minimal binaries installed on the host (`state/listing.txt:installed` mtime). The running VM/daemon are **not** restarted.
4. 2026‑08‑13T18:04:45.17Z — `CreateSession` for `task-vet-7c5f`; graph loader reads a `cache/lc` entry serialized in the new (`profiles`-less) format.
5. 18:04:45.38Z — stale deserializer: `missing field 'profiles'` → broken file deleted → error propagated as `layer-cache fetch: ()` → `graph resolution failed at CreateSession` → session composed with **empty package contributions** despite declaring `base, git, go, make, vim`.
6. 18:04:56Z onward — cache repopulated by re‑decode; all later sessions compose the full 7‑package set; `min bug` collected at 18:36:47Z.

## Recommended next steps
User:
- Restart the provider so the freshly installed code actually runs: stop/start `local-minvmd0` (`min vm restart` / `minvmd` stop+run), then re‑run the task that failed (`task-vet`). Any session created between 18:03:15 and the restart may have run without its package environment and should be re-run.
- If a stale entry recurs, clearing the guest layer cache (`/var/lib/minimal/cache/lc`) is safe — it is a pure cache.

minimal team:
- `crates/graph/src/loader.rs:311-320` — treat a cache deserialize failure as a **miss**, not an error: after `remove_file`, return `Ok(None)` so `load_layer` (`loader.rs:407-433`) falls back to `Layer::new`. Today one poisoned file fails an entire session's graph.
- Add a schema/format version (or version-namespaced subdirectory) to `LayerCacheDir` keys so cache files written by one build are never parsed by another; the `Layer` struct changed shape (`profiles` removal, `wire.rs:35`) without a version bump.
- Make upgrades detect a running daemon/VM older than the installed binaries and warn or auto-restart; identical `long_version` strings across differing builds make this skew invisible in `min --version` and in `socket-probe`. Include a build hash in `long_version`.

## Caveats
- The bundle does not contain the contents of the `installed` marker, so I infer "an upgrade happened at 18:03:15Z" from its mtime plus the immediate onset of the schema error; I cannot show the previous version number.
- The guest `state-listing.txt` was truncated by the listing cap before `cache/lc`, so I could not read the offending cache file's mtime or the writer's identity (in-guest `mip`, `mctx` project setup, or CLI-pushed entry — all three call `LayerCacheDir`, `crates/mip/src/cmd_run.rs:217`, `crates/mctx/src/project_setup.rs:104`, `crates/mctx/src/lib.rs:670`).
- The user gave no complaint text; if their actual pain was something else (e.g. the repeated `pty master error … os error 5` / `bash exited 127` at 17:31–17:33, which correlate with sessions that legitimately declared **no** packages, or the Tailscale/CGNAT own‑IP route warnings, which are informational and did not affect the vsock/unix-socket paths used here), the bundle shows no fault in those paths.
- Missing mTLS material (`config/client-cert.json` → `{"ca.pem": null, "client.pem": null}`, `host/dirs.txt` marks `client.pem`/`ca.pem`/`client.key` as missing) is expected for a local unix-socket provider and the probe handshake succeeded — not a fault. The `check` complaint "guest bundle missing errors.json" is a collector/manifest nit, not evidence of the incident.

norrietaylor:
Closes #1214.

`LayerCacheDir::get` deserialized a cache entry with `map_err(...)?`, so a single
unparseable file failed the whole graph load. The cache is an optimisation; a
corrupt entry should cost a re-evaluation, not the build.

- An entry that will not parse is now a **miss**: warn, best-effort remove, `Ok(None)`.
- Entries are namespaced under `v{CACHE_FORMAT_VERSION}` so entries written by one
  build are never handed to another build's deserializer.

The new `create_dir_all` in `insert` cannot introduce a fail-closed path: the sole
caller (`crates/graph/src/loader.rs`) is `if let Err(e) = self.lc.insert(...)`, which
logs and swallows.

### Test
`poisoned_layer_cache_entry_is_a_miss` warms a real `LayerCacheDir`, overwrites every
entry with `{"not": "a layer"}`, reloads, and asserts the graph is intact and the
cache re-populated with parseable entries.

### Reviewer note
The `v1/` namespace orphans any pre-existing entries at the cache root — they will
never be read and are never cleaned up. Harmless, but worth a follow-up if the cache
is large on developer machines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache reliability by isolating entries in a versioned cache namespace.
  * Corrupted or unreadable cache entries are now identified, removed, and regenerated from source data automatically.
  * Recovered cache entries are safely rewritten for subsequent use.
  * Enhanced diagnostics help identify problematic cache data while handling truncated, escaped, UTF-8, and binary content safely.

### DIFF
--- a/crates/graph/src/loader.rs
+++ b/crates/graph/src/loader.rs
@@ -295,6 +295,27 @@ impl LayerCacheDir {
     }
 }
 
+/// How much of a poisoned layer-cache entry to quote into the warning that precedes
+/// deleting it.
+const ENTRY_HEAD_BYTES: usize = 200;
+
+/// Renders the head of a layer-cache entry for a diagnostic, escaped so it stays on one
+/// log line. Truncated to [ENTRY_HEAD_BYTES], which is plenty: an entry we cannot parse
+/// is almost always JSON from another build's schema, and its first field names say so.
+fn entry_head(bytes: &[u8]) -> String {
+    let head = &bytes[..bytes.len().min(ENTRY_HEAD_BYTES)];
+    match std::str::from_utf8(head) {
+        Ok(s) => format!("{s:?}"),
+        // A multi-byte char cut in half by the truncation is not corruption: quote the
+        // valid prefix rather than writing the whole entry off as binary.
+        Err(e) if e.valid_up_to() > 0 => {
+            let valid = std::str::from_utf8(&head[..e.valid_up_to()]).expect("valid prefix");
+            format!("{valid:?}")
+        }
+        Err(_) => "<not valid UTF-8>".to_string(),
+    }
+}
+
 impl LayerCache for LayerCacheDir {
     type Error = ();
 
@@ -334,14 +355,26 @@ impl LayerCache for LayerCacheDir {
     fn get(&mut self, lo: &LoadOptions) -> Result<Option<Layer>, Self::Error> {
         let p = self.entry_path(lo);
 
-        if let Ok(f) = std::fs::File::open(&p) {
+        if let Ok(bytes) = std::fs::read(&p) {
             // An entry we can't parse is a miss, not a failure: drop it and let the
             // caller re-evaluate from source. Propagating here would fail the whole
             // graph load over a single poisoned file.
-            let layer: Layer = match serde_json_lenient::from_reader(f) {
+            let layer: Layer = match serde_json_lenient::from_slice(&bytes) {
                 Ok(layer) => layer,
                 Err(e) => {
-                    tracing::warn!("LayerCacheDir::get failed to deserialize: {}", e);
+                    // The entry is about to be unlinked, so this warning is the only
+                    // record that will survive it. Say which file, which cache key, how
+                    // big it was, and what it actually held -- the head of the JSON is
+                    // what tells you which build's schema wrote it.
+                    tracing::warn!(
+                        "LayerCacheDir::get failed to deserialize {} (input_hash {}, {} bytes): {}; \
+                         discarding the entry, the layer will be re-evaluated from source. Head: {}",
+                        p.display(),
+                        lo.input_hash().to_hex(),
+                        bytes.len(),
+                        e,
+                        entry_head(&bytes),
+                    );
                     std::fs::remove_file(&p).ok(); // best effort, delete broken file
                     return Ok(None);
                 }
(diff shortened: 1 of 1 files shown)


twitchyliquid64 (approved):
I'll note that I had it as an error so i could debug if/when it happens, what lead to this?

But yes long-term we do want it to warn but not explode

norrietaylor:
> I'll note that I had it as an error so i could debug if/when it happens, what lead to this?
> 
> But yes long-term we do want it to warn but not explode

If you still want visibility I can close this. This seems to be related to a session that bridged our profiles removal, breaking changes such as that aren't an every day occurrence.

twitchyliquid64 (approved):

norrietaylor:
`9788f93d` — kept your debugging signal rather than just softening it.

- Old warn logged the serde error but not *which* entry, then deleted the file — gone and unnamed
- Now: path, input hash, byte length, error, first 200 bytes quoted
- For schema drift the leading field names tell you which build wrote it

Chose delete over a `.corrupt` sidecar — a `Layer` shape change with no version bump poisons *every* entry at once, so the incident that motivates a sidecar is the one that doubles the cache on disk, and nothing GCs that dir.

PR #1260, opened 2026-08-21
fix(minimal): refuse a version-skewed daemon instead of self-destroying sessions (#1251)

### LINKED ISSUE #1251: diag: The host CLI was upgraded to 0.5.3-dev.20.g67f55884 roughly a minute before the bundle was taken, but th

**Bundle:** `minimal-diag-20260819T194703Z.tar.zst`
**Reporter context:** _(none)_

---

# Diagnosis

## Verdict
The host CLI was upgraded to **0.5.3-dev.20.g67f55884** roughly a minute before the bundle was taken, but the microVM is still running the **0.5.1-rc1.dev.23.g50e2a1a0** guest `minimald` (and a `minvmd` started the day before). Every `min` activation now dies at the `FinalizeSession` step: the old daemon finalizes the session successfully (it registers DNS and goes Active), the new client rejects/ mis-handles that response, bails out of `upload_and_finalize`, and immediately `DestroySession`s the box — so sessions appear to be created and vanish instantly. This is version skew.

## Confidence
**Medium-High.** The behavioural before/after boundary is exact (same daemon, same config, works at 18:54 with the old CLI, fails at 19:46:33 and 19:46:39 within seconds of the new CLI being installed), and the RPC sequence matches only one code path in `crates/minimal/src/lib.rs:2245-2257` (`upload_and_finalize` → `best_effort_destroy` on the *same* connection). It would be raised to High by (a) the CLI's stderr text from the failed run (`min` output is not in the bundle) and (b) the 0.5.1 `FinalizeSessionResponse` wire shape, to confirm which field trips the `#[serde(deny_unknown_fields)]` guard on `FinalizeSessionResponse` (`crates/minimald-rpc/src/lib.rs:420-432`).

## Skew check
- Host CLI that took the bundle (`manifest.json` `.version`): **0.5.3-dev.20.g67f55884**
- Provider `local-minvmd0` daemon (`socket-probe.json` get_version): **0.5.1-rc1.dev.23.g50e2a1a0**, `long_version` same, **stdlib 0.0.19**
- Guest bundle `meta.json`: **0.5.1-rc1.dev.23.g50e2a1a0**, stdlib **0.0.19**, pid 1, uptime 20250 s (started 2026-08-19 ~14:09Z, guest-internal reboot)
- `minvmd.toml` carries no version; host `minvmd run` (pid 39108) and `__krun-vmm` (pid 39110) have `etime 23:02:21` → started **2026-08-18 20:44:42Z**
- Install marker `state/listing.txt`: `f 2258 1787168769 644 installed` → **2026-08-19 19:46:09Z**, i.e. binaries were installed **54 s before the bundle** and **~23 h after** `minvmd`/the VM started, and ~5.6 h after the guest daemon started.

Every running daemon predates the currently installed binaries; the CLI is two minor-dev generations ahead of the guest daemon. → **skew is the root cause**.

## Evidence
- Working activation with the *old* CLI, `providers/local-minvmd0/guest/daemon-diag/logs/minimald.log.2026-08-19`:
  `18:54:50.612 … "minimald-v1-FinalizeSession" … rpc served outcome ok` → new connection → `minimald-v1-GetSessionRecord` → `minimald::session_host "session content" … "binding attached to session channel"` (session really started, ran until 18:55:10).
- Failing activations after the upgrade, same file / same daemon:
  - `19:46:33.434 rpc dispatched minimald-v1-FinalizeSession` → `19:46:33.435 registered PTask hostname … "ox-c521"` → `19:46:33.435 rpc served outcome="ok"` → **`19:46:33.436 rpc dispatched minimald-v1-DestroySession`** → `session destroyed … "ox-c521"` → `connection closed by peer reason="Protocol error: early eof"`.
  - Identical sequence 6 s later for `ox-9a91` (`19:46:39.396`–`19:46:39.399`).
  - Note: the daemon *succeeded* (DNS registered = session promoted to Active) yet the client tore it down on the **same** connection — that is `best_effort_destroy(&mut client, id)` in the `upload_and_finalize` error arm (`crates/minimal/src/lib.rs:2245-2257`, helper at `:1635-1677`). No `session_host` records, no exec/attach channel.
- Version identities: `providers/local-minvmd0/socket-probe.json` → `version=0.5.1-rc1.dev.23.g50e2a1a0, stdlib_version=0.0.19`; `…/guest/daemon-diag/meta.json` same; `manifest.json` `"version": "0.5.3-dev.20.g67f55884"`.
- Freshly installed binaries: `state/listing.txt` `installed` mtime `1787168769` = 2026-08-19T19:46:09Z; bundle `created_at 2026-08-19T19:47:03Z`; `host/process-tree.txt` shows `minvmd run` with `etime 23:02:21`.
- Wire-contract fragility that explains the client-side rejection: `crates/minimald-rpc/src/lib.rs` — *"`deny_unknown_fields` is load-bearing… Adding `activate_hooks` reopened exactly that hole, and the symptom is silent"* — a `FinalizeSessionResponse` whose shape differs across versions decodes as an RPC failure.
- Unrelated/incidental findings (documented so they are not mistaken for the cause):
  - Remote-cache 404s for the `claude-code` build, twice, both **before** the upgrade: `18:13:42.786 ERROR orchestrator cache_fetch { hash: "d9a1e5…", spec: "claude-code" } construction failed … No such object: minimal-staging-cache/4819811a…zst` and again `18:55:01.868`.
  - Stale config keys: 46× `WARN Ignoring declared profile 'dev', profile support was removed`, `[defaults]profile = "dev" defined, profiles were removed in 0.5.1` (from `config/loadouts/dev.toml.redacted` / `config.toml`).
  - `WARN pty master error; tearing down host error="I/O error (os error 5)"` at 19:46:27 — normal PTY EIO when the shell exits (bash exit 1); 19 occurrences going back to 8/18.
  - Tailscale CGNAT/own-IP route warnings in `host/net/routes.txt` — cosmetic here; the transport is a vsock/UDS, and every probe stage passed in 0 ms.
  - `check`'s "guest bundle missing errors.json" is a guest-collector version difference (again 0.5.1 guest vs 0.5.3 reader), not corruption.

## Failure chain
1. 2026-08-18 20:44:42Z — `minvmd run` + `gvproxy-min` + `__krun-vmm` started (old release).
2. 2026-08-19 ~14:09Z — the guest kernel rebooted in place (`kmsg.txt`: `EXT4-fs (vdb): 22 orphan inodes deleted / recovery complete`), bringing up guest `minimald` **0.5.1-rc1.dev.23** as pid 1.
3. 18:13 and 18:55Z — two `claude-code` sessions fail on a remote-cache 404 (separate, pre-existing issue); 18:54:50 activation otherwise works end-to-end, proving CLI↔daemon compatibility at that point.
4. **19:46:09Z** — new binaries installed (`installed` marker); `min` is now 0.5.3-dev.20 while the VM keeps running the 0.5.1 daemon.
5. 19:46:23–19:46:28Z — the last old session's shell exits (bash exit 1, pty EIO teardown).
6. 19:46:33.43Z — first post-upgrade activation: Create → ConfigureLoadout → **Finalize OK server-side** → client rejects the response → `DestroySession` 1 ms later; session `ox-c521` gone.
7. 19:46:39.39Z — repeat, `ox-9a91`, identical.
8. 19:47:03Z — the user runs `min bug` (`min bug` pid 4605, etime 00:00).

## Recommended next steps
1. Restart the provider so the guest daemon matches the CLI: stop and re-start the VM (`min vm stop`/`min vm start` for `local-minvmd0`, or kill `minvmd run` pid 39108 and let `min` relaunch it) — the VM and its guest image must be re-launched after every `min` upgrade. Then confirm `min version` shows the same version for CLI and daemon (`crates/minimal/src/lib.rs:3637-3647` prints daemon + stdlib).
2. Re-run the failing activation and capture `min`'s stderr; expect the message from `upload_and_finalize` (`FinalizeSession RPC failed` / `FinalizeSession failed: …`, `crates/minimal/src/lib.rs:1652-1676`).
3. For the minimal team: a self-destroying activation is a terrible symptom for a wire-shape mismatch. Add an explicit version/compat check after `GetVersion` at `connect_daemon` time (the CLI already fetches it in `crates/minimal/src/diag/net.rs:145`), and have `min` refuse-with-advice ("daemon is 0.5.1, CLI is 0.5.3 — restart the VM") rather than tearing down sessions; also make the `upload_and_finalize` failure print the decode error verbatim.
4. Separately: republish or rebuild `4819811a61f2…zst` in `minimal-staging-cache` (the `claude-code` dependency 404s), and drop the removed `profile = …` keys from `~/.config/minimal/config.toml` and `loadouts/dev.toml`.

## Caveats
- The bundle contains **no host-side CLI log or stderr** (minimald runs in the guest; `logs/minimald.log*` is legitimately absent per `logs/PROVENANCE.txt`), so the exact client error text for the failed finalize is inferred from the RPC pattern plus source, not read from a log.
- I could not read the 0.5.1 `FinalizeSessionResponse` definition (the source tree is a single-commit shallow clone at 0.5.3), so "which field mismatch" is a hypothesis; an alternative mechanism with the same signature would be a new-CLI patch/hook-script upload that the old daemon silently accepts but leaves without the patches-ready marker, causing finalize to answer `Errorable::Err` (the daemon logs `outcome="ok"` either way). Both are skew; the remediation is identical.
- The `claude-code` cache 404 is real and would still break that particular package after the restart; it is not what killed the 19:46 sessions.
- Guest-side reboot at ~14:09Z has no recorded cause in `kmsg.txt` (log starts at that boot); host `power.txt` shows only routine sleep/darkwake cycles, none coincident with the failures.

norrietaylor:
Closes #1251.

A CLI talking to a daemon from another build would fail partway through an
activation and tear down whatever it had already created. There was no version
check, and the failure was undiagnosable: every oneshot response is wrapped in an
`#[serde(untagged)]` `Errorable<T>`, and an untagged enum discards its per-variant
errors in favour of "data did not match any variant" — so a `deny_unknown_fields`
guard doing its job produced an error that could not name the offending field.

**`deny_unknown_fields` is not relaxed.** Two changes instead:

1. **A version gate.** `ensure_version_match` compares CLI and daemon versions and
   refuses with a message naming both builds and prescribing the recovery
   (`min stop`, then re-run). `MINIMAL_ALLOW_VERSION_SKEW=1` overrides with a warning.
2. **Decode errors quote the daemon's reply** verbatim, bounded to 1 KiB, so a bug
   report can name the field that tripped the guard without a live repro.

### Every daemon connection is now classified

`min stop` and the Ctrl-C cleanup paths are deliberately **ungated** — a cleanup that
refuses to run *is* the orphaned session, and `min stop` is the recovery the gate's
own message prescribes. Gated: `cmd_attach`, `cmd_exec`, `cmd_session_setup_zed`,
`cmd_ssh_forward`, and the dash `activate` path. Every ungated site carries a one-line
reason so the next reader does not have to redo the audit.

Gating the TUI required moving `ensure_version_match` / `version_skew_message` /
`SKEW_OVERRIDE_VAR` into `minimal-client` (the crate both front ends share), since
`minimal-tui` cannot depend on `minimal`. The message text and override semantics are
byte-for-byte unchanged.

### Tests
`every_daemon_connection_is_classified` and `only_the_activation_path_is_version_gated`
scan the sources at test time and assert the full gated/ungated inventory — a new
`Client::connect` fails the build until someone repeats this judgement. Both were
observed failing on a wrong classification before being made to pass.

### For reviewers
- `cmd_session_setup_zed` is the judgement call: read-only against the daemon, but it
  bakes daemon-sourced identity into persistent editor config. Gated on those grounds;
  a strict "multi-step mutation" reading would leave it ungated.
- `completion.rs` already routed through `connect_daemon`, so shell completions now
  fall silent against a skewed daemon. Pre-existing, untouched here.
- The source-scanning tests are an unusual shape for this tree and will need updating
  whenever a connection is legitimately added. That is the point, but it is a
  maintenance cost worth an explicit yes/no.
- `just e2e` / `just test-vm` not run (no KVM in the dev sandbox). `cmd_attach`,
  `cmd_exec`, and dash activate each now pay one extra `GetVersion` round trip.

## Summary by CodeRabbit

* **New Features**
  * Added daemon and CLI version compatibility checks across session activation, attach, exec, and related workflows.
  * Added clear version-skew errors, including support for legacy or unversioned daemons.
  * Daemon responses now report their version to improve compatibility diagnostics.
  * Added an override option for recovery and diagnostic operations.

* **Bug Fixes**
  * Improved undecodable response errors with a bounded excerpt of the original reply.
  * Prevented incompatible daemons from proceeding with session creation or activation.

### DIFF
--- a/crates/minimal-tui/src/rpc.rs
+++ b/crates/minimal-tui/src/rpc.rs
@@ -87,6 +87,11 @@ pub async fn discover(minimal_dir: Option<&Path>) -> Vec<Provider> {
         if !sock.exists() {
             continue;
         }
+        // Deliberately not version-gated: the dashboard is read-only until the
+        // user acts, it renders each provider's version in the sidebar, and a
+        // skewed daemon is precisely when an operator needs to see (and be able
+        // to destroy) the sessions on it. The one multi-step mutation the TUI
+        // drives, `activate`, gates on its own connection instead.
         match Client::connect(&sock).await {
             Ok(client) => providers.push(Provider {
                 label,
@@ -112,6 +117,8 @@ pub async fn connect_missing(providers: &mut Vec<Provider>, minimal_dir: Option<
         if !sock.exists() {
             continue;
         }
+        // Deliberately not version-gated, for the same reason as `discover`:
+        // a provider that reappears mid-run must still be listable.
         match tokio::time::timeout(CONNECT_TIMEOUT, Client::connect(&sock)).await {
             Ok(Ok(client)) => {
                 tracing::info!(provider = label, "provider connected");
@@ -249,6 +256,11 @@ pub async fn activate(
     contribution: sessions::wire::request::WireContribution,
 ) -> Result<SessionId, anyhow::Error> {
     let mut client = Client::connect(sock).await?;
+    // The dashboard's own copy of the create/upload/configure/finalize
+    // sequence, so it needs the same gate `min session activate` gets: on a
+    // skewed pair `FinalizeSession` fails after the record exists, and the
+    // abort below then destroys the session the user just created (#1251).
+    minimal_client::ensure_version_match(&mut client).await?;
     let id = match client
         .oneshot_rpc::<CreateSession>(CreateSessionRequest {
             config: SessionConfig {
@@ -456,5 +468,44 @@ mod tests {
         std::fs::write(dir.path().join(mfile::MFILE_NAME), "not valid toml = =").unwrap();
         let path = camino::Utf8PathBuf::from_path_buf(dir.path().to_path_buf()).unwrap();
         assert!(resolve_upload_root(&path).is_err());
+    }
+}
+
+#[cfg(test)]
+mod version_gate_tests {
+    /// The dashboard opens three daemon connections, and #1251's gate reached
+    /// none of them. Only `activate` runs a multi-step mutation, so only it is
+    /// gated; the two discovery connections stay open on purpose and must say
+    /// so, since a reader who finds an ungated connect has no other way to tell
+    /// an audited decision from an oversight.
+    #[test]
+    fn only_the_activation_path_is_version_gated() {
+        // Concatenated so this scanner does not match itself.
+        const NEEDLE: &str = concat!("Client", "::connect(");
+        let src = include_str!("rpc.rs");
+
+        let mut gated = 0;
+        let mut explained = 0;
+        for (i, line) in src.lines().enumerate() {
+            if !line.contains(NEEDLE) {
+                continue;
+            }
+            // The classification is always within a few lines of the connect:
+            // the gate call just after it, or the rationale comment just before.
+            let window: Vec<&str> = src.lines().skip(i.saturating_sub(8)).take(20).collect();
+            let text = window.join("\n");
+            if text.contains("ensure_version_match") {
+                gated += 1;
+            } else if text.contains("not version-gated") {
+                explained += 1;
+            } else {
+                panic!("unclassified daemon connection at rpc.rs:{}", i + 1);
+            }
+        }
+        assert_eq!(
+            (gated, explained),
+            (1, 2),
+            "the connection inventory changed"
+        );
     }
 }
--- a/crates/minimal-tui/src/rpc.rs
+++ b/crates/minimal-tui/src/rpc.rs
@@ -260,8 +260,11 @@ pub async fn activate(
     // sequence, so it needs the same gate `min session activate` gets: on a
     // skewed pair `FinalizeSession` fails after the record exists, and the
     // abort below then destroys the session the user just created (#1251).
-    minimal_client::ensure_version_match(&mut client).await?;
-    let id = match client
+    // The gate is the `must_match_version` on the create below and the version
+    // the reply echoes back — not a `GetVersion` ahead of them. Activation is
+    // the one path where an extra round trip is felt, and the create can carry
+    // the check for free.
+    let created = match client
         .oneshot_rpc::<CreateSession>(CreateSessionRequest {
             config: SessionConfig {
                 name,
@@ -274,13 +277,23 @@ pub async fn activate(
                 hooks_enabled: true,
                 attrs: Default::default(),
             },
+            // A daemon of another build refuses this before it allocates
+            // anything; `None` under the skew override, which is how an
+            // operator still gets through.
+            must_match_version: minimal_client::version_assertion(),
         })
         .await
         .context("CreateSession RPC failed")?
     {
-        Errorable::Ok(resp) => resp.id,
+        Errorable::Ok(resp) => resp,
         Errorable::Err { error } => return Err(anyhow::anyhow!(error)),
     };
+    // A daemon that predates the field ignored the assertion and echoes no
+    // version; that silence is itself the skew. Refuse now — before the
+    // upload and the finalize — leaving an unfinalized record the daemon
+    // reaps when this connection drops.
+    minimal_client::ensure_version_reported(created.daemon_version.as_deref())?;
+    let id = created.id;
 
     let flow = async {
         // The upload-root walk and VCS-root stat are blocking filesystem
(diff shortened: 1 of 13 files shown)


crates/minimal-tui/src/rpc.rs
```diff
@@ -249,6 +256,11 @@ pub async fn activate(
     contribution: sessions::wire::request::WireContribution,
 ) -> Result<SessionId, anyhow::Error> {
     let mut client = Client::connect(sock).await?;
+    // The dashboard's own copy of the create/upload/configure/finalize
+    // sequence, so it needs the same gate `min session activate` gets: on a
+    // skewed pair `FinalizeSession` fails after the record exists, and the
+    // abort below then destroys the session the user just created (#1251).
+    minimal_client::ensure_version_match(&mut client).await?;
```
twitchyliquid64: This works but adds another RTT to the activate path, something we've been trying to avoid. Maybe make the create session RPC carry an optional `must_match_version` string and have that RPC fail if minimald is a different version?

norrietaylor: Good call — `d6c23cd9`, no extra RTT on activate.

- `must_match_version` on `CreateSession`, checked before `create_session` — skewed create refused, nothing allocated
- Same for `cmd_attach`/`cmd_exec`/`setup_zed`/`ssh_forward`, folded into the `GetSessionRecord`/`ListSessions` they already send
- `connect_daemon` keeps `GetVersion`: fronts ~12 commands with differing first RPCs, none hot — stays safe-by-default

One thing I had backwards: `CreateSessionRequest` isn't `deny_unknown_fields`, so an old daemon *ignores* the field and creates the session — silently reopening #1251 for the commonest skew. Handshake is reciprocal now: daemon echoes `daemon_version`, absence = skew.

Trade-off: against an old daemon the create happens then we refuse, leaving an unfinalized record for the close-reap. `GetVersion` refused before anything existed. Shout if you'd rather keep the pre-flight there.

twitchyliquid64 (approved):

Issue #1210, opened 2026-08-12
diag: Two things happened, one causal and one cosmetic-but-real.


norrietaylor:
_Filed from [diag-portal](https://diag-portal.minimal-410.workers.dev)._

**Bundle:** `minimal-diag-20260812T163943Z.tar.zst`
**Reporter context:** Session was disconnected abruptly. Mouse events now written to terminal

---

# Diagnosis

## Verdict
Two things happened, one causal and one cosmetic-but-real. The session ended because the sandboxed process tree inside the guest was **SIGKILLed** (hakoniwa reported `code=125 … container received signal SIGKILL`), which closed every pty slave and surfaced as `EIO` on the pty master — minimald then tore the attachment down via the `ProcessExited` path. That path is the **only** teardown path in `minimald` that does *not* send the terminal "unwind codes" to the attached client, so the user's terminal was left with mouse reporting (and any other DEC private modes zellij had set) still enabled — hence mouse movements now printing escape sequences. The terminal corruption is a minimald teardown bug (`crates/minimald/src/session_host.rs`), not a client/terminal problem.

## Confidence
**High** for the mouse-events root cause (log-confirmed teardown reason plus the exact code path that omits the reset). **Medium-Low** for *why* the sandbox got SIGKILLed — the bundle contains no killer attribution (no kernel OOM in the guest ring buffer, no `killing session process` log from minimald). Guest-side `dmesg`/audit at incident time, or a `min` client-side log for that connection, would raise it.

## Skew check
- `manifest.json` `.version` (CLI that took the bundle): **0.5.0**
- `providers/local-minvmd0/socket-probe.json` get_version: daemon **0.5.0**, long_version **0.5.0**, stdlib **0.0.18**
- Guest `providers/local-minvmd0/guest/daemon-diag/meta.json`: **0.5.0**, stdlib **0.0.18**, pid 1, uptime 945097 s (≈10 d 22 h → started 2026‑07‑30 06:51)
- Host `minvmd` process 12010/12012 etime 13‑09:48 → started 2026‑07‑30 06:51; installed marker mtime `1785393923` = 2026‑07‑30 06:45:23 UTC → **binaries installed 6 minutes before the daemons started**, so nothing is running pre-install code.
[...]

PR #1261, opened 2026-08-21
fix(minimald): unwind the terminal when the session process exits (#1210)

norrietaylor:
Addresses the remaining half of #1210.

#1253 fixed the *notification* half — the user is now told when a session is
SIGKILLed — but it made the other half more visible: on teardown the terminal was
left in a broken state. Three of the four teardown arms wrote `self.unwind_codes()`;
the `TeardownDueToProcessExit` arm did not, so a session whose process simply exited
left mouse reporting, the alt screen, and cursor visibility as the session had set them.

The ProcessExit arm now emits the same unwind sequence as its siblings, and the client
gained a blind unwind so the reset still happens against an old daemon or a dropped
transport.

### Reviewer notes
- The client's blind unwind writes `\x1b[?1049l` unconditionally, while the daemon's
  `unwind_codes()` gates it behind `if live.alternate_screen()`. On xterm-family
  terminals rmcup-from-the-normal-buffer does a DECRC-style cursor restore, so after a
  clean detach the cursor can jump to a previously-saved position. Cosmetic — but it
  is cosmetic damage introduced by a fix about cosmetic damage. The doc comment's claim
  that every sequence "is a no-op against a terminal that was never in that mode" holds
  for every code in the set *except* `?1049l`; at minimum that claim wants qualifying.
- `session_via_ssh`'s interactive arm changed from `exec()` to fork+wait+exit, so `min`
  now sits between the pty and ssh. A signalled ssh used to make `min` die by that
  signal; it now exits 128+n. `$?` is identical, but a supervisor calling `waitpid` sees
  `WIFEXITED` where it saw `WIFSIGNALED`. Intentional and unit-tested.
- `scripts/e2e-attach-pty.py`'s hang guard SIGKILLs the pty child by pid — now `min`,
  not ssh — so a wedged attach can orphan an ssh plus its `min proxy` grandchild on the
  runner. Only reachable on the already-failing timeout path.
- `min dash`'s `attach_and_resume` does not arm `TerminalUnwind`, so the dash attach
  path lacks the belt-and-braces this PR adds to `min session attach`. Not a regression,
  but it is the other place a user attaches interactively.
- `notify_remote_teardown` computes `unwind_codes()` unconditionally, allocating a fresh
  parser even when `self.remote` is `None`. Forced by the borrow checker; an
  `if self.remote.is_some()` pre-check would avoid it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal cleanup when interactive sessions exit unexpectedly or are interrupted.
  - Restores terminal modes such as mouse input, alternate screen, cursor visibility, bracketed paste, and application key modes.
  - Preserves and propagates the session’s exit status, including signal-based termination codes.
  - Prevents unnecessary terminal resets during normal daemon-managed exits.


crates/minimal/src/attach.rs
```diff
@@ -12,11 +12,76 @@
...
+/// expects, whatever the session left it in: mouse reporting and its encodings
+/// off, bracketed paste off, application cursor keys and keypad off, off the
+/// alternate screen, cursor visible, drawing attributes and focus reporting
+/// reset.
+///
+/// The same set `minimald` computes per session as its unwind codes, minus the
+/// "only if currently set" narrowing — a client has no screen model to diff
+/// against, and every sequence here is a no-op against a terminal that was
+/// never in that mode.
+const TERMINAL_UNWIND: &[u8] = concat!(
```
twitchyliquid64: All of this already exists in the session host, grep for unwind codes

twitchyliquid64:
Yeah if the plan is to replace the session host unwind codes we should move/delete that code in this PR also, duplicating it likely isn’t ideal

crates/minimal/src/attach.rs
norrietaylor: Right that it shouldn't live in two places — single-sourced in `sessions` now (`1d3bb332`), both sides reference the same constants.

But they're not equivalent, so I'd keep the session-host version:

- `unwind_codes()` is state-aware — diffs live vs clean via `input_mode_diff`, emits `?1049l`/`?25h` only when actually set.
- The client constant is blind; it has no screen model to diff against.
- Deleting the daemon side would make the common path stop narrowing. Strictly worse.

Your comment did surface a real bug in mine, though: `?1049l` isn't the no-op I claimed. On xterm-family, rmcup-from-the-normal-buffer does a DECRC-style cursor restore — so after a clean detach the client's extra write can jump the cursor and redraw the prompt over existing content. A fix about cosmetic damage causing cosmetic damage.

So the fallback now fires only when ssh dies abnormally (255 or a signal):

- Clean teardown → daemon spoke, client stays quiet, no duplicate write.
- Broken transport → daemon never spoke, blind write earns its place.

Cross-PR note: I'd justified the fallback partly with "a daemon predating #1210 sends none." #1260 gates `cmd_attach`, and `session_via_ssh` sits downstream — once it lands, a CLI can't attach across builds at all. That half is void; dropped-transport is the only surviving case. Comment updated.

twitchyliquid64 (approved):

PR #1262, opened 2026-08-21
fix(mfile,graph,mctx): expand patch homes explicitly, drop credentials (#1204)

### LINKED ISSUE #1204: diag: min task run fails inside the microVM because a package-declared file mapping ~/.claude.json (from claud

**Bundle:** `minimal-diag-20260811T195202Z.tar.zst`
**Reporter context:** ➜  minimal git:(main) ✗ min task run fmt-check
Applying loadouts: default
Running task fmt-check in session task-fmt-check-95fc...
minimald: failed to spawn process: create mapped file I/O error at path /.claude.json: Read-only file system (os error 30)

---

# Diagnosis

## Verdict
`min task run` fails inside the microVM because a package-declared file mapping `~/.claude.json` (from `claude-code`'s `env_file_mappings`) is tilde-expanded against the *daemon process's* `HOME`, which is `/` for guest `minimald` (PID 1). Sandbox setup then tries to create `/.claude.json` on the guest's read-only ext4 rootfs (`/dev/vda / ext4 ro`) and dies with `EROFS`. Nothing is wrong with the user's machine, the VM, the socket path, or their config.

## Confidence
High for the mechanism (error string, HOME value, mount flags, and the code path all line up). Medium on *which* declaration pulls `claude-code` into these two tasks' closures — the project's `minimal.toml` and the `default` loadout are not in the bundle. Seeing the project `minimal.toml`/`~/.config/minimal/loadouts/default.toml` and a `min task run --verbose` with the resolved package closure would raise it to full confidence.

## Skew check
- Host CLI (`manifest.json` `.version`): **0.5.1**
- Provider daemon `local-minvmd0` (`socket-probe.json` get_version): **0.5.1** / long_version 0.5.1, **stdlib 0.0.18**
- Guest daemon (`providers/local-minvmd0/guest/daemon-diag/meta.json`): **0.5.1**, long_version 0.5.1, **stdlib 0.0.18**
- Binaries installed at `1786477621` = 2026-08-11T19:47:01Z (`state/listing.txt` `installed` marker); `minvmd run` pid 19151 etime 04:40 and VM `started_at=1786477643` = 19:47:23Z — i.e. **started 22 s after** the current binaries were installed; guest daemon uptime 281 s matches. No process predates its binary.
- All four identities agree; nothing in the failure (an EROFS from a filesystem mapping) is a protocol/version symptom.

**skew ruled out**

## Evidence
- Failing spawn, twice, at the same point in the task lifecycle:
  - `providers/local-minvmd0/guest/daemon-diag/logs/minimald.log.2026-08-11:128` — `{"channel_id":"7","error":"create mapped file I/O error at path /.claude.json: Read-only file system (os error 30)","message":"exec: failed to spawn next process in sequence; reporting exit 1"}` at `2026-08-11T19:51:41.680805Z` (session `task-fmt-check-95fc`)
  - same error at `19:49:31.330451Z` for session `task-shell-ff63` (`boot.log:120`).
- Guest `HOME` is `/`: `providers/local-minvmd0/guest/daemon-diag/env.json` → `{"HOME": "/", "PATH": ...}`; also echoed in the guest console `logs/kmsg.txt:279  HOME=/`. minimald is PID 1 in the VM (`meta.json`: `"pid": 1, "in_microvm": true`).
- Guest root filesystem is read-only: `providers/local-minvmd0/guest/daemon-diag/disk.json` → `"/dev/vda / ext4 ro,relatime 0 0"`, with the writable volume only at `/dev/vdb /var/lib/minimal ext4 rw,noatime`. `boot.log:50` — `switched to upstream rootfs (pivot_root) device="/dev/vda"`.
- Mechanics in source:
  - `crates/mfile/src/lib.rs` (`impl From<EnvPatches> for Vec<common::FsMapping>`): `let h = home_dir().unwrap(); ... if let Some(stripped) = path.strip_prefix("~/") { h.join(stripped) }` — expands `~/` against the *current process* home, with `create_if_missing: true`.
  - `crates/sandbox2/src/config.rs:605-616` — for a missing mapped file it does `fs::write(&m.host_path, "{}")` and maps the error to `Error::IO("create mapped file", …)`, exactly the logged string.
  - `crates/mctx/src/env.rs:615` — `.with_wd(args.cwd.clone(), false, patch.into())` is the only `WdSetup::BoundDir` construction, i.e. the task-exec path is the one that consumes `EnvPatches` (the session path at `crates/minimald/src/env.rs:375` has that call commented out).
  - The mapping's origin: `crates/graph/src/env_setup.rs:133,169` shows `env_file_mappings = [{ read_only = false, path = "~/.claude.json", class = 'Credential }]` surviving into `SetupForPackages.fs_mappings`, while the *session* composer deliberately drops `class = 'Credential` entries (`crates/minimald/src/sessions/composables.rs:681` `credential_class_fs_mappings_are_filtered_out`) — hence `fs_mappings=[]`, `daemon_patches=0` in the ConfigureLoadout logs while the task-exec path still carries it.
- The failure is at spawn, after everything else succeeded: `ConfigureLoadout` `outcome="ok" duration_ms=152`, `FinalizeSession ok`, DNS `registered PTask hostname task-fmt-check-95fc…` at `19:51:41.523935Z`, then the EROFS 157 ms later and immediate `DestroySession`.
- Not a connectivity/skew/hang problem: `socket-probe.json` is green at every stage (`stat/connect/handshake/get_version` all ok), `minvmd_alive: true`, 0 collector errors. `status.json` `"minimald_alive": false` is a host-side probe artifact (minimald lives in the guest, per `logs/PROVENANCE.txt`), and the guest daemon plainly served RPCs throughout.
- Counter-evidence checked: `task-python-cbb6` (19:50:21→19:50:37) ran through the identical session pipeline **without** the error and lived 3.4 s, so the fault is not "all tasks are broken" — it is specific to tasks whose package closure includes `claude-code` (the `cozy` loadout lists `"claude-code"` in `config/loadouts/cozy.toml.redacted`). The `own-IP/CGNAT` Tailscale warnings and the `check` FAIL (`guest bundle … missing errors.json`) are unrelated to this failure.

## Failure chain
1. 19:47:01Z — current 0.5.1 binaries installed; 19:47:23Z — `minvmd` boots the microVM, guest `minimald` starts as PID 1 with `HOME=/`, pivots to the read-only `/dev/vda` rootfs, writable state only at `/var/lib/minimal`.
2. 19:51:29.208Z — `min task run fmt-check` → `CreateSession` `task-fmt-check-95fc`; workspace upload (1.21 GB) completes 19:51:41.257Z.
3. 19:51:41.409Z — loadout composition succeeds (`daemon_patches=0`: the credential-class `~/.claude.json` mapping is filtered on *this* path); `SubmitVerdict`, `FinalizeSession`, DNS registration all ok by 19:51:41.524Z.
4. 19:51:41.68Z — task exec builds a `WdSetup::BoundDir` sandbox from `SetupForPackages.fs_mappings`, which *does* include `~/.claude.json`; `From<EnvPatches>` expands `~` via `home_dir()` = `/` → host_path `/.claude.json`; `create_if_missing` tries `fs::write("/.claude.json", "{}")` on the ro rootfs → `EROFS (os error 30)`.
5. Spawn fails, exit 1 is reported to the client (the message the user saw), the session is destroyed at 19:51:41.806Z. Same sequence had already happened at 19:49:31 for `task-shell-ff63`.
6. 19:52:02Z — `min bug` bundle collected.

## Recommended next steps
- Minimal team (this is a product bug, two places):
  1. Stop expanding `~` against the ambient process home in `impl From<EnvPatches> for Vec<common::FsMapping>` (`crates/mfile/src/lib.rs`). Thread the *sandbox/session* home (e.g. `sandbox2::SESSION_HOME` / `EnvArgs::home`, cf. `crates/minimald/src/env.rs:60` `HOME_ROOT`) into the conversion, and fail loudly if a mapping would resolve outside a writable tree instead of `fs::write`-ing into `/`.
  2. Apply the credential-class filter consistently: `crates/graph/src/env_setup.rs` keeps `class = 'Credential` file mappings that `crates/minimald/src/sessions/composables.rs` explicitly drops. The task-exec path (`crates/mctx/src/env.rs:615`) should use the same filtered set, so guest task runs never try to mount host Claude credentials at all.
  3. Improve the error: `create mapped file … /.claude.json` should name the contributing package and the unexpanded `~/.claude.json` so the user knows the culprit is `claude-code`.
- User workaround right now: remove/avoid `claude-code` in the package closure used by these tasks (drop it from the `default`/`cozy` loadout you apply to `min task run`, or from the task's `packages`), then re-run `min task run fmt-check`. Alternatively run the task with the loadout disabled to confirm the mapping is the trigger.
- If you want it confirmed before changing anything: `min task run fmt-check` with the loadout that omits `claude-code` should succeed while `task-python-*`-style tasks already do.

## Caveats
- The bundle does not contain the project's `minimal.toml` or the `default` loadout file (only `config/loadouts/cozy.toml.redacted`), so I infer `claude-code` enters the failing tasks' closures from the applied loadout; I cannot show the exact declaration. Only `cozy.toml` proves `claude-code` is in play somewhere in this user's setup.
- The bundle logs the fs mapping only as `fs_mappings=[]` on the session path; the task-exec path does not log its resolved mapping list, so the `~ → /` expansion is established by code reading (`mfile`/`sandbox2`) plus `HOME=/`, not by a log line printing the pre-expansion path.
- I cannot rule out that a *host-side* `min task run` (outside the VM) works fine for the same task — the bug as reconstructed requires the daemon's `HOME` to be `/`, which is a guest-only condition; on macOS the same mapping would resolve to `/Users/norrie/.claude.json` and silently be created.
- Guest-bundle `errors.json` is missing (`check` FAIL), so an unreported guest collector error cannot be fully excluded, though `manifest`/`errors` show none.

norrietaylor:
Closes #1204.

Two defects on the sandbox environment path.

**1. `~` was expanded against the ambient home.** `impl From<EnvPatches> for
Vec<common::FsMapping>` called `home_dir().unwrap()`. In the guest `minimald` runs as
pid 1 with `HOME=/`, so `~/.claude.json` became `/.claude.json` on a read-only ext4
root — EROFS at session create. Homes are now expanded **explicitly** from a
caller-supplied value, and a `/`-rooted home is refused up front rather than silently
producing junk paths at the filesystem root.

**2. The session and task-exec paths disagreed about credentials.**
`composables.rs` dropped `class = 'Credential` mappings while `graph/src/env_setup.rs`
kept them. The two paths now agree.

### Reviewer notes — please read before merging
- **This converts the symptom, it does not fully fix the guest case.** The only
  production caller still passes an ambient home (`std::env::home_dir()` in
  `crates/mctx/src/env.rs`), because `EnvArgs` has no `home` field to plumb one
  through. In the guest, a surviving `~/`-rooted State-class mapping now hard-errors
  with a clear pre-flight refusal instead of EROFS-at-create. Better, but the mapping
  still does not work; plumbing a real home to `minimald/src/exec.rs` is follow-up.
- **Dropping `class = 'Credential` is a functional removal on host runs too**, not just
  in the guest. A `mip run` task that previously received `~/.claude.json` (or an
  upstream package's `~/.aws/credentials`) now runs without it, signalled only by a
  `tracing::warn!`. Expect "not logged in" from tools with no error from minimal itself.
  Deliberate, documented under `TODO(secrets)`, and consistent with the session path —
  but it wants a release note.
- The `/`-home guard is string-exact (`s != "/"`), so `HOME=//` and `HOME=/.` slip past
  it and reproduce the original `/.claude.json`. Improbable, but comparing normalized
  components would be robust.
- `HOME=/` on a *writable* root (minimald as root in a container) regresses from
  "works" to "errors". Arguably correct — dotfiles at `/` are junk — but it is the one
  configuration that gets stricter.
- The renamed test `fs_mappings_drop_credential_class` **tightens** its assertion to the
  new narrower behaviour; it is not a weakened gate.

> [!NOTE]
> ### Expand `~/` patch homes explicitly via `EnvPatches::to_fs_mappings` and drop `Credential`-class fs mappings
> - Replaces the implicit `From<EnvPatches> for Vec<FsMapping>` with `EnvPatches::to_fs_mappings(&self, home: Option<&Path>)`, which requires the caller to pass a home directory and returns a structured `PatchHomeError` when `~/` cannot expand to a usable home (e.g. `/` or unset) instead of panicking or mapping to `/`
> - `env_setup::read_mapping` filters out fs mappings with `class = 'Credential` (`CREDENTIAL_CLASS_TAG`) and logs a warning per dropped entry; `SetupForPackages::build` now skips these entries
> - Adds `SetupForPackages::fs_mapping_packages: HashMap<String, String>` to record which package declared each surviving mapping; `mctx::env` uses it to attribute sandbox file-creation IO errors back to the declaring package or task and the original `~/` path
> - `graph::CREDENTIAL_CLASS_TAG` is now a public re-export so `minimald` and other crates import it instead of redefining
> - Risk: `From<EnvPatches> for Vec<FsMapping>` is removed; any out-of-tree caller relying on that impl will fail to compile. `~/` paths now error instead of silently mapping to `/` when home is unusable — check `env.rs` build sites in [crates/mctx/src/env.rs](https://github.com/gominimal/minimal/pull/1262/files#diff-b7da06c3b65c0833a934029f47dea9d800eba35146808b4f255d22c88f753995) and [crates/minimald/src/env.rs](https://github.com/gominimal/minimal/pull/1262/files#diff-ece548260b2c5052569423263e9b54827b352ed496728bed2f1bf4c52c251dc1)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 80b867a.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Filesystem mappings now retain package information for clearer diagnostics.
  - `~/` paths resolve against the configured ambient or session home directory.
  - Environment setup errors identify the package or task associated with failed mappings.
  - Environment and patch mappings now use stable ordering for consistent results.

- **Bug Fixes**
  - Credential-related filesystem mappings are filtered with warnings.
  - Invalid or unclassifiable mappings now produce clearer errors.
  - Home-directory paths are validated more reliably, while valid absolute paths continue to work.

### DIFF
--- a/crates/graph/src/env_setup.rs
+++ b/crates/graph/src/env_setup.rs
@@ -3,6 +3,7 @@ use std::{
     path::PathBuf,
 };
 
+use crate::env_mapping::{EnvFsMapping, EnvMappingClass, EnvMappingKind};
 use crate::{BuildSpecRef, Graph};
 use decode::AttrValue;
 use mfile::EnvPatches;
@@ -30,64 +31,6 @@ pub struct SetupForPackages {
     pub fs_mapping_packages: HashMap<String, String>,
 }
 
-/// The string form of the nickel enum tag `'Credential` as it appears after
-/// `decode`'s AttrValue conversion. Compared as a plain string at runtime, so
-/// if `decode::AttrValue`'s enum-tag rendering ever changes (angle-brackets,
-/// hash prefix, etc.) the checks here and in `minimald`'s session-composition
-/// extractor silently stop matching, and credential mappings flow through
-/// unfiltered.
-///
-/// The regression guards are `fs_mappings_drop_credential_class` in this
-/// module and `credential_class_fs_mappings_are_filtered_out` in
-/// `minimald::sessions::composables`; if the rendering changes, both fail and
-/// this constant is where to update.
-// TODO(schema-stability): keep in sync with `decode::AttrValue::EnumTag`
-// rendering.
-pub const CREDENTIAL_CLASS_TAG: &str = "Credential";
-
-/// Reads one `env_dir_mappings` / `env_file_mappings` entry into its declared
-/// path and patch setting, dropping `class = 'Credential` entries.
-///
-/// TODO(secrets): credential-class mappings are dropped rather than routed
-/// through a separate secrets channel, matching what the session-composition
-/// path already does in `minimald`'s `extract_fs_mappings`. Until the secrets
-/// strategy lands they simply aren't seen by the sandbox — which is also what
-/// keeps a task running inside the guest from trying to mount host
-/// credentials that aren't there (#1204).
-fn read_mapping(
-    package: &str,
-    attr: &'static str,
-    entry: &AttrValue,
-) -> Option<(String, mfile::PatchSetting)> {
-    use mfile::PatchSetting;
-
-    let entry = entry.as_map().unwrap();
-    let path = entry.get("path").unwrap().as_string().unwrap().clone();
-    if entry
-        .get("class")
-        .and_then(|c| c.as_string())
-        .map(String::as_str)
-        == Some(CREDENTIAL_CLASS_TAG)
-    {
-        tracing::warn!(
-            package = %package,
-            attr = %attr,
-            path = %path,
-            "dropping Credential-class fs mapping; the secrets strategy is deferred, so \
-             credentials do not reach the sandbox through package attributes",
-        );
-        return None;
-    }
-    Some((
-        path,
-        if *entry.get("read_only").unwrap().as_bool().unwrap() {
-            PatchSetting::ReadOnly
-        } else {
-            PatchSetting::ReadWrite
-        },
-    ))
-}
-
 impl SetupForPackages {
     /// Computes environment settings that need to be applied to an environment containing the
     /// given packages.
@@ -108,25 +51,36 @@ impl SetupForPackages {
 
         for dep in i.into_iter() {
             let b = g.get(dep).unwrap();
-            if let Some(dirs) = b.attrs.get("env_dir_mappings") {
-                for mapping in dirs.as_list().unwrap() {
-                    if let Some((path, setting)) =
-                        read_mapping(&b.name, "env_dir_mappings", mapping)
-                    {
-                        fs_mapping_packages.insert(path.clone(), b.name.clone());
-                        patch.dir.insert(path, setting);
-                    }
+            for mapping in EnvFsMapping::decode_all(b)? {
+                // TODO(secrets): credential-class mappings are dropped rather
+                // than routed through a separate secrets channel, matching
+                // what the session-composition path already does in
+                // `minimald`'s `extract_fs_mappings`. Until the secrets
+                // strategy lands they simply aren't seen by the sandbox —
+                // which is also what keeps a task running inside the guest
+                // from trying to mount host credentials that aren't there
+                // (#1204).
+                if mapping.class == EnvMappingClass::Credential {
+                    tracing::warn!(
+                        package = %b.name,
+                        attr = %mapping.kind.attr(),
+                        path = %mapping.path,
+                        "dropping Credential-class fs mapping; the secrets strategy is \
+                         deferred, so credentials do not reach the sandbox through package \
+                         attributes",
+                    );
+                    continue;
                 }
-            }
-            if let Some(files) = b.attrs.get("env_file_mappings") {
-                for mapping in files.as_list().unwrap() {
-                    if let Some((path, setting)) =
-                        read_mapping(&b.name, "env_file_mappings", mapping)
-                    {
-                        fs_mapping_packages.insert(path.clone(), b.name.clone());
-                        patch.file.insert(path, setting);
-                    }
-                }
+                let setting = if mapping.read_only {
+                    mfile::PatchSetting::ReadOnly
+                } else {
+                    mfile::PatchSetting::ReadWrite
+                };
+                fs_mapping_packages.insert(mapping.path.clone(), b.name.clone());
+                match mapping.kind {
+                    EnvMappingKind::Dir => patch.dir.insert(mapping.path, setting),
+                    EnvMappingKind::File => patch.file.insert(mapping.path, setting),
+                };
             }
             if let Some(wiring) = b.attrs.get("env_state_wiring") {
                 let mut apply_wiring = |entry: &AttrValue| -> Result<(), std::io::Error> {
@@ -239,6 +193,51 @@ mod tests {
                 )]),
             }
         )
+    }
+
+    /// An entry the typed decode can't classify stops the whole setup with
+    /// an error naming the package, rather than reaching the sandbox as an
+    /// unclassified — and therefore unfiltered — mapping. This is what the
+    /// old `class` string comparison could not do: it answered "not
+    /// `Credential`" for anything it didn't recognise and let the mapping
+    /// through.
+    #[test]
+    fn an_unclassifiable_mapping_fails_the_whole_setup() {
+        let layer = Layer::new_for_test(
+            indoc! {
+                "
+            let {BuildSpec, ..} = import \"minimal.ncl\" in
+
+            let
+                b1 = {
+                    name = \"b1\",
+                    build_deps = [],
+                    cmd = \"\",
+                    attrs = {
+                        env_file_mappings = [{ read_only = false, path = \"~/.claude.json\" }],
+                    },
+                } | BuildSpec,
+            in
+            [b1]
+            "
+            }
+            .to_string(),
+        )
+        // Nickel's own `class | [| 'Credential, 'State |]` contract does not
+        // catch this: a record contract's undefined field is lazy, so an
+        // entry that simply omits `class` evaluates cleanly and arrives here
+        // unclassified. That is precisely the case the old string comparison
+        // waved through.
+        .expect("nickel accepts an entry that omits `class`");
+        let g = Graph::new().ingest(layer).unwrap();
+
+        let err = SetupForPackages::build(&g, [g.by_name("b1").unwrap()])
+            .expect_err("an entry with no class must not decode");
+        assert_eq!(err.kind(), std::io::ErrorKind::InvalidData);
+        assert!(
+            err.to_string().contains("b1") && err.to_string().contains("class"),
+            "the error must name the package and the offending field, got {err}"
+        );
     }
 
     #[test]
--- a/crates/graph/src/env_setup.rs
+++ b/crates/graph/src/env_setup.rs
@@ -21,6 +21,71 @@ pub struct SetupForPackages {
     pub needs_internet: bool,
     /// The filesystem mappings accumulated from `env_file_mappings` and `env_dir_mappings` attrs.
     pub fs_mappings: EnvPatches,
+    /// Which package contributed each entry of `fs_mappings`, keyed by the
+    /// path exactly as the package declared it (`~/`-rooted paths are not
+    /// expanded here). Last writer wins, matching `fs_mappings` itself.
+    ///
+    /// Consumed when a mapping can't be honoured, so the error can name the
+    /// package to go and look at rather than only the path it landed on.
+    pub fs_mapping_packages: HashMap<String, String>,
+}
+
+/// The string form of the nickel enum tag `'Credential` as it appears after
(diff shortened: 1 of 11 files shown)


crates/graph/src/env_setup.rs
```diff
@@ -21,6 +21,71 @@ pub struct SetupForPackages {
...
+/// extractor silently stop matching, and credential mappings flow through
+/// unfiltered.
+///
+/// The regression guards are `fs_mappings_drop_credential_class` in this
+/// module and `credential_class_fs_mappings_are_filtered_out` in
+/// `minimald::sessions::composables`; if the rendering changes, both fail and
+/// this constant is where to update.
+// TODO(schema-stability): keep in sync with `decode::AttrValue::EnumTag`
+// rendering.
+pub const CREDENTIAL_CLASS_TAG: &str = "Credential";
```
twitchyliquid64: Instead of clumsily working with the `AttrValue` form, it might be better to explictly parse/decode the env mappings attributes into rust structs/enums that can be used throughout.

twitchyliquid64:
> This converts the symptom, it does not fully fix the guest case. The only production caller still passes an ambient home (std::env::home_dir() in crates/mctx/src/env.rs), because EnvArgs has no home field to plumb one through.

Im fairly sure the minimald equivalent has the session homedir as an env arg or equivalent. Not sure if `mctx` path matters for anything but mip (it might matter for task exec in a minimald session)

twitchyliquid64 (commented):
Otherwise LGTM

crates/graph/src/env_setup.rs
norrietaylor: `1d386727`. New `graph::env_mapping`: `EnvFsMapping` + `EnvMappingClass` enum, typed errors. `CREDENTIAL_CLASS_TAG` deleted; both extractors share it.

Your instinct beat the tidiness argument — it turned up a live fail-open:

- nickel's `class | [| 'Credential, 'State |]` doesn't catch a *missing* `class` (record contract fields are lazy)
- so a classless mapping evaluated clean, reached the filter, and was waved through as non-credential
- now a hard error naming the package

Checked all 8 upstream `pkgs` files — every entry declares a class, so no blast radius.

norrietaylor:
**Homedir** — right, and there are *two* minimald call sites: `exec.rs:191` and `env.rs:1153`.

- Added `mctx::PatchHome { Ambient, Session(DaemonAbsPath) }` on `EnvArgs`; every call site now states which home it means
- `SessionChannel` already carried `home` — no plumbing needed
- `mip` + nested in-sandbox stay `Ambient`, explicitly
- `/`-guard compares normalized components now, so `HOME=//` and `HOME=/.` no longer rebuild `/.claude.json`
- Also needed `sandbox2::Config::with_home` — without it the mapping landed at the session home while `$HOME` stayed `/`, making the fix cosmetic

norrietaylor:
**Force-pushed to split the third concern out** (`1d386727` → `ebbbec67`), per gominimal/inbox#549.

Evan's ordered-carriers comment — `EnvPatches.dir`/`.file` and `SetupForPackages.env_vars` / `.state_dirs` / `.fs_mapping_packages` as `BTreeMap`/`BTreeSet` — is now gominimal/inbox#552 and is off this branch. It has to land atomically across `mfile`, `decode`, `graph`, `mctx` and `minimald` (the type change doesn't compile crate-by-crate) and it retypes the same functions the fixes here touch, so bundled it read as a five-crate mechanical sweep sitting on top of the behavioural change.

What's left is the two behavioural fixes: session home plumbing (`mctx::PatchHome`, `sandbox2::Config::with_home`, the normalized-component `/`-home guard) and the typed `graph::env_mapping` decode replacing the stringly-typed class comparison.

Nothing here depends on the split-out change. `to_fs_mappings` still returns a deterministic sequence — directories then files, each sorted by declared path on the way out — so mapping order and the entry an error names are stable already. `crates/decode/src/lib.rs` drops out of the diff entirely; the PR goes from 12 files / +1370 −208 to 11 / +1205 −180.

`just ci`'s gates pass on `ebbbec67` (fmt-check, clippy, workspace tests, doctests; `cargo-deny` and `cargo-nextest` weren't available in the environment, so the test tier ran as `cargo test --workspace` — all green, no dependency changes for deny to weigh).

The pre-split commit `1d386727` stays reachable from this force-push timeline entry, and inbox#552 points at it — the two sequence-asserting tests it added are recoverable from there rather than needing to be rewritten.

twitchyliquid64 (approved):

PR #1308, opened 2026-08-28
fix(exec)!: tag exec-channel requests instead of sniffing the string

twitchyliquid64:
ssh's exec request carries one string (RFC 4254 §5.2), and the daemon decided what it meant by looking at it: anything starting with `min ` was one of its own commands, everything else was a shell command for the session. That conflated three intents in one untagged string, and both failure modes were real.

An argv could not be sent. ssh joins its trailing arguments with single spaces and the far side reshells the result, so
`min session exec s sh -c 'echo A B C'` arrived as `sh -c echo A B C` and `A` became sh's $0 — it never printed. Quoting the words on the client fixed the splitting but broke the sniffing, because both read the same bytes.

A session command was stolen. `min --version` names the session's own `min`, but the daemon claimed the prefix and refused it, so that binary was simply unreachable through exec.

Requests are now tagged, in a vocabulary both sides share (`minimald-rpc::exec`): min://shell, min://argv, min://task/run, min://package/build and min://check. Argv words travel as JSON — NUL, the one delimiter that could never collide with the payload, is exactly the one execve cannot carry — and are exec'd with no shell in between. Anything untagged belongs to the session, which is the rule that makes the hijack impossible.

Adds `min session run <session> <task>`, which names the task form explicitly and restores running a declared task against a session you already have. A task may now share a name with a daemon subcommand or a program on PATH.

BREAKING CHANGE: the exec-channel wire contract changed, so a client and daemon must be built together — the existing version gate already refuses a skewed pair. `min session exec <s> 'min check'` and `min session exec <s> min run <task>` no longer route to the daemon; they run in the session. The daemon forms are reached by `min session run` and `min task run`.

Fixes: gominimal/inbox#558

Claude-Session: https://claude.ai/code/session_01Si8R82wVqVvvWiKVJAmEJX

## Checklist

- [ ] `BREAKING CHANGE:` footer present if this is a breaking change

## Summary by CodeRabbit

- **New Features**
  - Added `min session run` to execute named tasks in existing sessions.
  - Enhanced `min session exec` with shell commands and literal argument execution.
  - Improved handling of commands containing spaces and special characters.
  - Commands resembling daemon commands now run correctly within sessions.

- **Bug Fixes**
  - Improved task output and exit-code reporting.
  - Invalid execution requests now provide clear errors and exit status.

- **Documentation**
  - Updated CLI, session, task, and architecture documentation for new commands and execution behavior.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,7 +46,7 @@ The CLI reference overview is [docs/reference/cli.md](docs/reference/cli.md).
 | `minimal-client` | SSH client transport to `minimald` over the UDS bridge, shared by `min` and the TUI. |
 | `minimal-tui` | `min dash`: the session-manager TUI (ratatui/crossterm, Elm-style loop). |
 | `minimald` | The session daemon: an SSH server hosting sessions and task/sandbox executions. |
-| `minimald-rpc` | Wire contract for `minimald`'s oneshot SSH RPCs. |
+| `minimald-rpc` | Wire contract for `minimald`'s oneshot SSH RPCs and the exec channel's request vocabulary. |
 | `minvmd` | Host daemon that boots Linux microVMs via libkrun and bridges host UDS to in-VM vsock. |
 | `mip` | The Minimal package/build CLI. |
 | `mlog` | JSON file-log layer both `minimald` and `minvmd` write through; one definition of the on-disk log format. |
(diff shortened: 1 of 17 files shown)


norrietaylor (approved):

norrietaylor:
Approved but the `minimal-review` comment looks like it needs a check.

PR #1313, opened 2026-08-28
feat(packaging): CI-owned AUR, Homebrew, and deb/rpm/apk distribution

0chroma:
Replaces the hand-maintained AUR package with CI-owned publishing and adds Homebrew, deb, rpm, and apk distribution. The monorepo holds every packaging template. The satellite repos (the AUR repo, the brew tap) become generated output that publish scripts clone, render, and push. Plan reference: expand packaging past `minimal-bin` on the AUR, per the distribution plan.

## What changed since the first review round

The review round found defects that all trace to two root causes: publishing logic inline in the frozen workflows (untested by construction) and templates that nothing ever rendered. The second push fixes the defects and the causes.

- **Homebrew publishes on the stable promote, not on tag push.** Homebrew has no channels, so a tag push shipped every release (an RC tag included) straight to every user with no opt-out and no gate. `publish-brew` now lives in `promote.yml` next to `publish-aur`, behind the same `target == stable && !dry_run` gate, and refuses prereleases outright.
- **The promote sha resolves to a released semver through a tested script.** `git describe --exact-match` only matched when the promoted sha IS the tagged commit. The nightly that gets promoted sits after the tag commit, so the AUR publish was a silent permanent skip (and with `git` missing from the `archlinux:base-devel` container, a swallowed exit 127). `scripts/resolve-release-version.sh` uses the tag on the sha itself, else the newest released `v*` tag reachable from it. Prereleases never resolve and never fall through to an older release, which would publish a downgrade.
- **The release gate's logic moved into tested scripts.** `resolve-release-version.sh`, `assert-release-version.sh`, `latest-staged-version.sh`, and `backfill-version-row.sh`, each with a `scripts/*_test.sh` harness. The pkg upload goes through `stage-release.sh --pkg-dir/--pkg-only` instead of an inline `gcloud storage cp`. The default promote target filters to commit rows, so the new `versions/<semver>/` rows cannot hijack it and fail the run after the approval gate.
- **The AUR package now ships minvmd**, matching the deb/rpm/apk file list. VM-backed Linux sessions need it on `PATH`. Every staged row carries the static-musl build, so the package stays self-contained.
- **The nfpm packaging verifies before it runs.** Downloaded artifacts are checked against the staged `components` manifest (the SHA-256s `stage-release.sh` wrote and the installer trusts) before anything chmods, executes, or packages them.
- **The Homebrew formula installs correctly**: `bin.install ... => "min"` (the previous form created a `bin/min/` directory and installed no executable), stage blocks use relative paths, and the trimmed libkrun dylib ships into the formula's own `lib/`, which is the only place minvmd's `@loader_path/../lib` rpath can reach.
- **The renderer refuses to write a half-stamped file**: set-but-empty variables fail (printenv exits 0 for them), bash 5.2 `patsub_replacement` is disabled so a literal `&` in MAINTAINER cannot re-inject its token, and any surviving `@@` fails the run.
- `package.version` falls back to `0.5.4` (next unreleased) instead of the already-shipped `0.5.3`, the exact wrong-tarball-version failure `assert-version` exists to catch. `dist-build`/`pkg-nfpm` are `[linux]`-scoped with `quote()`d arguments.

New harnesses run under `just test-shell` locally and through the workspace suite via `crates/common/tests/shell_harnesses.rs`, which discovers `scripts/*_test.sh` by convention (the same extension point `shell_lint.rs` uses, since the frozen workflows cannot schedule shell tests).

## Verification

- `just ci`: fmt, clippy, cargo-deny, and all 1849 workspace tests pass. `just test-ignored` fails on one test, `mctx tests::task_env`, which fails on this workstation on `origin/main` too. Running `just ci` here needs `ASDF_SHELLCHECK_VERSION=0.11.0` (the asdf shim has no version pin).
- `just test-shell`: shellcheck on every harness plus 50 assertions across the six new harnesses: semver resolution against a fixture git repo, version assertion against fixture manifests, the staged-version lookup against a stubbed `gcloud`, the renderer, and both publishers driven end to end against `file://` fixture remotes (no network, no credentials).
- `scripts/lint-shell.sh` passes on all 42 scripts.
- Every publish-script dry run renders its template and asserts the output carries no unrendered tokens and only real digests.
- nfpm smoke-tested for all six format and arch pairs (deb control, postinst, symlink, apk `.post-install`).

`just pkg-nfpm 0.5.3` will fail until the semver row exists for that release: `just backfill-version-row <sha> 0.5.3` builds it from the sha row.

## Before the first publish (user actions)

- AUR bot account plus `secrets.AUR_SSH_PRIVATE_KEY`.
- Create `gominimal/homebrew-minimal` plus `secrets.BREW_TAP_TOKEN`.
- CODEOWNER approval for the workflow changes (`publish-brew` moved out of `release.yml`. `promote.yml` gained a job and two explicit `permissions` blocks).

## Not in this PR

- Repo-tree hosting for apt, dnf, and apk (reprepro, createrepo_c, apk index) lives in the infra repo. This PR only stages packages under `versions/<semver>/pkg/`.
- AUR source package and aports submission stay deferred.

## Summary by CodeRabbit

- **New Features**
  - Added Arch Linux binary packages with architecture-specific VM host binaries.
  - Added Apple Silicon Homebrew support, including shell completions and required runtime libraries.
  - Added Debian, RPM, and APK packages for supported architectures.
  - Added automatic AppArmor profile loading during Arch installation and upgrades.

- **Improvements**
  - Release publishing now validates versions and more reliably handles stable releases.
  - Updated the application version to 0.5.4.
  - Improved recovery and restaging of previously published release artifacts.
  - Added package smoke tests and broader release validation coverage.


packaging/homebrew/minimal.rb.tmpl
```diff
@@ -0,0 +1,79 @@
...
+
+    # minvmd boots the guest microVM (libkrun comes from the slp/krun
+    # dependency above). The guest payload (kernel/rootfs/initramfs) is not
+    # in the formula: minvmd resolves it from the minimal data dir
+    # ($XDG_DATA_HOME/minimal, else ~/.local/share/minimal) — which the
+    # curl|sh install fills — or from the MINVMD_KERNEL_PATH /
+    # MINVMD_ROOTFS_PATH / MINVMD_INITRAMFS overrides; the release also
+    # publishes those arm64 artifacts.
+    resource("minvmd").stage do |staged|
+      (bin/"minvmd").install staged.staging/"minvmd-macos-arm64"
```
norrietaylor: packaging/homebrew/minimal.rb.tmpl:51 — staged.staging/"..." calls / on a Mktemp, which defines no such method. brew install minimal dies with NoMethodError. Same shape at line 60. Mktemp#run already chdirs, so bin.install "minvmd-macos-arm64" => "minvmd" is the working form.

packaging/homebrew/minimal.rb.tmpl
```diff
@@ -0,0 +1,79 @@
...
+  resource "gvproxy" do
+    url "https://github.com/gominimal/minimal/releases/download/v@@VERSION@@/gvproxy-darwin-arm64"
+    sha256 "@@SHA_GVPROXY@@"
+  end
+
+  def install
+    # The release asset minimal-macos-arm64 is the `min` binary (the bin
+    # target of the `minimal` crate, named for the crate in the release);
+    # the installer installs it as `bin/min`.
+    (bin/"min").install "minimal-macos-arm64"
```
norrietaylor: packaging/homebrew/minimal.rb.tmpl:35 — (bin/"min").install "minimal-macos-arm64" creates a directory bin/min/ and moves the binary inside it. No min executable is ever installed; the ln_sf, generate_completions_from_executable, and test do blocks then all operate on a directory.

norrietaylor (commented):
Also flagged, below the cut: publish-brew has no continue-on-error and its preconditions (tap repo, BREW_TAP_TOKEN) are out of scope, so every tag release goes red until those land; nfpm packaging sits before the install.sh staging and the unstable flip, so a transient failure aborts a release mid-publish; publish-aur runs after the irreversible channel flip and is the only job with no permissions: block; no preremove/postremove, so /etc/apparmor.d/minimald is orphaned on uninstall; the epoch=1 comment's vercmp example is backwards; dist/ isn't gitignored and dist-build.sh does rm -rf on an unvalidated env knob; the completions incantation now exists in three drifted copies; package-nfpm.sh re-downloads ~257 MB it already has on disk.

scripts/publish-aur.sh
```diff
@@ -0,0 +1,186 @@
...
+export GIT_SSH_COMMAND="$ssh_base -o StrictHostKeyChecking=accept-new"
+
+git clone --depth 1 "$AUR_REPO_URL" "$workdir/aur" \
+    || die "clone of $AUR_REPO_URL failed (check AUR credentials; see the header)"
+
+"$RENDER" "$TEMPLATE" "$workdir/aur/PKGBUILD"
+cp "$INSTALL_HOOK" "$workdir/aur/minimal-bin.install"
+
+if command -v makepkg >/dev/null 2>&1; then
+    (cd "$workdir/aur" && makepkg --printsrcinfo > .SRCINFO) \
```
norrietaylor: scripts/publish-aur.sh:149 — makepkg --printsrcinfo refuses to run as root (the EUID == 0 guard fires before the printsrcinfo early-return; Arch declined to exempt it, FS#67158). Container jobs run as root, so this always dies — and the > .SRCINFO redirect has already truncated the file.

.github/workflows/promote.yml
```diff
@@ -261,8 +263,47 @@ jobs:
...
+            - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+              with:
+                  persist-credentials: false
+                  # Full history + tags: the semver resolution below matches
+                  # the promoted sha against the v* tags.
+                  fetch-depth: 0
+            - name: Install curl
+              # base-devel carries git and makepkg but not the curl CLI the
+              # script uses to fetch versioned artifacts from the public bucket.
+              run: pacman -Sy --noconfirm --needed curl
```
norrietaylor: .github/workflows/promote.yml:293 — the archlinux:base-devel container has neither git nor openssh (and already has curl, so the pacman -Sy curl step is a no-op). Without git, git describe exits 127, || true swallows it, semver= is empty, the publish step is skipped — and the job reports success. A silent permanent no-op on every stable promotion.

.github/workflows/promote.yml
```diff
@@ -261,8 +263,47 @@ jobs:
...
+                   fi
+
+    publish-aur:
+        # The AUR package tracks promoted semver releases (its PKGBUILD is
+        # keyed on versions/<semver>/ in the bucket), so only stable
+        # promotions publish, and only when the promoted sha is exactly a
+        # semver-tagged commit. A stable promotion of an untagged staged sha
+        # skips the publish with a notice.
+        needs: promote
+        if: ${{ inputs.target == 'stable' }}
```
norrietaylor: Blockign: promote.yml:279 — publish-aur is gated only on target == 'stable' and never checks inputs.dry_run. A rehearsal promotion pushes a real, irreversible commit to the AUR. The sibling job in release.yml gets this right.

.github/workflows/release.yml
```diff
@@ -906,6 +919,32 @@ jobs:
                     --artifacts-dir artifacts \
                     --version "$SHORT_SHA" \
                     --bucket gs://minimal-one
+            - name: Also stage under the semver tag
```
norrietaylor, line 911: release.yml:922/933 — staging under the semver tag after the sha row makes versions/<semver>/components the newest object, hijacking promote.yml's "latest staged" default resolution (SHA=0.5.4). Fails the hex guard after the human approval gate; with override_provenance it points stable at a non-commit.

.github/workflows/release.yml
```diff
@@ -906,6 +919,32 @@ jobs:
...
+              # renders) fetches from versions/<semver>/ — only tag pushes
+              # carry a semver. Same inert row layout as the sha-keyed upload
+              # above; the channel pointers still key off the sha row.
+              if: ${{ github.ref_type == 'tag' }}
+              env:
+                  SEMVER: ${{ github.ref_name }}
+              run: |
+                  scripts/stage-release.sh \
+                    --artifacts-dir artifacts \
+                    --version "${SEMVER#v}" \
```
norrietaylor, line 922: release.yml:933 — no already-shipped release has a versions/<semver>/ row (verified: versions/0.5.3/components → 404). So just pkg-nfpm 0.5.3, the exact command justfile:172 documents, fails; nothing backfills.

.github/workflows/release.yml
```diff
@@ -930,3 +969,27 @@ jobs:
...
+    publish-brew:
+        # The brew tap tracks the GitHub Release assets directly (Homebrew has
+        # no channel concept), so it publishes on tag push — unlike the AUR,
+        # which waits for the stable promote (promote.yml's publish-aur job).
+        # The tap repo must exist and BREW_TAP_TOKEN (a fine-grained PAT with
+        # write access to it) must be set (user actions).
+        runs-on: ubuntu-latest
+        timeout-minutes: 20
+        needs: [release]
+        if: ${{ github.ref_type == 'tag' && !inputs.dry_run }}
```
norrietaylor: release.yml:982 — publish-brew has no prerelease guard. The tag glob matches v0.6.0-rc1, and the RC overwrites the tap's only formula. Homebrew has no channels, so users can't opt out.

scripts/publish-aur.sh
```diff
@@ -0,0 +1,186 @@
...
+
+[ -n "${PKGVER:-}" ] || die "PKGVER is required (the promoted semver, without the v prefix)"
+
+# The PKGBUILD maintainer line.
+MAINTAINER="${MAINTAINER:-minimal <security@minimal.dev>}"
+export MAINTAINER
+case "$PKGVER" in
+    v*) die "PKGVER must not carry the v prefix: '$PKGVER' (use ${PKGVER#v})" ;;
+esac
+printf '%s\n' "$PKGVER" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-][0-9A-Za-z.-]+)?$' \
```
norrietaylor: scripts/publish-aur.sh:72 — the validator accepts a -rc1 tail but pacman forbids hyphens in pkgver; makepkg's lint rejects it with a message blaming the template.

Cargo.toml
```diff
@@ -49,7 +49,7 @@ package.rust-version = "1.95"
 # Shared base version for the workspace. This is the fallback the `version` crate
 # reports when no `v*` git tag is reachable; the real release version comes from
 # `git describe` at build time. See crates/version/build.rs.
-package.version = "0.5.0-rc1"
+package.version = "0.5.3"
```
norrietaylor: Cargo.toml:52 — package.version = 0.5.3 is an already-shipped version, so a no-.git build of any post-0.5.3 commit reports 0.5.3. That's the exact failure release.yml:125's new assert-version job says it exists to prevent; the fallback should be 0.5.4.

scripts/render-packaging.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# $(cat) drops trailing newlines; the sentinel x keeps the file byte-exact.
+content="$(cat "$template"; printf x)"
+content="${content%x}"
+
+# One pass over the distinct well-formed tokens; the leftover check below
+# catches the malformed rest.
+while IFS= read -r token; do
+    name="${token//@@/}"
+    if ! value="$(printenv "$name")"; then
```
norrietaylor: scripts/render-packaging.sh:52 — two holes in the "never write a half-stamped file" guarantee: printenv returns 0 for a set-but-empty variable (so a missing sha tool ships sha256 "" and the leftover scan can't see it — needs [ -n "$value" ]), and on bash ≥ 5.2 patsub_replacement expands a literal & in the replacement back to the match, so an & in MAINTAINER re-injects the token.

packaging/arch/PKGBUILD-bin.tmpl
```diff
@@ -0,0 +1,86 @@
...
+    install -Dm644 "${srcdir}/minimald.apparmor-tunable" \
+        "${pkgdir}/usr/share/minimal/apparmor/tunables/minimald"
+    install -Dm755 "${srcdir}/install-apparmor-profile.sh" \
+        "${pkgdir}/usr/share/minimal/apparmor/install-apparmor-profile.sh"
+
+    # Upstream ships no pre-packaged completions; the binary generates them.
+    # XDG overrides redirect its user-level target dirs into srcdir, and
+    # ZDOTDIR keeps its compinit-dump cleanup off the build host's ~.
+    chmod +x "${srcdir}/min"
+    XDG_DATA_HOME="${srcdir}/completions" \
```
norrietaylor, line 87: packaging/arch/PKGBUILD-bin.tmpl:74 — missing BASH_COMPLETION_USER_DIR, so min silently skips bash and the next install -Dm644 aborts package() with ENOENT in a clean chroot. dist-build.sh:145 sets it and documents why.

packaging/arch/PKGBUILD-bin.tmpl
```diff
@@ -0,0 +1,86 @@
...
+url="https://minimal.dev"
+license=("Apache-2.0")
+depends=(git)
+optdepends=(
+    "bash: shell integration and completions"
+    "zsh: shell integration and completions"
+    "fish: shell integration and completions"
+)
+install=minimal-bin.install
+_bucket="https://storage.googleapis.com/minimal-one"
```
norrietaylor: packaging/arch/PKGBUILD-bin.tmpl:28 — _bucket is hardcoded with no token, so MINIMAL_BUCKET_URL changes the checksums but not the stamped URLs.

scripts/publish-aur.sh
```diff
@@ -0,0 +1,186 @@
...
+cd "$workdir/aur"
+
+# The AUR server does not check commit identity; default it so a fresh CI
+# container (no global git config) still commits. An existing identity wins.
+git config user.name  >/dev/null || git config user.name "minimal-ci"
+git config user.email >/dev/null || git config user.email "minimal-ci@users.noreply.archlinux.com"
+
+git add -A
+
+if [ "$DRY_RUN" -eq 1 ]; then
```
norrietaylor, line 241: scripts/publish-aur.sh:164 — --dry-run is unreachable without production credentials, a network, and a clone. Nothing in the workspace ever renders these templates, which is why five separate template defects above reach review unexercised.

justfile
```diff
@@ -152,6 +152,27 @@ minimal-cli:
...
+# THE single build entrypoint for a shippable build: the four release binaries
+# for a target triple, built exactly the way release.yml's build jobs build
+# them (they call this same script), plus completions from the built `min`.
+# Per-package invocations (fat-LTO links serialize), FEATURES for the guest
+# minimald, and the fail-loud static-libkrun requirement for minvmd live in
+# scripts/dist-build.sh — read its header for the LIBKRUN_PREFIX /
+# COMPLETIONS_DIR / FEATURES env knobs.
+#
+# Build the shippable binaries + completions for a target triple (e.g. just dist-build x86_64-unknown-linux-musl).
+dist-build target:
```
norrietaylor, line 168: justfile:164 — dist-build/pkg-nfpm carry no [linux]/[macos] attribute, and dist-build.sh's host check compares architecture only, ignoring OS. On macOS both are offered: one skips the mandatory minvmd codesign and builds minimald natively; the other execs a musl ELF after a full fat-LTO build.

.github/workflows/promote.yml
```diff
@@ -261,8 +263,47 @@ jobs:
...
-                      -f event_type=reference-docs-promoted \
-                      -F "client_payload[sha]=${SHA}"
-                    echo "Dispatched reference-docs-promoted to gominimal/webapp for ${SHA}"
+                     gh api repos/gominimal/webapp/dispatches \
+                       -f event_type=reference-docs-promoted \
+                       -F "client_payload[sha]=${SHA}"
+                     echo "Dispatched reference-docs-promoted to gominimal/webapp for ${SHA}"
+                   fi
+
+    publish-aur:
```
norrietaylor, line 262: Blocking: Tagging before nightly permanently poisons the sha, so this job can never fire for it.

.github/workflows/promote.yml
```diff
@@ -261,8 +263,47 @@ jobs:
...
+        # The AUR package tracks promoted semver releases (its PKGBUILD is
+        # keyed on versions/<semver>/ in the bucket), so only stable
+        # promotions publish, and only when the promoted sha is exactly a
+        # semver-tagged commit. A stable promotion of an untagged staged sha
+        # skips the publish with a notice.
+        needs: promote
+        if: ${{ inputs.target == 'stable' }}
+        runs-on: ubuntu-latest
+        container: archlinux:base-devel
```
norrietaylor, line 275: This will cause a zizmor permissions regression. We keep this workflows pretty locked down (especially on release)

.github/workflows/release.yml
```diff
@@ -930,3 +969,27 @@ jobs:
                     --channel unstable \
                     --version "$SHORT_SHA" \
                     --bucket gs://minimal-one
+
+    publish-brew:
```
norrietaylor: Blocking: This ships to users on tag push, bypassing the documented promotion gate.

.github/workflows/release.yml
```diff
@@ -90,6 +90,43 @@ jobs:
                   [ "$fail" -eq 0 ] && echo "all five required checks are green on $sha"
                   exit "$fail"
 
+    assert-version:
```
norrietaylor, line 93: **Logic in the frozen YAML instead of `scripts/`.**

CONTRIBUTING: "add it to `scripts/` or the `justfile`" Let the lane call it, not by growing the YAML. 

- this job — inline `sed` extraction of `package.version`, comparison, and error text, ~20 lines of untested logic;
- `promote.yml:294-299` — inline `git describe ... || true` semver resolution;
- `release.yml:944-947` — inline `gcloud storage cp dist/*.deb …`, while every other staging write goes through `stage-release.sh`.

The repo's own precedent for exactly this shape is `verify-nightly-provenance.sh` + `verify-nightly-provenance_test.sh` + `just test-promote-gate`. Had the `git describe` step been a script, `scripts/lint-shell.sh` would cover it and a `_test.sh` with a stubbed environment would have caught the missing-`git` silent no-op flagged on promote.yml:293 — that defect reaches review only because the logic sits in YAML.

justfile
```diff
@@ -152,6 +152,27 @@ minimal-cli:
...
+    scripts/dist-build.sh {{target}}
+
+# Build the .deb/.rpm/.apk packages for a staged semver release from the
+# installer bucket via the pinned nfpm (fetched + sha256-verified by the
+# script). Packages land in dist/. Apt/dnf/apk repo-tree hosting lives in the
+# infra repo, not here.
+#
+# Package a staged semver's Linux artifacts, e.g. `just pkg-nfpm 0.5.3`.
+pkg-nfpm pkgver:
```
norrietaylor, line 182: **New release-critical shell ships with no test harness.**

`git diff --name-only main...pr-1313 | grep -i test` returns nothing: five new scripts (~1,000 lines), no `_test.sh`, and no lane or `just ci` path calls the new `dist-build` / `pkg-nfpm` recipes. Two in-repo precedents both stub network and auth — `install_test.sh` (+ `just test-installer`) and `verify-nightly-provenance_test.sh` (+ `just test-promote-gate`). This is the mechanism behind the unexercised-template finding on publish-aur.sh:164.

One gate does hold, worth recording: all five scripts pass shellcheck, and they are covered — `scripts/lint-shell.sh` globs `scripts/*.sh` and `crates/common/tests/shell_lint.rs` drives it from the workspace suite. That convention-discovered gate works as designed here.

packaging/nfpm.yaml
```diff
@@ -0,0 +1,143 @@
...
+#                    renamed to their installed names (min, minimald, mip,
+#                    minvmd, gvproxy-min)
+#   COMPLETIONS_DIR  completions generated from the built `min` (the
+#                    scripts/dist-build.sh technique)
+#   APPARMOR_DIR     the repo's packaging/apparmor tree
+#   APPARMOR_LOADER  the repo's scripts/install-apparmor-profile.sh
+# The scripts.* paths are NOT env-expanded by nfpm, so postinstall points at
+# the fixed .scratch location the driver materializes on every run.
+#
+# The file list mirrors the AUR package (aur/PKGBUILD) with one deliberate
```
norrietaylor: **Dangling reference.** `aur/PKGBUILD` does not exist in this tree — the file is `packaging/arch/PKGBUILD-bin.tmpl`. Same stale path at `scripts/dist-build.sh:130`. (`publish-aur.sh:145`'s `$workdir/aur/PKGBUILD` is correct; that one is the cloned AUR repo.)

Line 24-26 also cites "the packaging plan's Phase 3 file list", which is not in the repo — there is no packaging spec under `docs/specs/`.

norrietaylor (changes requested):
I'm excited to get this in, but I would feel a lot more comfortable if the design was more modular and presented in a new workflow. 

Consider something like this:

```
# .github/workflows/publish-packages.yml
on:
  workflow_call:
    inputs:
      version: {type: string, required: true}   # semver, no v
      sha:     {type: string, required: true}   # the promoted short sha
      dry_run: {type: boolean, default: false}
  workflow_dispatch:
    inputs: {version: …, sha: …, dry_run: …}
permissions:
  contents: read

```
and in promote.yml, one job:

```
publish-packages:
    needs: promote
    if: ${{ inputs.target == 'stable' }}
    uses: ./.github/workflows/publish-packages.yml
    with:
        version: ${{ needs.promote.outputs.semver }}
        sha:     ${{ needs.promote.outputs.sha }}
        dry_run: ${{ inputs.dry_run }}
    secrets: inherit
```

You will want to resolve the semver on the runner, and keep the YAML a thin scheduler, without a bunch of embedded scripting. 

Recommend reading (or having your agent read): 

- `CONTRIBUTING.md` specifically the test-extension contract
- `docs/ci-strategy.md` 
- docs/internal/release-pipeline.md

scripts/dist-build.sh
```diff
@@ -0,0 +1,165 @@
...
+            export "$LINKER_VAR=musl-gcc"
+        fi
+        ;;
+esac
+
+# One invocation per package — see the header for why the fat-LTO links must
+# serialize and why minvmd never shares an invocation with the rest.
+cargo build --release --locked --target "$TARGET" --package mip
+cargo build --release --locked --target "$TARGET" --package minimal
+
```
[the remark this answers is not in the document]
0chroma: Keeping the cargo calls in the script. dist-build.sh is the one definition of a shippable build (the release jobs call it, so CI builds == downstream builds), and the justfile recipe `dist-build` is the local entrypoint that delegates to it — the per-package invocation ordering and the fail-loud libkrun checks are release-pipeline logic the repo convention keeps in `scripts/` (the same direction as the other inline-YAML findings here). Making the script call back into `just` would invert that ownership and add a just dependency to the release lane.

scripts/publish-brew.sh
```diff
@@ -0,0 +1,213 @@
...
+        *)         die "unknown argument: $arg (try --help)" ;;
+    esac
+done
+
+[ -n "${PKGVER:-}" ] || die "PKGVER is required (the promoted semver, without the v prefix)"
+case "$PKGVER" in
+    v*) die "PKGVER must not carry the v prefix: '$PKGVER' (use ${PKGVER#v})" ;;
+esac
+printf '%s\n' "$PKGVER" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-][0-9A-Za-z.-]+)?$' \
+    || die "PKGVER '$PKGVER' is not a semver X.Y.Z (optional -prerelease/+build)"
```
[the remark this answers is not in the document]
0chroma: Adopted for build metadata (`1.2.3+build.7` passes now). Prereleases stay rejected on purpose, per the finding on release.yml: Homebrew has no channels, so an RC would overwrite the tap's only formula for every user with no opt-out. The error message names that reason.

scripts/publish-brew.sh
```diff
@@ -0,0 +1,213 @@
...
+
+git commit -m "minimal $PKGVER"
+
+# Sanity guard: only ever push to the tap repo we cloned.
+remote_url="$(git remote get-url origin)"
+[ "$remote_url" = "$BREW_TAP_REPO" ] \
+    || die "refusing to push: origin '$remote_url' is not $BREW_TAP_REPO"
+# The tap's default branch; GitHub repos default to main (an empty repo takes
+# the name of the first push).
+git push origin HEAD:refs/heads/main
```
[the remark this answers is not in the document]
0chroma: Keeping the direct push to the tap's default branch. The tap repo is generated output of this repo (like the AUR repo this PR pushes to), not a development tree — the never-commit-to-main rule here is about this repo's branches. Homebrew taps are only read from the default branch, so a PR-and-merge flow would strand every release until someone merged it by hand, which defeats CI-owned publishing. The guard right above the push still pins origin to the repo it cloned.

.github/workflows/promote.yml
```diff
@@ -165,6 +165,8 @@ jobs:
             id-token: "write"
             # verify-nightly-provenance.sh lists nightly workflow runs.
             actions: "read"
+        outputs:
+            version: ${{ steps.promote.outputs.sha }}
```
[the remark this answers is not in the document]
0chroma: Flagged in the PR description's user-actions list. The workflow changes shrank in this round (publish-brew moved to promote.yml, inline logic moved to scripts/), and the new jobs declare explicit minimal `permissions` blocks; approval still needed from @gominimal/minimalists.

crates/common/tests/shell_harnesses.rs
```diff
@@ -0,0 +1,77 @@
...
+        .to_path_buf()
+}
+
+#[test]
+fn all_script_test_harnesses_pass() {
+    let scripts_dir = repo_root().join("scripts");
+
+    let mut harnesses: Vec<PathBuf> = std::fs::read_dir(&scripts_dir)
+        .unwrap_or_else(|e| panic!("cannot read {}: {e}", scripts_dir.display()))
+        .filter_map(|entry| entry.ok().map(|e| e.path()))
```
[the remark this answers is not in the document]
0chroma: Fixed in 54f9d174: directory-entry errors now panic with the scripts dir path instead of silently shrinking the suite.

crates/common/tests/shell_harnesses.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    assert!(
+        !harnesses.is_empty(),
+        "no scripts/*_test.sh harnesses found — the convention-discovered shell gate found nothing to run"
+    );
+
+    let mut failures = Vec::new();
+    for harness in &harnesses {
+        let name = harness.file_name().and_then(|n| n.to_str()).unwrap_or("?");
+        let output = Command::new("bash")
```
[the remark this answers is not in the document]
0chroma: Fixed: the runner now dispatches each harness under its declared shebang interpreter — install_test.sh runs under sh, the contract the shell-installer lane proves under dash.

scripts/assert-release-version.sh
```diff
@@ -0,0 +1,77 @@
...
+# crates/version/build.rs derives the version from `git describe` and falls
+# back to `package.version` only when no `v*` tag is reachable (e.g. a source
+# tarball with no .git). A stale package.version therefore makes exactly the
+# builds this repo ships from a tarball report the wrong `-V`. Tag pushes are
+# the only flow carrying a semver tag to assert; workflow-dispatch and the
+# nightly (release-<sha>) flows have none and skip.
+#
+# Usage: scripts/assert-release-version.sh [--tag v0.5.4] [--cargo-toml FILE]
+#   --tag         the ref being released (default: GITHUB_REF_NAME, or a fatal
+#                 error when neither is set)
```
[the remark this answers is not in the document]
0chroma: Fixed the text in 54f9d174: a non-tag flow intentionally asserts nothing and exits 0; the usage no longer calls that fatal.

scripts/publish-aur_test.sh
```diff
@@ -0,0 +1,171 @@
...
+missing=""
+for tool in git curl; do
+    command -v "$tool" >/dev/null 2>&1 || missing="$missing $tool"
+done
+if [ -n "$missing" ]; then
+    echo "publish-aur_test: skipping, no$missing on PATH"
+    exit 0
+fi
+
+root="$(mktemp -d 2>/dev/null || mktemp -d -t minimal-aurtest)"
```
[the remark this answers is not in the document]
0chroma: Fixed in 54f9d174: the publisher's nobody payload re-exports the caller's PATH and the scratch HOME (runuser resets both), and the harness chmods its root traversable so the stand-in can reach the stub.

scripts/publish-brew_test.sh
```diff
@@ -0,0 +1,148 @@
...
+    bad "the dylib lands in the prefix's lib/ under the load command's name"
+fi
+if [[ "$out" == *'bin.install "minimal-macos-arm64" => "min"'* ]]; then
+    ok "the CLI installs as bin/min (not a bin/min directory)"
+else
+    bad "the CLI installs as bin/min (not a bin/min directory)"
+fi
+
+# Every 64-hex digest in the diff must be one of the fixture assets'.
+digests="$(for a in "${assets[@]}"; do sha256sum "$release/$a"; done | cut -d' ' -f1)"
```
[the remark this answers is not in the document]
0chroma: Fixed: the harness extracts the publisher's portable sha256_file (one definition) instead of calling sha256sum directly.

scripts/resolve-release-version_test.sh
```diff
@@ -0,0 +1,122 @@
...
+# The head sits after v0.6.0; an RC sha must not fall through to v0.5.3.
+expect 3 "prerelease" "sha tagged only with an RC skips (no downgrade fallback)" -- resolve "$rc_sha"
+expect_out "0.6.0" "sha after an RC still resolves the released tag" -- resolve "$untagged"
+
+# Two tags on one commit: the newest release wins.
+expect_out "0.5.4" "newest of two tags on the same commit wins" -- resolve "$dual"
+
+# A repo with commits but no tags at all skips rather than fabricating one.
+git init -q "$repo/no-tags"
+git -C "$repo/no-tags" commit -q --allow-empty -m only
```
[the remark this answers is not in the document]
0chroma: Already fixed before this review landed: identity travels through GIT_AUTHOR_*/GIT_COMMITTER_* exports covering every repository the fixture creates, verified with GIT_CONFIG_GLOBAL=/dev/null (no identity anywhere).

scripts/resolve-release-version.sh
```diff
@@ -0,0 +1,115 @@
...
+    exit 3
+}
+
+# Tags pointing at the SHA itself, newest version first (--points-at takes
+# the next bare argument as its object, so every option precedes it).
+# Empty when none do.
+exact="$(git_out tag --sort=-version:refname --format='%(refname:short)' \
+              --points-at "$SHA" | released_tag)"
+if [ -n "$exact" ]; then
+    printf '%s\n' "${exact#v}"
```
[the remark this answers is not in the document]
0chroma: Fixed in 54f9d174: only the newest exact-match tag is resolved, and the harness's dual-tag case now really tags two releases on one commit (it previously tagged one, so the case was not exercised).

.github/workflows/promote.yml
norrietaylor: During the last promote job, I realized that this issues is part of our legacy process and need fixing as well. 
Still, I would prefer we sort out the "order of operations" before we exacerbate it further. 

I'd still like this to be more modular if possible. Is there a way we can stage build, stage, then release the packages in separate steps that align with with a build, release(ie stage), and promote workflows we already have?

norrietaylor (approved):
I'd like to see us move to a a process where we can build and smoke packages before we publish them, but this PR is a step in the right direction

scripts/pkg-smoke.sh
```diff
@@ -0,0 +1,288 @@
...
+for i in "${!fmts[@]}"; do
+    case "${fmts[$i]}" in
+        deb)
+            boxes+=("min-test-deb")
+            images+=("debian:latest")
+            installs+=("sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y")
+            uninstalls+=("sudo dpkg -r minimal")
+            # The expected version string is checked against the manager's own
+            # record; dpkg-query -f '\${Version}' expands inside the box.
+            version_probes+=("dpkg-query -W -f '\\\${Version}\\n' minimal")
```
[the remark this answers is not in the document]
0chroma: Accepted, fixed in 50bfb140 with a form that removes the quoting question entirely: `dpkg-query -W minimal | cut -f2` (the default `-W` output is TAB-delimited), so nothing shell-sensitive crosses the distrobox hop.

One correction to "can never pass": the pre-fix live run against the shipped 0.5.3 deb did pass (e4b11352, host run through distrobox on a Vanilla OS host). That is consistent with your own first web query above: distrobox historically stringifies and re-parses arguments, which strips the escape. dpkg-query therefore received `${Version}\n` without the backslash and the check matched. The escape semantics diagnosis was right, the conclusion about the runtime path was not. The fix makes the behavior intentional either way. The `usage()` range is fixed to the package-nfpm.sh sentinel form in the same commit.

PR #1320, opened 2026-08-28
feat: if fish is set to a known and installed shell, start it instead of bash for new sessions

evanspearman:
## Summary

  Attaching to a session always dropped you into bash. The only way to land
  somewhere else was a `PROMPT_COMMAND` that `exec`s another shell — a hack that
  fires per prompt, is a composed var the MOTD recipe's `unset` can eat, and runs
  after bash has already started.

  Setting `SHELL` in a loadout's `[vars]` now picks the shell the attach mints:

  ```toml
  packages = ["fish"]

  [vars]
  SHELL = "/usr/bin/fish"
```

  Five shells, each because the registry ships a package for it and the daemon
  already installs it an attach-env TERM hook: bash, sh, zsh, fish, nu.
  Only the file name of SHELL is read, so a value carried over from the host
  (/opt/homebrew/bin/fish, a Nix store path) resolves fine. Anything else — not
  installed, or not in the table — starts bash and prints one line saying so,
  naming the package to add. Nothing fails.

  The daemon was most of the way here already: it installs per-shell hooks for
  exactly these shells and publishes the attach-env in POSIX/fish/JSON form, with
  the reason in the code ("a session's shell is whatever the user runs, not only
  the bash the daemon spawns"). Only the spawn was still hardcoded.

  SHELL is declared, not sniffed — nothing reads the $SHELL of the terminal you
  ran min from. No client or wire change.

  Three things the docs now state plainly: bash keeps its existing rc suppression
  while other shells read their own startup files from the session home (where
  patches land — deliberate); the shell is chosen once, at the attach that mints
  it; and the orientation banner rides PROMPT_COMMAND, so a non-bash session
  doesn't print it (TERM refresh works in all five).

  Testing

  just fmt-check and just clippy clean. cargo test --workspace --locked --no-fail-fast — 1862 passed, 14 failed; all 14 are pre-existing
  vcs: invalid remote path failures in mctx/minimald env::tests from a
  poisoned checkout cache in my environment, confirmed by a control run with my
  changes stashed. 13 doctests pass.

  New coverage: 10 resolver unit tests over a fake rootfs (default bash with
  today's exact argv; installed shell + its argv; host paths resolved by file name;
  sh through the symlink the bash package ships; absent shell falls back with a
  notice naming the package — nushell, not nu; a present-but-unknown shell
  still falls back; malformed values; table self-consistency), plus a
  layer_session_env case pinning that a composed SHELL reaches the map.

  Not run: just e2e. A daemon won't start in my environment — a global git
  insteadOf rewrite makes checkouts::Repo::new reject the cached clone, the
  same root cause as the 14 unit failures. The two new e2e blocks (a SHELL=/usr/bin/sh
  loadout asserting $0 through a real pty attach, and an uninstalled zsh
  asserting bash + the notice) are bash -n-valid but unexecuted and need a run on
  a normal host before merge.

  Checklist

  - [x] BREAKING CHANGE: footer present if this is a breaking change

  Docs touched: `reference/loadouts.md` (new `SHELL` section + the attach-shell opener) and `concepts/sessions.md`.

## Summary by CodeRabbit

* **New Features**
  * Interactive sessions now launch the shell selected through `SHELL`.
  * Supported shells include bash, sh, zsh, fish, and nushell, with shell-specific prompts and startup behavior.
  * Invalid or unavailable shells fall back to bash with installation guidance.
  * Improved session command handling, including task execution within existing sessions and reliable argument and exit-code forwarding.

* **Documentation**
  * Updated session and loadout guides with shell selection, requirements, fallback behavior, and prompt details.

* **Tests**
  * Expanded coverage for shell selection, fallback behavior, prompts, command routing, and task execution.

### DIFF
--- a/crates/minimald/src/env.rs
+++ b/crates/minimald/src/env.rs
@@ -560,6 +560,22 @@ impl Env {
         self.sandbox.new_container().map_err(sandbox_err_to_io)
     }
 
+    /// The assembled session rootfs on the daemon's filesystem.
+    ///
+    /// Exposed so the launcher can answer "is this shell installed in
+    /// this session" before it spawns one
+    /// ([`crate::session_shell::resolve`]) — the same question
+    /// [`install_min_helpers`] answers about directories it writes into.
+    /// Only meaningful once [`Env::build`] has returned: the rootfs is
+    /// what that assembles.
+    ///
+    /// Read only by the real `SandboxLauncher` (`cfg(not(test))`), which
+    /// the mock launcher stands in for under test.
+    #[cfg_attr(test, allow(dead_code))]
+    pub(crate) fn rootfs(&self) -> std::path::PathBuf {
+        self.sandbox.rootfs()
+    }
+
     /// Builds a command to run `program` inside the given container.
     pub fn command<I, S>(
         &mut self,
--- a/crates/minimald/src/lib.rs
+++ b/crates/minimald/src/lib.rs
@@ -17,6 +17,7 @@ pub mod server;
 mod session;
 mod session_delta;
 pub mod session_host;
+mod session_shell;
 pub mod session_sop;
 mod sessions;
 mod sftp;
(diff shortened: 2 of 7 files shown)


norrietaylor (approved):
Appreciate that you added e2e tests!

PR #1331, opened 2026-08-31
feat(containers): plumb container specs into graph, spechash gen, cache meta

twitchyliquid64:
* Accumulate container specs into `Graph`
 * Implement spec hashing for container specs
 * Add cache meta for entries which are containers
 * Add support for streaming into/out-of the graph wire representation

## Checklist

- [ ] Docs updated if behavior changed
- [ ] `BREAKING CHANGE:` footer present if this is a breaking change

## Summary by CodeRabbit

- **New Features**
  - Added support for importing, storing, retrieving, and serializing container specifications.
  - Added stable container specification hashing, including package references, image settings, optional values, maps, and sets.
  - Added synchronous and asynchronous container read/write support.
  - Added ordered container iteration and name-based retrieval.

- **Bug Fixes**
  - Graph loading now validates referenced packages and rejects duplicate container names with a clear error.
  - Container metadata is excluded from build searches by specification name.

- **Tests**
  - Expanded coverage for container persistence, hashing, ordering, malformed data, and sync/async compatibility.

### DIFF
--- a/crates/graph/src/spec_hasher.rs
+++ b/crates/graph/src/spec_hasher.rs
@@ -1,9 +1,10 @@
 use crate::{
-    BuildDep, BuildOutput, BuildSpec, BuildSpecRef, Graph, RuntimeDep, SourceFetch, SpecHash,
+    BuildDep, BuildOutput, BuildSpec, BuildSpecRef, Error, Graph, RuntimeDep, SourceFetch, SpecHash,
 };
 use blake3::Hasher;
 use common::{SubsetSpec, Target, target};
-use decode::AttrValue;
+use decode::{AttrValue, Container};
+use nickel_lang_core::term::IndexMap;
 use smallvec::SmallVec;
 use std::collections::HashMap;
 use std::io::Write;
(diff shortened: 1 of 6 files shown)


crates/graph/src/spec_hasher.rs
```diff
@@ -418,6 +419,174 @@ impl SubsetHasher {
...
+        Some(argv) => {
+            enc.tag(b"some");
+            enc.index(argv.len());
+            for arg in argv {
+                enc.bytes(arg.as_bytes());
+            }
+        }
+        None => enc.tag(b"none"),
+    }
+}
```
norrietaylor, line 555: `enc.index(argv.len())` fixes the arity, not the element boundaries. The elements are concatenated after it, so two argvs of the same length whose concatenations agree hash identically:

```
["/bin/sh", "-c"]  ->  some | 2 | /bin/sh-c
["/bin/sh-", "c"]  ->  some | 2 | /bin/sh-c
```

Two containers with different exec-form entrypoints then share a SpecHash, and the second resolves to the first's cache entry and ships the wrong entrypoint.

`container_hash_separates_string_boundaries` covers the split/joined case (`["/a", "/b"]` vs `["/a/b"]`), which differs in arity and is caught by the count prefix. The equal-arity case is uncovered.

A per-element length prefix, or the in-band marker `encode_string_map` already uses for `k`/`v`, closes it.

crates/graph/src/spec_hasher.rs
```diff
@@ -418,6 +419,174 @@ impl SubsetHasher {
...
+/// the image config, so neither order nor a repeat says anything.
+fn encode_string_set<E: SpecEncoder, S: AsRef<str> + Ord>(mut values: Vec<S>, enc: &mut E) {
+    values.sort_unstable();
+    values.dedup();
+
+    enc.index(values.len());
+    for value in &values {
+        enc.bytes(value.as_ref().as_bytes());
+    }
+}
```
norrietaylor, line 588: Same framing gap as `encode_opt_argv`, and reachable from a valid `minimal.ncl`:

```
volumes = ["/a", "/a/b"]  ->  2 | /a/a/b
volumes = ["/a/a", "/b"]  ->  2 | /a/a/b
```

Both sort ascending to the same concatenation and carry the same count. Both pass `check_volume_path` and decode's duplicate check, so two distinct image declarations share one cache key.

The count prefix does not separate them: it is written after `dedup`, so it is a function of the set's cardinality, not of where the elements divide.

`exposed_ports` goes through the same path; `"{port}/{proto}"` makes a collision harder to construct but does not rule it out.

crates/graph/src/spec_hasher.rs
```diff
@@ -418,6 +419,174 @@ impl SubsetHasher {
...
+        encode_opt_bytes(arch.as_ref().map(|a| a.as_nickel_str()), enc);
+        enc.tag(b"-working_dir");
+        encode_opt_bytes(working_dir.as_deref(), enc);
+        enc.tag(b"-user");
+        encode_opt_bytes(user.as_deref(), enc);
+        enc.tag(b"-stop_signal");
+        encode_opt_bytes(stop_signal.as_deref(), enc);
+
+        enc.tag(b"-env_vars");
+        encode_string_map(env_vars, enc);
```
norrietaylor, line 507: `encode_string_map` sorts by key, so `env_vars` order never reaches the hash. `container_map_ordering_survives_roundtrip` in `wire.rs` asserts the opposite for the same field — "an image config is order-sensitive, so the wire format must not reorder them" — and `container_hash_ignores_collection_order` asserts `{A:1,B:2}` and `{B:2,A:1}` hash equal. One of the two is wrong.

OCI `Env` is a JSON array of `KEY=VALUE` strings, not an object, so order is part of the config blob and therefore part of the image digest. Taking the wire test as correct, two declarations differing only in `env_vars` order produce different images under one SpecHash, and the second is served the first's image from cache.

`labels` and `config` become JSON objects, so sorting them is sound provided emission sorts as well. `env_vars` wants sequence semantics — the same treatment `cmd` and `entrypoint` get.

norrietaylor (commented):
Three issues in the container SpecHash encoding, all with the same consequence: the hash under-determines the image it keys, so two distinct declarations can share a cache entry and the second is served the first's artifact.

Two are missing per-element framing (`encode_opt_argv`, `encode_string_set`); the third is `env_vars` being hashed as an unordered map while `wire.rs`'s own test treats its order as load-bearing. Details inline.

norrietaylor (approved):
Approved, but the bots looks like they may be on to a couple things. You may want to address those comments first.

PR #1341, opened 2026-09-02
feat(op): capture upstream NOTICE/COPYRIGHT files at source extraction (#1040)

### LINKED ISSUE #1040: Upstream NOTICE/COPYRIGHT capture at build time (minimal-side; #282 remainder)
The one remainder from gominimal/inbox#282 (now closed): the attribution manifest ships SPDX **canonical** license texts, but several licenses require the upstream's **own** files to travel with distribution — Apache-2.0 §4(d) requires the NOTICE file's attribution contents; many permissive licenses require the upstream COPYRIGHT notice verbatim.

**Why this is minimal-side:** only the builder holds the unpacked source tree. build-servers seals the manifest from graph + cache data after the fact; it never sees upstream file contents. Emission has to happen at build/package time.

**Sketch:**
- At package time, capture `NOTICE`, `NOTICE.txt`, `COPYING`, `COPYRIGHT` (top-level, case-insensitive; bounded size) from the source tree into a sidecar object keyed by content hash (or into the output tarball under a reserved path).
- The per-commit attribution manifest (build-servers#194, `<commit>.attribution.json`) gains an optional per-package `notice` pointer next to the existing `licenses` mapping — the manifest is already the carrier; this is one field.
- Scope: package-level only, same as gominimal/inbox#282. Statically-vendored third-party code remains SBOM territory (minimal#700).

**Acceptance:** for an Apache-2.0 package whose upstream ships a NOTICE (e.g. several in the catalog), the attribution bundle reachable from the cache includes that NOTICE's contents, without network access to upstreams.

bryan-minimal:
Closes #1040 (minimal side). Follow-up in gominimal/build-servers seals the pointer into the attribution manifest.

## What

Some licenses require the upstream's own files to travel with a distribution — Apache-2.0 §4(d) for `NOTICE`, verbatim copyright notices for most permissive licenses. Only the builder holds the unpacked tree, so this captures them at extraction time:

- `op::notices` — after `extract_compressed_tar`, scan the tree root (or its lone top-level dir when no `strip_prefix`) for `NOTICE*` / `COPYRIGHT*` / `COPYING*` (case-insensitive, regular files only, ≤512 KiB, ≤8 files) and record `<cache>/notices/<source-sha256>.json` — `{schema_version, source_sha256, files:[{name, sha256, text?, base64?}]}` (`text` when valid UTF-8, `base64` otherwise — bytes travel verbatim), selection by metadata before any read, sorted, written once (write-then-rename), idempotent. "Scanned, nothing" is recorded too.
- `lcache::Cache::notices_dir()` + the dir created at `at_dir`.
- `SourceLoad` records for `Web` sources only, keyed by the declared `sha256` — the key the sealed attribution manifest already carries per package (`sources[].sha256`). Local files are project-side and never appear there. Best-effort: a failure logs a warning and never fails the build.
- `mip cache notices <sha256>` prints a record.

## Why this shape

Keying by the archive hash means one record per tarball rather than per package build, stable across rebuilds and arches, and a pure server-side join for build-servers: `sources[].sha256` → `notices/<sha256>` object → optional `notices` field on `PackageAttribution` (schema 1 → 2, additive). No output-tarball or graph changes.

## Audit backing the scope

Streamed all 86 source tarballs of the 71 Apache-2.0 packages in manifest `ff3ca2f4`: 7 ship a `NOTICE` today (bat, codex, cups, maven, otel-collector, trivy, typst), 8 more ship `COPYRIGHT`/`COPYING`. Details on #1040.

## Tests

- `op::notices`: stem matching, regular-file/symlink/size/sort rules, lone-top-dir descent, idempotent record + read-back, non-digest key rejection.
- `op::sources`: end-to-end `SourceLoad` over a generated tar.gz with `strip_prefix` records the NOTICE under the declared sha256.

## Summary by CodeRabbit

- **New Features**
  - Added attribution notice tracking for remote source archives.
  - Added a cache command to retrieve recorded notices as formatted JSON using an archive checksum.
  - Notices are saved for extracted archives in both temporary and specified locations.
  - Added support for preserving binary notice files while displaying readable text when available.

- **Improvements**
  - Notice file selection is now deterministic and limited to a maximum number of files.
  - Notice recording failures generate warnings without interrupting builds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4724,6 +4724,7 @@ name = "op"
 version = "0.0.1"
 dependencies = [
  "anyhow",
+ "base64 0.23.1",
  "common",
  "decode",
  "flate2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4738,6 +4738,7 @@ dependencies = [
  "ot",
  "rayon",
  "sandbox2",
+ "serde",
  "serde_json_lenient",
  "sha2 0.11.0",
  "shlex",
--- a/crates/lcache/src/lib.rs
+++ b/crates/lcache/src/lib.rs
@@ -243,7 +243,7 @@ impl Cache<LocalDir> {
     pub fn at_dir<P: AsRef<Path>>(p: P) -> Result<Self, std::io::Error> {
         let fs = LocalDir::with_base(p)?;
 
-        for dir in &["temp", "meta", "alog"] {
+        for dir in &["temp", "meta", "alog", "notices"] {
             match fs.mkdir(dir) {
                 Ok(()) => Ok(()),
                 Err(e) => {
@@ -296,6 +296,12 @@ impl Cache<LocalDir> {
 
         let p = self.inner().fs.path().join(subpath);
         std::fs::remove_dir_all(p).map_err(CacheErr::from)
+    }
+
+    /// Where upstream attribution files captured from extracted sources
+    /// live: `<cache>/notices/<source-sha256>.json`.
+    pub fn notices_dir(&self) -> PathBuf {
+        self.inner().fs.path().join("notices")
     }
 
     /// Allocates a temporary directory in the same filesystem as the rest of the cache.
--- a/crates/mip/src/cmd_cache.rs
+++ b/crates/mip/src/cmd_cache.rs
@@ -25,6 +25,14 @@ pub enum CacheArgs {
         #[arg(long, value_parser = parse_duration, default_value = "14d")]
         older_than: Duration,
     },
+    /// Show the upstream attribution files captured from a source archive
+    ///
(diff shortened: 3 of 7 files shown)


bryan-minimal:
Addressed both CodeRabbit findings in 047d9ae: `scan_dir` now selects and truncates on metadata before reading anything (`scan_caps_the_file_count_by_name_order`), and contents are lossless — `text` only when valid UTF-8, `base64` otherwise, with the sha256 always over the original bytes (`scan_keeps_non_utf8_bytes_verbatim`). Record shape in the PR body updated accordingly.

norrietaylor (approved):

PR #1346, opened 2026-09-03
feat: support loadout.toml in per-loadout directories

evanspearman:
Summary

  Loadouts could only be discovered at <config>/minimal/loadouts/<name>.toml, which left nowhere to put a loadout you want under version control — the definition is a loose file in a directory shared with every other loadout. This adds a second layout, <config>/minimal/loadouts/<name>/loadout.toml, so a whole
  loadout (definition, $LOADOUT_ROOT assets, hook scripts) is one self-contained directory you can git clone into place.

  The change is confined to discovery. A loadout already owned a directory at <loadouts_dir>/<name>/ — the anchor for $LOADOUT_ROOT and external hook scripts — and Source::loadout_dir derives it from the loadout's name, not from its file's path. In the new layout that directory is simply the one holding
  loadout.toml, so expansion, composition, hook staging, the wire types, minimald, and minimal-tui are all untouched. Migration is mkdir dev && mv dev.toml dev/loadout.toml.

  - sessions::client::disk — adds Layout, LOADOUT_FILE_NAME, read_loadout_dir, and load_loadout as the single name→loadout resolver; list_loadouts enumerates both layouts keyed by name. New LoadError::{ConflictingLayouts, NotFound}.
  - Defining one name in both layouts is a hard error, not a precedence rule — mirroring mfile::File::from_dir's ConflictingLayouts for minimal.toml / .minimal/minimal.toml. Whichever file lost would otherwise go on looking live while having no effect.
  - min — activation, the zero-config default probe, min loadout list, and min bug collection all handle both layouts.
  - Also closes a path-traversal wart in the function being replaced: --loadout ../../evil was interpolated into a filename unvalidated and read an arbitrary .toml. The post-read stem check never caught it, since file_stem("../../evil.toml") is evil. Names are now validated before the join.

  Two documented limitations: a cloned <name>/ contains .git, which a broad $LOADOUT_ROOT/**/* would sweep in; and <name>/ must be a real directory, not a symlink into a dotfiles repo, because resolve_under_anchor refuses a symlinked anchor. That security check is deliberately left alone.

  Testing

  cargo fmt --all -- --check and cargo clippy --workspace --all-targets --locked -- -D warnings clean. Full cargo test --workspace --locked green.

  New coverage: 15 tests in disk.rs (both layouts, conflict, not-found naming both candidates, asset-directory-is-not-a-loadout, a flat loadout named loadout), 6 in loadouts.rs, an anchor-equivalence integration test in client_flow1.rs proving both layouts resolve $LOADOUT_ROOT to the same directory, and
  tests/bug.rs extended so a directory-layout loadout is proved collected and redacted.

  Manual, against a scratch --config-dir:

  $ min --config-dir $cfg loadout list        # dev (dir form) + flat, exit 0
  $ min --config-dir $cfg loadout list        # after adding dev.toml:
  loadout `dev` is defined twice:
    /tmp/lo-check/minimal/loadouts/dev.toml
    /tmp/lo-check/minimal/loadouts/dev/loadout.toml
  remove one.
  error: 1 loadout failed to load                                    # exit 1

  $ min ... session activate . --loadout typo
  error: --loadout `typo`: no loadout `typo` in `.../loadouts` (looked for `typo.toml` and `typo/loadout.toml`)

  A real activation with a directory-layout loadout printed Applying loadouts: vcdev and reached FinalizeSession with its $LOADOUT_ROOT patch and external hook script both resolving; it then failed on no such package: base, which is the local environment lacking a package cache.

  Checklist

  - [x] Docs updated if behavior changed — docs/reference/loadouts.md (layouts, conflict rule, both tree diagrams, version-control caveats), docs/concepts/loadouts.md, docs/reference/cli-min.md, README.md
  - [x] BREAKING CHANGE: footer present if this is a breaking change — not breaking; the flat layout is unchanged, and a bare <name>/ asset directory beside a flat <name>.toml still resolves to the flat file (pinned by a test)

## Summary by CodeRabbit

* **New Features**
  * Loadouts now support standalone `<name>.toml` files and `<name>/loadout.toml` directory layouts.
  * Directory-based loadouts can be cloned as self-contained bundles.
  * `$LOADOUT_ROOT` and lifecycle hooks resolve to the loadout’s name-based directory for either layout.
  * Loadout listings include both layouts while retaining valid entries when others fail.

* **Bug Fixes**
  * Invalid names, duplicate definitions, malformed loadouts, and unsafe symlinked directories now produce errors.
  * User-defined default loadouts correctly override the built-in fallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,8 +145,9 @@ exit
 
 The project's `minimal.toml` describes what every contributor's session
 needs; a **loadout** carries what *you* want on top: your editor, terminal
-multiplexer, shell config, and dotfiles. Loadouts are single TOML files under
-`~/.config/minimal/loadouts/`:
+multiplexer, shell config, and dotfiles. Loadouts live under
+`~/.config/minimal/loadouts/`, either as `<name>.toml` or — to keep one under
+version control, alongside the files it ships — as `<name>/loadout.toml`:
 
 ```toml
 # ~/.config/minimal/loadouts/dev.toml
@@ -177,7 +178,9 @@ on_activate = { type = "inline", value = "hx --grammar fetch >/dev/null 2>&1 || 
 
 Apply one with `min session activate --loadout dev --attach .`, or list it in
 `default_loadouts` under `[loadouts]` in `~/.config/minimal/config.toml` to have it join every
-session automatically. `min loadout list` shows what's available. The full
+session automatically. `min loadout list` shows what's available, in either
+layout — so `git clone <repo> ~/.config/minimal/loadouts/dev` is enough to
+pick up a loadout someone else published. The full
 schema (file patches, lifecycle hooks, environment-variable inheritance,
 composition rules) is in the
 [loadouts reference](docs/reference/loadouts.md).
--- a/crates/minimal/src/diag/collect.rs
+++ b/crates/minimal/src/diag/collect.rs
@@ -80,14 +80,55 @@ pub async fn config(w: &mut BundleWriter, paths: &DiagPaths) -> Result<(), anyho
     )
     .await;
 
+    // Both loadout layouts: `<name>.toml` and `<name>/loadout.toml`. Only
+    // the definition file either way — the rest of a loadout's directory
+    // ($LOADOUT_ROOT assets, hook scripts) is the user's own content and
+    // has never been collected.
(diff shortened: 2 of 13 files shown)


norrietaylor (approved):
Looking forward to this!

PR #1370, opened 2026-09-10
test(minimald): prove RustSlirp tap creation rootless and as root

norrietaylor:
## Summary

Spike harness, not a behaviour change. Draft on purpose: it exists to get an answer
out of CI, not to merge as-is.

Assessing spec 012 (#1354) surfaced a cheaper design than the spec prices for. Spec
012's complexity — `plan()` returning data, `abandon()`, `NetPlan`/`TapSpec`,
`PhaseOneAttachGuard`, the `unsafe from_raw_fd` — exists because there are **two**
own-IP attach mechanisms with **two** rollback owners:

- **DM2** (native Linux, `minimald` unprivileged): hakoniwa's `RustSlirp` makes the
  tap *inside* the sandbox's userns+netns, rootless, and hands the fd back via
  `Child.rustslirp_tapfd`.
- **DM1/3/4** (in-VM, `minimald` root): `open_tap` + `move_tap_into_netns` make the
  tap in the guest's root netns under `CAP_NET_ADMIN` and `ip link set … netns` it in.

If the in-VM path can use `RustSlirp` too, the two collapse into one mechanism
differing only in relay transport (unix socket vs vsock) — which is what
`docs/specs/03-spec-networking/networking-with-diagrams.md` already says the
difference should be. Diagram 7 labels *both* host panels `PTasks (each: userns +
netns + tap)`; diagram 6 maps R1 → B1 "Linux user namespaces (unprivileged netns +
tap)". The current in-VM mechanism is the deviation, so unification converges on the
documented design rather than departing from it.

Settled by reading hakoniwa at the pinned rev (`souk4711/hakoniwa@0810ffa3`,
`src/unshare/newnet/rustslirp.rs`) and needing no test: **`RustSlirp` does not replace
gvproxy.** Despite the name there is no slirp, no TCP/IP stack and no NAT — it forks,
`setns`es into the target's userns then netns, brings up `lo`, builds a TUN/TAP via
`tun_rs`, adds a default route and returns the fd over `SCM_RIGHTS`. A one-for-one
rootless substitute for `open_tap` + `move_tap_into_netns`, nothing more. gvproxy
remains switch + NAT + port-forward API either way.

### The blocking question

Does `RustSlirp`'s `setns(CLONE_NEWUSER)` + in-netns tap creation still succeed when
the hakoniwa parent is **root**? `sandbox2/src/lib.rs` calls `.uidmap(1000)`, so only
the unprivileged case has ever run. Root should be a superset — `CAP_SYS_ADMIN` for
the setns, `CAP_NET_ADMIN` in the joined netns — but the combination is untested.

### What this adds

One harness, `crates/minimald/tests/rustslirp_root_integration.rs`, building the
container exactly as `sandbox2::new_container` does for an own-IP PTask:

- `rustslirp_tap_comes_up_unprivileged` — the **DM2 baseline**, which has no automated
  coverage today.
- `rustslirp_tap_comes_up_as_root` — the blocking question, re-execing the test binary
  under `sudo -E env …` so the hakoniwa parent is root. Separate test functions rather
  than an env-var A/B, since CI cannot flip a new variable without a workflow edit.

A `preflight()` separates a missing host prerequisite from a real negative. Without
`/dev/net/tun` **neither** mechanism can work, and reporting that as "RustSlirp handed
out no tap fd" would read as a failed spike and wrongly trip its kill criteria — I hit
exactly that locally. The probe also captures the sandboxed child's exit status, so a
dead sandbox and an unmade tap are distinguishable.

Both are `#[ignore]`d and additionally early-return unless `MINIMALD_NETNS_TEST` is
set, matching `netns_root_integration.rs` and `nsenter_integration.rs`, so a default
`cargo test` never unshares anything.

### Why no workflow change

`.github/workflows/` is frozen; none of this needs it. The `_root_integration` suffix
is auto-discovered by `ci-linux-native`'s `minimald-root-integration` job, which
already supplies sudo, unprivileged userns
(`kernel.apparmor_restrict_unprivileged_userns=0`), a netns-capability probe and
`MINIMALD_NETNS_TEST=1`. **No KVM and no VM needed** — this is the fastest available
signal on the blocking question.

Refs: #1354

## Testing

Everything below is local; the proofs themselves cannot run on my workbench, which is
why this PR exists.

- `cargo test -p minimald --test rustslirp_root_integration --no-run --locked` — builds
  clean, no warnings.
- Ungated run: `test result: ok. 2 passed; 0 failed` — both self-skip green, so a
  default `cargo test` cannot be reddened by this.
- `cargo clippy -p minimald --tests --locked -- -D warnings` — clean.
- `cargo fmt -p minimald -- --check` — clean.
- `every_integration_test_binary_ends_in_integration` naming guard — passes.

**Not run locally, and cannot be:** this host has no `/dev/net/tun`, no `/dev/kvm` and
runs as uid 1000. A gated run here correctly reports the prerequisite rather than a
result:

```
prerequisite missing: /dev/net/tun is absent, so NEITHER tap mechanism can
work here. This is a host limitation, not a result for RustSlirp.
```

The `minimald-root-integration` job is the execution venue. What its result means:

- **Both pass** → unification is viable. Spec 012 steps 4–5 get materially cheaper: one
  attach path, one rollback owner, and deletion of `attach_own_ip`, `open_tap`,
  `move_tap_into_netns`, `tap_netns_commands`, the `ip`/`nsenter` subprocess dependency
  and its `trusted_program` hardening (~200 lines plus tests). `plan()` and `abandon()`
  still survive — `RustSlirp` sets the address inside hakoniwa's spawn
  (`create_tun_device` → `.ipv4(…)`), so the pre-spawn lease and its rollback
  obligation remain.
- **Root fails and it is not a config-level fix in `sandbox2`** → stop. Unification
  would mean patching a git-pinned third-party fork, which changes the cost calculus
  entirely; spec 012 proceeds as written. The harness is worth keeping either way — it
  documents the root/rootless behaviour permanently and the baseline is coverage that
  does not exist today.

The follow-on transport-generalization patch and its VM-lane harness are deliberately
**not** in this PR; they are gated on this answer.

## Checklist

- [x] Docs updated if behavior changed — no behaviour change; test-only
- [x] `BREAKING CHANGE:` footer present if this is a breaking change — n/a

### DIFF
--- a/crates/minimald/tests/rustslirp_root_integration.rs
+++ b/crates/minimald/tests/rustslirp_root_integration.rs
@@ -1,0 +1,328 @@
+//! Spike proofs for unifying the own-IP tap mechanism on hakoniwa's `RustSlirp`.
+//!
+//! Today an own-IP PTask gets its tap in one of two ways, and which one depends
+//! on whether `minimald` is root:
+//!
+//! * **DM2, native Linux, unprivileged** — `RustSlirp` creates the tap *inside*
+//!   the sandbox's user+network namespace and hands the fd back via
+//!   `hakoniwa::Child::rustslirp_tapfd`. No host `CAP_NET_ADMIN`.
+//! * **DM1/3/4, in a libkrun VM, root** — [`minimald::net::switch::open_tap`] plus
+//!   `ip`/`nsenter` ([`minimald::net::switch::tap_netns_commands`]) create the tap
+//!   in the guest's root netns and move it in. Needs `CAP_NET_ADMIN`.
+//!
+//! Two mechanisms means two rollback owners, and it is the root cause of most of
+//! the machinery proposed in `docs/specs/12-spec-sandbox-network-provider`.
+//! `docs/specs/03-spec-networking/networking-with-diagrams.md` says the two
+//! deployments should differ only in *relay transport* — diagram 7 labels both
+//! host panels `PTasks (each: userns + netns + tap)`, and diagram 6 maps the
+//! isolation requirement to building block B1, "Linux user namespaces
+//! (unprivileged netns + tap)", at no per-invocation root. `RustSlirp` is B1; the
+//! in-VM `open_tap` path is not.
+//!
+//! So: can the in-VM path use `RustSlirp` too? It is a superset-of-privilege
+//! question — `RustSlirp` `setns`es into the target's user *and* network
+//! namespaces (`souk4711/hakoniwa`, `hakoniwa/src/unshare/newnet/rustslirp.rs`),
+//! and the rootless path assumes an unprivileged userns, which `sandbox2` gets
+//! from its `uidmap(1000)`. Root should be able to do everything that path does,
+//! but the combination has never been exercised. These two proofs exercise it:
+//!
+//! * `rustslirp_tap_comes_up_unprivileged` — the DM2 baseline. This mechanism is
+//!   in production on the native path today and has no automated coverage: the
+//!   own-IP proofs in `netns_root_integration.rs` drive the *other* mechanism
+//!   (`sudo ip tuntap add` + `open_tap` + `tap_netns_commands`), and
+//!   `scripts/session-e2e.sh` has no own-IP case at all. Without this there is no
+//!   baseline to compare the root result against.
+//! * `rustslirp_tap_comes_up_as_root` — the same proof with the hakoniwa parent
+//!   running as root, which is what `minimald` is inside the guest. A pass says
+//!   the in-VM path can drop `open_tap`/`move_tap_into_netns` and use the
+//!   rootless mechanism; a failure says unification needs a change in hakoniwa
+//!   itself (a git-pinned third-party dependency) and should be abandoned.
+//!
+//! Both build the container exactly as [`sandbox2::Sandbox::new_container`] does
+//! for an own-IP PTask — same uid/gid map, same namespaces, same `RustSlirp`
+//! settings — so a drift in that construction shows up here rather than only in
+//! a live session.
+//!
+//! Both are `#[ignore]` and additionally early-return unless `MINIMALD_NETNS_TEST`
+//! is set. Auto-discovered by the native lane's `minimald-root-integration` job
+//! via the `_root_integration` binary-name suffix
+//! (`-E 'binary(/_root_integration$/)'`) — ubuntu-latest with sudo and
+//! unprivileged userns. Neither KVM nor gvproxy is needed: neither proof attaches
+//! to a switch, because tap *creation* is the step that differs.
+//!
+//! To run locally you need a host with `/dev/net/tun`, unprivileged userns, and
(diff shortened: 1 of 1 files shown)


norrietaylor:
## CI result: Q1 passes

`minimald-root-integration` green on `21c7ef65` ([run](https://github.com/gominimal/minimal/actions/runs/34519776286/job/103014173017), 2m16s). Both proofs pass:

- `rustslirp_tap_comes_up_unprivileged` — the DM2 baseline holds.
- `rustslirp_tap_comes_up_as_root` — **`RustSlirp`'s `setns(CLONE_NEWUSER)` + in-netns tap creation works with a root parent.**

### Why this green is trustworthy

A green from a gated `#[ignore]` harness is worth checking before it is believed, since the same colour is what a silent skip produces. Four things had to hold, and do:

1. The job exports `MINIMALD_NETNS_TEST: "1"` and passes `--run-ignored all`, so `gated()` returns true and the `#[ignore]` is flipped — the tests were not skipped.
2. `--no-tests=fail` means the `binary(/_root_integration$/)` filterset must match something.
3. `preflight()` panics if `/dev/net/tun` or unprivileged userns is missing, so "no tap" cannot be reported when the host could never have made one either way.
4. `21c7ef65` added the last one: libtest exits 0 when its filter matches nothing, so asserting only on `sudo`'s exit status could not distinguish a passing root proof from an inner run that selected no tests. The outer branch now requires the inner run to report a test passing. The first push (`b750882b`) was also green, but that green did not yet exclude this case.

### What it means for spec 012 (#1354)

The blocking question is settled in favour of unification: the in-VM own-IP path can use the same rootless mechanism as the native one, so the two mechanisms differ only in relay transport — which is what `docs/specs/03-spec-networking/networking-with-diagrams.md` already describes.

Consequences for the spec as written:

- Steps 4–5 get materially cheaper: one attach path with one rollback owner, and `attach_own_ip`, `open_tap`, `move_tap_into_netns`, `tap_netns_commands`, plus the `ip`/`nsenter` subprocess dependency and its `trusted_program` hardening become deletable (~200 lines plus tests).
- 012-005 becomes uniform across deployment models rather than native-first.
- `plan()` and `abandon()` **still survive**. `RustSlirp` sets the address inside hakoniwa's spawn (`create_tun_device` → `.ipv4(…)`), so the pre-spawn lease and its rollback obligation remain regardless. Unification buys one mechanism, not the trait reshape.

### Scope note

Two caveats on how far this generalises, neither blocking. The runner is x86_64 on a GitHub-hosted kernel, not the libkrun guest kernel; and this proves tap creation, not end-to-end connectivity through the vsock relay.

Both are Q2's job — boot a guest, mint an own-IP session, assert egress + ingress + PTask↔PTask — which needs the transport-generalization patch and a harness in `crates/minvmd/tests/` to reach the KVM and macOS lanes. Deliberately not in this PR.

PR #1560, opened 2026-09-18
refactor: return errors instead of panicking on untrusted input

0chroma:
## Summary

Library code panicked on inputs it does not control. Parsing a malformed ELF
or PE binary, reading a non-UTF-8 export name, listing a package directory with
a non-UTF-8 file name, building an OCI spec, initializing the GCS client, or
creating a checkout tempdir all called `unwrap()`. Each of those is reachable
from a file, a config, or a network response, so a panic there takes down the
caller instead of returning an error.

This converts 35 such calls across six files to `?`, `ok_or_else`, or
`map_err` with context. No new error variants: every converted path reuses an
existing variant and the crate's existing `From` impls. Behavior changes only
where the old code panicked. It now reports an error.

## Left alone

- `graph/src/planner.rs`: every non-test `unwrap()` is an internal invariant
  (a key the algorithm just inserted, a vector it knows is non-empty). None
  takes external input, and converting them would only relocate the panic.
- `minimald/src/store.rs`: all sites are `expect(...)` with explicit invariant
  messages about actor liveness, which is the intended panic.
- Mutex locks, `matches!`-guarded values, `Regex::new` on constants, in-memory
  buffer writes, and all test code.

## Testing

```
just fmt-check   # pass
just clippy      # pass, no warnings
just test        # pass, 2103 tests run: 2103 passed, 13 skipped
```

`just test-ignored` needs root and could not run here. This phase is optional
and reverts file by file if any conversion proves unwelcome.

## Checklist

- [x] Docs updated if behavior changed (none: no behavior change)
- [x] `BREAKING CHANGE:` footer present if this is a breaking change (not breaking)

## Summary by CodeRabbit

* **Bug Fixes**
  * Replaced several crash scenarios with clear, propagated errors during file checks, dependency analysis, checkout creation, and sandbox setup.
  * Improved handling of invalid or non-UTF-8 filenames, paths, symbols, and repository data.
  * Remote cache initialization and build-graph setup now report configuration and storage failures instead of terminating unexpectedly.
  * OCI image generation now surfaces serialization and platform errors.
  * Package fetching, cache finalization, temporary-directory creation, and command lookup now properly report failures.


norrietaylor (approved):

norrietaylor:
Probably worth cleaning up the correctness issue before merging

PR #1562, opened 2026-09-18
docs(spikes): fold the SP2 review into the cohort classifier finding

norrietaylor:
Follow-ups on the merged cohort cgroup classifier spike from review of [#1539](https://github.com/gominimal/minimal/pull/1539).

- **Finding D3 names its condition.** The cgroup namespace confines migration only when cgroup2 is mounted with `nsdelegate`, which systemd sets (confirmed on the lima host's mount line) and a guest where the daemon is pid 1 must set itself. Without it the box's only barrier is write permission on the common ancestor, which a box running as the daemon's uid has. The host's cgroup2 mount must also stay out of the box's mount namespace. Action item 5 and the "VM-backed host" line carry both.
- **Action item 4 narrows the carve-out** to the resolver's address and port only, never loopback-wide: the NET spec's design reasoning already rejected a loopback baseline exception for every box because it gives the shared-namespace lane undeclared reach.
- **New action item 10** for T38 [#1482](https://github.com/gominimal/minimal/issues/1482) and T9 [#1460](https://github.com/gominimal/minimal/issues/1460): on a native host the resolver a host-address box reaches is the host's own, so a deny-all box with the carve-out resolves arbitrary upstream names, while design §5.3 expects those queries evaluated against the cohort's rules.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Documentation**
  - Clarified cgroup2 isolation requirements, including `nsdelegate` for guest mounts and hiding host mounts from box namespaces.
  - Confirmed support for native and VM-backed hosts.
  - Documented resolver handling for deny-all and native allow-list boxes, including forwarded-query considerations.
  - Clarified that unavailable host IP enforcement is reported as `none` for policy decisions rather than automatically rejecting declarations.
  - Assigned follow-up actions for resolver ownership, forwarding behavior, and architecture decisions.

PR #1563, opened 2026-09-19
docs(specs): close the macOS loopback and cgroup classifier questions in NET

norrietaylor:
## What

Closes the NET spec's HIGH (macOS per-box loopback mechanism) and MEDIUM (cgroup classifier layout, co-resident scope) open questions. Both were settled by the two spikes in [#1539](https://github.com/gominimal/minimal/pull/1539) and by the rulings in [gominimal/arch#78](https://github.com/gominimal/arch/issues/78), folded into networking design v0.8 by [gominimal/arch#79](https://github.com/gominimal/arch/pull/79). The LOW HTTP/2 and HTTP/3 question stays, waiting on [gominimal/arch#73](https://github.com/gominimal/arch/issues/73).

## Changes

- **Open questions**: the two settled markers are removed. The design reasoning states the settled answers in the spec's own words: the macOS mechanism is a root-held boot re-apply of the reserved range installed by the resolver advisory's command; the classifier decides on a per-box cgroup leaf beside the daemon's, confined by a cgroup namespace on a delegation-boundary mount, which is the box host's obligation; co-resident host-address boxes are in scope; a native host takes one privileged install step, with no per-box enforcement surfaced until then (ruling 1, shape 1).
- **NET-003** gains a sub-requirement: a host-address box resolves through a resolver Minimal owns, never the host's resolver (ruling 2): the node's DNS layer inside a VM-backed host, the box zone's answerer on a native host.
- **A new MEDIUM open question** carries what ruling 2 left open: on a native host, which component forwards a non-deny-all host-address box's upstream names under its rules, or whether native host-address allow lists are CIDR-only until one exists. The native answerer forwards nothing by design, so the spec does not claim a channel that does not exist.
- **NET-079** gains three sub-requirements: the deny-all carve-out is the answerer's address and port only, never loopback-wide; every process of a host-address box stays in the cgroup its verdict is decided on; a native host that cannot decide per box prints the install advisory at session start with no prompt and records the absent enforcement.
- **NET-124** gains a sub-requirement: NODATA and NXDOMAIN answers carry the zone's authority record so the host resolver caches them. The macOS loopback spike measured that an uncacheable negative stalls every lookup on the host.

No existing ID or test name changes. Spike findings: [macOS loopback alias](https://github.com/gominimal/minimal/blob/main/docs/spikes/2026-09-22-macos-loopback-alias.md), [cohort cgroup classifier](https://github.com/gominimal/minimal/blob/main/docs/spikes/2026-09-22-cohort-cgroup-classifier.md) (review follow-up in [#1562](https://github.com/gominimal/minimal/pull/1562)).

## Verification

`spec_lint.py` reports nothing blocking; the only advisories are the pre-existing unresolved test names for unbuilt requirements. The NET plan on [#1437](https://github.com/gominimal/minimal/issues/1437) is re-cut against this branch in the same pass.

## Summary by CodeRabbit

- **Documentation**
  - Clarified networking behavior when classification is unavailable: deny-all and egress host-address rules continue without individual verdicts and are recorded as unenforced rather than rejected.
  - Clarified that native non-deny-all name-based rules do not permit connections; native hosts use the host resolver, while address-based rules remain enforced.
  - Updated guidance on DNS behavior, confinement requirements, enforcement advisories, and native-host DNS forwarding rationale.


docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -126,6 +126,10 @@ included, with every refusal logged (NET-001 to NET-004).
   tier:     T0
   verify:   cargo nextest run -p minimald host_min_internal_resolves_to_host_reach_address_per_mode
   <!-- S1a/AC2; prose 3; ubiquitous; "inside boxes" per design §7.1 without qualification: an own-address box on a native host sits on the switch too; a host-address box on a native host shares the host's namespace and its answer; resolution only: reach over the name is local reach under the box's egress rules (NET-079; design §7.1 local names), so a deny-all box resolves it and reaches nothing -->
+  - WHILE a box is a host-address box THE SYSTEM SHALL resolve its lookups through the box zone's answerer and never through the host's own resolver.
+    tier:   T0
+    verify: cargo nextest run -p minimald host_ip_box_resolves_through_answerer
+    <!-- design §4.1 (the deny-all carve-out) and §5.3; state-driven; on a native host the stub a host-address box would otherwise reach is the host's resolver, which forwards any name upstream, so a deny-all box would resolve arbitrary names through the carve-out; the answerer forwards nothing and holds only the zone -->
```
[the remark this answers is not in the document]
norrietaylor: Taken in cc13e960. The NET-003 rule stays unconditional in ownership, a resolver Minimal owns and never the host's, which is arch#78 ruling 2's shape, and now says which resolver per host: the node's DNS layer inside a VM-backed host, the box zone's answerer on a native host. The design reasoning says the native answerer forwards nothing, so a native host-address box with a name allow list has nothing that resolves those names today, and a new MEDIUM open question carries that gap (NET-066, NET-079) instead of the spec claiming a channel that does not exist. The NET-079 carve-out now points at that Minimal-owned resolver rather than naming the answerer.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -126,6 +126,10 @@ included, with every refusal logged (NET-001 to NET-004).
   tier:     T0
   verify:   cargo nextest run -p minimald host_min_internal_resolves_to_host_reach_address_per_mode
   <!-- S1a/AC2; prose 3; ubiquitous; "inside boxes" per design §7.1 without qualification: an own-address box on a native host sits on the switch too; a host-address box on a native host shares the host's namespace and its answer; resolution only: reach over the name is local reach under the box's egress rules (NET-079; design §7.1 local names), so a deny-all box resolves it and reaches nothing -->
+  - WHILE a box is a host-address box THE SYSTEM SHALL resolve its lookups through a resolver Minimal owns and never through the host's own resolver: the node's DNS layer inside a VM-backed host, the box zone's answerer on a native host.
```
msample: Over-broad on a native host. As written this binds every host-address box on a native host to the box zone's answerer, which forwards nothing and answers unknown names NXDOMAIN (design §7.1). A host-address box with no `egress` section is not deny-all (NET-074 scopes the deny-all default to own-address boxes), so on a native host it would keep unrestricted egress and lose every upstream name.

gominimal/arch#78 ruling 2 is narrower than this. It names the resolver the deny-all carve-out points at, and it says `host_ip` boxes resolve through Minimal's resolver *so that §5.3's semantics hold*, where §5.3's resolver forwards under the cohort rules. Where no Minimal component forwards, the ruling has nothing to bind yet; that is exactly the open question this PR adds, so the requirement should not pre-empt it.

Suggested scope: the VM-backed half stays unconditional (the node's DNS layer forwards under rules, NET-066); the native half binds only the deny-all box, whose carve-out NET-079 already points at the answerer; the native non-deny-all box belongs to the open question, with the interim stated there. For example:

> - WHERE the host is VM-backed, WHILE a box is a host-address box THE SYSTEM SHALL resolve its lookups through the node's DNS layer and never through the host's own resolver.
> - WHERE the host is not VM-backed, WHILE a host-address box is declared deny-all THE SYSTEM SHALL resolve its lookups through the box zone's answerer and never through the host's own resolver.

The design reasoning at lines 874 to 877 understates the same point: it is not only the allow-list box that has nothing to resolve upstream names today, it is every native host-address box that is not deny-all. The foundry bot's first thread had this right; naming the resolver per host did not remove the consequence.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -507,7 +511,19 @@ included, with every refusal logged (NET-001 to NET-004).
...
+  <!-- S9b/AC2; prose 50; state-driven; the per-box decision for host-address boxes is the box host's classifier (design §4.1, UC3), matching the box's own cgroup before any source translation; outside the box host the cohort is one identity (NET-078) and the resident union is the floor -->
+  - WHILE a host-address box is declared deny-all THE SYSTEM SHALL admit the box's connections to the resolver Minimal owns for it, at that resolver's address and port, and to no other loopback destination.
+    tier:   T0
+    verify: cargo nextest run -p minimald host_ip_deny_all_reaches_only_the_answerer
+    <!-- design §4.1; state-driven; the one carve-out from a deny-all verdict, by address and port, never loopback-wide: a loopback baseline exception for every box is rejected in the design reasoning; with NET-003's resolver rule a deny-all box resolves exactly the names that resolver holds and reaches nothing else -->
+  - THE SYSTEM SHALL keep every process of a host-address box inside the cgroup its verdict is decided on, so that no process in the box can move itself or a child out of it.
```
msample: Concurring with the foundry thread here, with the arch anchor: design §4.1 v0.8 makes the placement inescapable for every `host_ip` box ("minimald MUST make a box's cgroup placement inescapable from inside the box"), because the classifier decides every host-address box's verdict on its cgroup, not only a deny-all one. Nested under NET-079's deny-all WHILE with a bare `ubiquitous` marker, this reads as deny-all-scoped. Give it its own trigger, `WHILE a box is a host-address box`, or lift it beside NET-079.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -507,7 +511,19 @@ included, with every refusal logged (NET-001 to NET-004).
...
+    tier:   T0
+    verify: cargo nextest run -p minimald host_ip_box_cannot_leave_its_cgroup
+    <!-- design §4.1; ubiquitous; the box host's obligation, not a property of the kernel: the box is confined by a cgroup namespace rooted at its own placement on a mount that treats namespaces as delegation boundaries, with the host's cgroup mount kept out of the box's mount namespace; a box as the daemon's user with write access to a common ancestor can otherwise migrate itself -->
+  - WHERE the host is not VM-backed, WHILE the box host cannot decide per box, because the classifier's privileged install step is not installed or the host cannot confine a box as above, WHEN a session starts THE SYSTEM SHALL print an advisory naming the exact command that installs the step, with no privilege prompt, and record that host-address boxes have no per-box enforcement on that host.
```
msample: Two things. First, the advisory names "the exact command that installs the step" for both causes, but when the cause is that the host cannot confine a box (cgroup2 mounted without `nsdelegate`, or the host's cgroup mount cannot be kept out of the box's mount namespace), installing the step clears nothing, so the advisory would send the operator to a command that changes nothing. Name the cause, and name the command only when the missing step is the cause.

Second, the comment says the two rejected shapes "are rejected in the design". Design §7.4 v0.8 records the chosen shape; the rejections are on gominimal/arch#78 ruling 1. Cite the ruling.

On the foundry thread below about macOS: the WHERE is fine as it stands. The spec models macOS hosts as VM-backed (line 94), so "not VM-backed" is the co-resident Linux host and nothing else. That thread can close on that basis.

msample (commented):
Reviewed at cc13e960 against gominimal/arch main 675b096. The closed HIGH and MEDIUM questions are settled by design v0.8 and the two spikes; the macOS boot-step prose and NET-124's cacheable negatives match §7.1; NET-079's carve-out, the cgroup obligation, and the native install step match §4.1, §7.4, and ruling 1 as held on #1562.

Four threads inline, most severe first: NET-003's new sub-requirement is over-broad on a native host, where every host-address box that is not deny-all would lose upstream names; the cgroup-confinement sub-requirement reads as deny-all-scoped (concurring with the foundry thread, with §4.1's MUST as the anchor); the native advisory names the install command for a cause it does not fix and cites the design for rejections that live on gominimal/arch#78; and the new open question needs its interim stated.

On the other bot threads: CodeRabbit's wording for NET-124 (the SOA record in the authority section) is correct per RFC 2308 and should be taken. The foundry macOS thread is moot: the spec models macOS hosts as VM-backed (line 94), so "not VM-backed" is the co-resident Linux host and nothing else.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -966,20 +1008,14 @@ daemon does, and the attribute that makes that gap visible to policy is EHE's.
...
-  macOS publishes every box at `127.0.0.1`, and NET-010 binds distinct addresses
-  once it is; the interim is a per-host state, not a platform exception.]
-- [NEEDS CLARIFICATION (MEDIUM): what is the cgroup layout for the two-address
-  classifier that gives each host-address box its own identity inside the box
-  host (NET-079), and are host-address boxes on a co-resident Linux host in
-  scope of NET-078 to NET-080? [Design §7.4][design] applies the profile's
-  naming and addressing to that host and now permits a node-local Box Egress
-  Proxy inside its boundary at the advisory tier; it says nothing about the
-  classifier.]
+- [NEEDS CLARIFICATION (MEDIUM): on a native host, which component forwards a
```
msample, line 1021: The question should also state the interim, since NET-003 and NET-079 now bind T0 behaviour around it: while it is open, what does a native host-address box that is not deny-all resolve through, and what does the host's per-box enforcement claim cover there? The honest interim is the host's resolver, with CIDR rules decided by the classifier and name rules (NET-066) not enforced on that host, and the enforcement the host records under NET-079 must not read as covering names. Without that sentence T38 has to invent the interim.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
norrietaylor: Taken in ff22bf9a, as suggested: the VM-backed half binds every host-address box to the node's DNS layer; the native half binds only the deny-all box to the answerer. A native box that is not deny-all, with or without an `egress` section, is bound by nothing there and belongs to the open question. The design reasoning now says so for every non-deny-all native box, not only the allow-list one. The VM-backed half carries a new test name, so the plan's T2 gains a test line.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
norrietaylor: Taken in ff22bf9a: the confinement sub-requirement has its own trigger, `WHILE a box is a host-address box, whatever its declaration`, and its comment says it is not inherited from the deny-all WHILE, with the sibling-verdict consequence as the reason.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
norrietaylor: Taken in ff22bf9a: the advisory names the cause, and the install command only when the missing step is the cause; the comment lists both causes and says installing the step clears only the first. On the citation: the spec convention keeps issue numbers out of docs/specs, so the comment cites the ruling as the architecture's ruling on the native install step, at design §7.4 and its v0.8 change history entry, which is where the design records ruling 1 by number. Say if you want the bare number regardless.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
norrietaylor: Taken in ff22bf9a: the question states the interim in your words, and the same interim appears in the design reasoning and in the NET-079 advisory comment, so T38 reads it in three places: the host's resolver, address rules decided by the classifier, name rules not enforced on that host, and recorded per-box enforcement covering addresses and never names.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -507,7 +511,19 @@ included, with every refusal logged (NET-001 to NET-004).
...
+    tier:   T0
+    verify: cargo nextest run -p minimald host_ip_deny_all_reaches_only_the_answerer
+    <!-- design §4.1; state-driven; the one carve-out from a deny-all verdict, by address and port, never loopback-wide: a loopback baseline exception for every box is rejected in the design reasoning; with NET-003's resolver rule a deny-all box resolves exactly the names that resolver holds and reaches nothing else -->
+  - THE SYSTEM SHALL keep every process of a host-address box inside the cgroup its verdict is decided on, so that no process in the box can move itself or a child out of it.
+    tier:   T0
+    verify: cargo nextest run -p minimald host_ip_box_cannot_leave_its_cgroup
+    <!-- design §4.1; ubiquitous; the box host's obligation, not a property of the kernel: the box is confined by a cgroup namespace rooted at its own placement on a mount that treats namespaces as delegation boundaries, with the host's cgroup mount kept out of the box's mount namespace; a box as the daemon's user with write access to a common ancestor can otherwise migrate itself -->
```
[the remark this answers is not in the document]
norrietaylor: Taken in ff22bf9a: the sub-requirement now carries its own trigger, `WHILE a box is a host-address box, whatever its declaration`, and its comment marks it state-driven and not inherited from the deny-all WHILE.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -507,7 +511,19 @@ included, with every refusal logged (NET-001 to NET-004).
...
+  - WHERE the host is not VM-backed, WHILE the box host cannot decide per box, because the classifier's privileged install step is not installed or the host cannot confine a box as above, WHEN a session starts THE SYSTEM SHALL print an advisory naming the exact command that installs the step, with no privilege prompt, and record that host-address boxes have no per-box enforcement on that host.
+    tier:   T0
+    verify: cargo nextest run -p minimald native_host_advises_classifier_install_without_prompt
+    <!-- design §7.4; state+event; the ruleset needs a capability the native daemon lacks, so a native host takes one privileged install step in NET-122's advisory pattern; until then the host has no per-box host-address enforcement and says so at session start, visible to policy; refusing host-address egress declarations until then, and scoping the classifier to VM-backed hosts, are rejected in the design -->
```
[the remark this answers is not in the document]
norrietaylor: Not changed. The spec models macOS hosts as VM-backed (the platform paragraph near line 94), so `WHERE the host is not VM-backed` is the co-resident Linux host and nothing else; there is no native macOS branch for this requirement to cover. The advisory wording did change in ff22bf9a for a different reason: it names the cause, and the install command only when the missing step is the cause.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -618,6 +634,10 @@ included, with every refusal logged (NET-001 to NET-004).
   tier:     T0
   verify:   cargo nextest run -p minimald non_a_in_zone_query_is_nodata
   <!-- S1b-2b; design §7.1 (answer semantics); event-driven; never NXDOMAIN, since negative caching is name-wide and browsers pair A with HTTPS-type queries -->
+  - WHEN the answerer returns NODATA or NXDOMAIN THE SYSTEM SHALL carry the zone's authority record in the answer so that the host resolver can cache the negative.
```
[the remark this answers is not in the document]
norrietaylor: Taken in ff22bf9a with your wording: the zone's SOA record in the authority section of the response.

docs/specs/18-spec-box-networking/18-spec-box-networking.md
```diff
@@ -507,7 +515,19 @@ included, with every refusal logged (NET-001 to NET-004).
...
+    tier:   T0
+    verify: cargo nextest run -p minimald native_host_advises_classifier_install_without_prompt
+    <!-- design §7.4; state+event; the causes are the privileged step not installed, or the host unable to confine a box as above (cgroup2 mounted without delegation-boundary namespaces, or the host's cgroup mount not keepable out of the box's mount namespace); installing the step clears only the first, so the command is named only for it; the ruleset needs a capability the native daemon lacks, so a native host takes one privileged install step in NET-122's advisory pattern; until then the host has no per-box host-address enforcement and says so at session start, visible to policy; refusing host-address egress declarations until then, and scoping the classifier to VM-backed hosts, are the shapes the architecture's ruling on the native install step rejected (design §7.4 and its v0.8 change history); the enforcement the host records covers addresses, never names, while the open question on native forwarding is open -->
```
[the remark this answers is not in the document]
norrietaylor: Half taken in 3404f7cb, half declined on the architecture's ruling.

Taken: NET-079 gains the explicit exception you asked for. On a native host where the box host cannot decide per box, a deny-all box, or one carrying an `egress` section, runs with no per-box verdict and its declaration unenforced, recorded as such, and is never refused on that ground; the parent rule's comment points at it. That also resolves the reading you flagged of the interim: a name rule admits nothing on that host, and the spec now says why in the open question and the design reasoning, since admission comes only from resolution through the resolver Minimal owns and a direct-to-address flow is admitted by address rules alone (design §5.3). So address-only enforcement admits nothing a name rule would have refused; it admits only what a CIDR rule allows.

Declined: a fail-closed deny verdict. That is shape 3 of [gominimal/arch#78](https://github.com/gominimal/arch/issues/78), which ruling 1 rejected: a missing install step is surfaced at session start and made visible to policy as the node attribute in design §7.4, never a hard failure for a mode whose chooser accepted the reduced tier. Whether such a box may run there is policy's call on that attribute, not the box host's.

msample (approved):
Approving. Checked the fold against design v0.8 (§4.1, §5.3, §7.1, §7.4, §12, change history) and both spike docs: the settled answers land faithfully — sibling-leaf layout, `nsdelegate` + mount-hygiene confinement conditions, socket-keeps-cgroup, the address-and-port carve-out, the RFC 2308 SOA behavior including that a resolver reload doesn't clear the stall — with no ID or test-name churn, and the new MEDIUM question is honest about the native forwarding gap. The earlier review rounds' scoping fixes all verify at `3404f7cb`.

Four findings, all non-blocking, tracked so they outlive this thread:

1. **Native answerer reachability** (gominimal/arch#82) — NET-003's native sub + NET-079's carve-out need the answerer reachable on a port a box's stub can express; §7.1 conditions the `:53` listener on NetworkManager, so on systemd-resolved hosts `host_ip_box_resolves_through_answerer` has no implementable path yet. Design-letter fix, then one comment sentence here.
2. **Attribute locus** (#1565) — "policy's call on the recorded attribute" risks a node-asserted reading; §4.4 makes `host_ip_enforcement` provider/admin-asserted with the daemon probe only cross-checking via heartbeat, and the attribute is EHE's.
3. **Interim resolver admission** (#1565) — what admits a non-deny-all native box's DNS flow to the host resolver's loopback address under its egress rules, given the carve-out is "never loopback-wide"; implicit carve-out or required declaration, either should be stated.
4. **Cosmetics** (#1565) — paragraph break in the design reasoning, "at each boot", "minimald never re-applies it", optional NET-125 cross-ref to the SOA sub.

